PR #45, opened 2020-03-26
Update installation & contribution instructions & add Makefile

radeksimko:
This change effectively enforces developers to build through the make target
and prevents server of version `0.0.0` from ever appearing in the wild.

As for the makefile logic:
 - if the version is not provided we source the latest one from GitHub API
 - in the "happy path" (being online and GitHub's API working correctly) `make build` "just works"
 - otherwise (e.g. when offline) we can still build e.g. via `VERSION=0.1.0 make build`

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 ## Reporting Feedback
 
-Terraform Language Server is an open source project and we appreciate
+Terraform language server is an open source project and we appreciate
 contributions of various kinds, including bug reports and fixes,
 enhancement proposals, documentation updates, and user experience feedback.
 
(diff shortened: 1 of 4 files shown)


tommyknows:
Sorry for chiming in here, but I'm not quite sure why this has been introduced. 
This change prevents people from just installing the language server through go get:
```
$> go get github.com/hashicorp/terraform-ls
$> terraform-ls
panic: Please use 'make build' to compile and install

goroutine 1 [running]:
github.com/hashicorp/terraform-ls/version.init.0()
        /Users/ramon/Documents/go/pkg/mod/github.com/hashicorp/terraform-ls@v0.2.0/version/version.go:27 +0xc5
```

I don't see why a server version 0.0.0 should explicitly be avoided, could you elaborate?
It would be nice if this project was just go-gettable like pretty much every other Go project.

PR #48, opened 2020-04-11
build: Code-sign all MacOS & Windows releases

### LINKED ISSUE #44: build: Add MacOS notarization
As noted in https://github.com/hashicorp/terraform-ls/pull/43 MacOS/darwin builds are prevented from being launched on MacOS Catalina because they are not signed.

We should look into signing - it seems to be supported in goreleaser:
https://goreleaser.com/customization/#signing-with-gon

radeksimko:
While this is tested (in a private fork) and working, I'm marking this PR as draft for now.

This is because it relies on [an upstream goreleaser PR](https://github.com/goreleaser/goreleaser/pull/1414) which adds more flexible build hooks and it would be great for us to avoid relying on a personal fork of mine.

In other words I'm hoping this PR will get merged and we could just use latest upstream goreleaser release.

Closes #44


radeksimko:
PR was merged 🎉 , just waiting for next release.

PR #51, opened 2020-04-15
Implement textDocument/formatting

### LINKED ISSUE #28: Implement textDocument/formatting
https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_formatting

This will allow IDEs to format code, like `terraform fmt` would format it.

radeksimko:
Closes #28


radeksimko:
I'm half thinking if it's worth still replacing the `hclwrite.Format` with a call out to `terraform fmt`. The only downside is a little more plumbing/mocking in tests and potentially different failure scenarios, but I'm still leaning more towards doing it.

aeschright:
I think this looks good, aside from making a decision on that detail.

musabmasood:
Dumb question, how can I enable this in vscode? `terraform fmt` doesn't seem to be running on save. Just updated to 2.0.1 plugin, everything else seems to be working fine.

aeschright:
It uses the VSCode settings here: https://github.com/hashicorp/vscode-terraform/issues/375#issuecomment-642707592

You won't see a call out to `terraform fmt` specifically because it's routed through the language server.

PR #67, opened 2020-04-24
build: Add CI setup

### LINKED ISSUE #60: build: Add CI step for tests
We should add a CI step that can run for every incoming PR.

radeksimko:
Closes #60 

--- 

This is currently WIP. Some pipelines are failing for the following reasons

 - Even though tests are mocking all Terraform executions, the discovery of the binary still remains unmocked, which causes some tests to fail as it can't find `terraform` in `$PATH`
   - Mocks need to be added for discovery
 - Due to our code-signing being a post-build hook, `goreleaser` snapshot build fails too.
   - I don't think we should be signing our PR builds - just need to find a way of building binaries without post build hooks, perhaps avoid using GoReleaser in this context (PR) altogether

--- 

Depends on:
https://github.com/goreleaser/goreleaser/issues/1469
https://github.com/hashicorp/terraform-ls/pull/73
https://github.com/hashicorp/terraform-ls/pull/70


radeksimko:
Opened https://github.com/goreleaser/goreleaser/issues/1469

internal/terraform/exec/exec_test.go
```diff
@@ -13,7 +14,7 @@ func TestExec_timeout(t *testing.T) {
 		SleepDuration: 100 * time.Millisecond,
 		Stdout:        "Terraform v0.12.0\n",
 	})
-	e.SetWorkdir("/tmp")
+	e.SetWorkdir(os.TempDir())
```
paultyng, line 17: Does this create any files? May be cleaner to create a tmp dir and remove it on a t.Cleanup? https://golang.org/pkg/io/ioutil/#TempDir

radeksimko: It doesn't - or at `version` execution in particular shouldn't and there is no other code that would. It just essentially needs to `cd` into a valid directory on the filesystem, so it should be safe to just use just `os.TempDir()` here, I think, but I'm working through the other tests to assess the situation there

radeksimko: I have pulled this out into a separate PR: https://github.com/hashicorp/terraform-ls/pull/73
Review is welcomed there. 😄

radeksimko:
https://github.com/goreleaser/goreleaser/pull/1520 (adding support for `goreleaser build --skip-post-hooks`) was merged, just pending release. Then I can leverage that new command here in this PR and finally get it in. 😄

PR #78, opened 2020-04-30
terraform/exec: Treat TempDir related vars as safe to pass through

### LINKED ISSUE #77: Make server work under non-admin user on Windows
### Server Version

```
0.1.0
```
(launched as regular, non-admin user on Windows 10 Pro)

### Terraform Version

```
0.12.24
```

### Client Version

```
Sublime Text LSP v0.9.8
```

```hcl
provider "aws" {
}
```

### Log Output

```
2020/04/29 22:29:10 exec.go:172: Starting C:\ProgramData\chocolatey\bin\terraform.exe ["terraform" "version"] in "C:\\Users\\RADEKS~1\\AppData\\Local\\Temp"...
2020/04/29 22:29:10 exec.go:138: Waiting for command to finish ...
2020/04/29 22:29:10 rpc_logger.go:45: Error for "initialize" (ID 1): [-32098] failed to get version: terraform (pid 5008) exited (code 1): exit status 1
stdout: ""
stderr: "Couldn't setup logging tempfile: open C:\\Windows\\terraform-log151023907: Access is denied."
```

### Expected Behavior

`terraform version` should've run successfully, with log file being created in temporary directory.

### Actual Behavior

Server stopped because Terraform stopped because Terraform couldn't create the temp log file in `C:\Windows` (which is the [default `TempDir` location](https://golang.org/pkg/os/#TempDir) if no ENV variables are available, as Windows doesn't seem to have a static system-wide temp dir? 😱 ).

### Steps to Reproduce

1. Open a folder with the above config
2. Open file with the above config from that folder
3. Watch log

--- 

### Proposal

We just need to whitelist a few more ENV variables here

https://github.com/hashicorp/terraform-ls/blob/4bb04ee41d984629e4e7f456b8f7f9405d8e253e/internal/terraform/exec/exec.go#L98-L109

radeksimko:
Closes #77

### DIFF
--- a/internal/terraform/exec/exec.go
+++ b/internal/terraform/exec/exec.go
@@ -19,6 +19,12 @@ import (
 	"github.com/hashicorp/terraform-ls/logging"
 )
 
+// Environment variables to pass through to Terraform
+var passthroughEnvVars = []string{
+	// This allows Terraform to find custom-built providers
+	"HOME", "USER",
+}
+
 // cmdCtxFunc allows mocking of Terraform in tests while retaining
 // ability to pass context for timeout/cancellation
 type cmdCtxFunc func(context.Context, string, ...string) *exec.Cmd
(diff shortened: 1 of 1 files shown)


radeksimko:
re `TF_VAR_*` - We can't really do anything useful with these variables yet (we only run `version` and `providers schema`).

More importantly though it is questionable whether the user can ensure that these ENV vars _actually_ get propagated to the IDE in the first place and whether the IDE would pass them down to the server.

And even if they did, I have some doubts we'd interpret them the way user expects them to be interpreted - specifically because single IDE instance can serve many directories, each requiring different variables/context. So unless all of your variables apply to all projects you're working on, it would cause more confusion, because these variables will share the lifecycle of the IDE.

Terraform should however automatically load `terraform.tfvars` and we could also introduce a config option `load_tfvars` which would additionally attempt to load tfvars file of a given name in any directory or something like that.

Issue #88, opened 2020-05-08
Make project go get-able


radeksimko:
> Sorry for chiming in here, but I'm not quite sure why this has been introduced. 
> This change prevents people from just installing the language server through go get:
```
$> go get github.com/hashicorp/terraform-ls
$> terraform-ls
panic: Please use 'make build' to compile and install

goroutine 1 [running]:
github.com/hashicorp/terraform-ls/version.init.0()
        /Users/ramon/Documents/go/pkg/mod/github.com/hashicorp/terraform-ls@v0.2.0/version/version.go:27 +0xc5
```

> I don't see why a server version 0.0.0 should explicitly be avoided, could you elaborate?
> It would be nice if this project was just go-gettable like pretty much every other Go project.

_Originally posted by @tommyknows in https://github.com/hashicorp/terraform-ls/pull/45#issuecomment-625709368_

radeksimko:
Hi @tommyknows
There was a few motivating factors behind that PR you commented on.

Rest assured that `go get`ing and build-from-source installation is generally something we intend to support going forward. The PR was merely aiming to communicate this is not meant to be _the default way_ of installation for end-users and I'll try to explain why below.

## Installation Tracking

It reduces our chance to track installations and such data can be critical in decision making about the future of the project.

There are ways we can address that:
 - GitHub exposes some data for repository owners about number of `go get`s of their Go modules over time
 - we front our import paths with a proxy that can provide such data - which comes with its own challenges that may not outweigh the benefits
 - we reduce the number of people that will use that method of installation, which makes the skew in numbers negligible

These are not necessarily mutually exclusive options and I think we can look into combining them carefully to reduce the negative effects.

## Prescribing build environment

It limits our ways of communicating how the server is supposed to be compiled (e.g. with what version of Go or what compiler flags).

This could lead to folks trying to compile the server with an older/newer version of Go or passing certain compilation flags, which could in turn lead to them experiencing issues that are not reproducible by anyone else. It happens quite rarely in Go, but I have experienced that myself and these kinds of issues are the most difficult to deal with.

I believe/hope this will eventually be addressed by Go itself as the tooling matures - it looks like `go` stanza in `go.mod` is heading in that direction.

## Identifying dev builds

Identification of dev builds is often needed when other maintainers and contributors file issues. It is useful to know whether 

[...]

PR #98, opened 2020-05-12
Support simplified building and go get

paultyng:
I modified the usage of ldflags to support the goreleaser default variables it fills in and removed the `build` target from the Makefile. I think in the docs we should just encourage people to use `goreleaser --snapshot` locally, unfortunately you can't seem to skip the Artifactory step currently.

Curl / Make are not available by default on Windows, and we should probably make it more approachable for as many devs as possible to contribute. Considering the length of the Makefile now we could also just remove it but we should definitely call out the vendor usage for a canonical test environment.


paultyng:
This would address #88

main.go
```diff
@@ -3,15 +3,23 @@ package main
...
+
+	"github.com/hashicorp/terraform-ls/commands"
+)
+
+var (
+	// These vars will be set by goreleaser.
+	version string = `dev`
+	commit  string = ``
+	date    string = ``
+	builtBy string = ``
```
radeksimko, line 16: As discussed offline I'm ok with these variables being in `main` package - as we can pass them down to wherever they're needed and it is better than global state in `version` package.

However (1) I think it'd be cleaner to still keep them in a separate file and (2) it would be better to retain some of the logic from the `version` package:

 - variable names `version`/`prerelease` - which can be un-exported, but I think we should retain the meaning and avoid setting `version` to `dev`. I'd say the meaningful default is `version = "0.0.0"` `prerelease = "dev"`
 - semver check via `init()` - which won't be `panic`ing on `go get/build/install` anymore if `version` defaults to `0.0.0`, but still gives us safety net against building a version which isn't semver-compatible

paultyng: A separate file is a good call.

Why have separate fields? It gets complicated for snapshots as its already going to send a prerelease string to the version (ie. `1.2.3-SNAPSHOT-hash`) or if your goreleaser tag is prerelease, it will mess it up wouldn't it? As the `version` var and the `prerelease` var will kinda be redundant. Could just do `version string = "0.0.0-dev"` couldn't you?

radeksimko: As for GoReleaser - it may do by default, but it won't with custom ldflags, like we have now (without this patch), except that these flags would be less verbose with variables being in `main`.

As for why - I don't have any better reason other than keeping it aligned with other HashiCorp Go projects ([terraform](https://github.com/hashicorp/terraform/blob/master/version/version.go#L19), [packer](https://github.com/hashicorp/packer/blob/master/version/version.go#L17), [consul](https://github.com/hashicorp/consul/blob/master/version/version.go#L23), [vault](https://github.com/hashicorp/vault/blob/master/sdk/version/version.go#L12), [nomad](https://github.com/hashicorp/nomad/blob/master/version/version.go#L19)). 

That said I appreciate you challenging this convention and I have not asked teams _why_ they do it and _if_ they still feel this is right in today's context, as the separate package thing probably isn't in retrospect and this may be just a relic of some inflexible release tooling we may still have in place.

So I'll ask around next week - and if there are no strong opinions/reasons then I'm happy to go ahead with the "GoReleaser style" you're proposing! 😄

radeksimko:
As discussed/clarified via DMs I have opened https://github.com/goreleaser/goreleaser/issues/1532

radeksimko:
https://github.com/goreleaser/goreleaser/pull/1557 will be part of the next GoReleaser release, which means we can leverage `RawVersion`, `Prerelease` and `IsSnapshot` variables in the config and retain the variable names and meanings as we use in other HashiCorp projects and still use GoReleaser's snapshot builds.

https://goreleaser-git-f-snapshot-versioning.goreleaser.now.sh/customization/templates/

We can put this PR hold temporarily until the next release comes out as that should happen 🔜

Issue #49, opened 2020-04-14
Publish release binaries to releases.hashicorp.com


paultyng:
In addition to the GitHub release.

We should also build the JSON index as well to determine latest version.

radeksimko:
I'd expect `hc-releases` here to do most of the hard work (including the index building?), but I'm not that familiar with it, so I'll need to read through the implementation/docs.

We could plug it into the existing GitHub Action either as a next step after GoReleaser, or send a patch to GoReleaser to support global post hooks, similar to pre-hooks: https://goreleaser.com/customization/#Global%20Hooks
and then plug it in there.

paultyng:
I think probably easiest to just chain it afterwards in the GitHub action unless there is some context you'd want to reuse that goreleaser would provide?

radeksimko:
The benefit of reusing GoReleaser pipeline is that we could publish Homebrew/Scoop formula with our releases.hashicorp.com URLs, which make it much easier for us to track downloads in much more detailed way - including the ability to segment downloads via Homebrew or Scoop.

Technically we could do that even without any GoReleaser patches, but we'd be publishing Homebrew/Scoop updates before the binaries are actually uploaded and available, which is a bit awkward.

Download tracking via GitHub API is currently _very minimal_ as it doesn't go beyond a single number representing total download count 😑 , so we wouldn't be able to identify Homebrew or Scoop downloads that way and tracking weekly or monthly downloads would also be more difficult.

radeksimko:
I created https://github.com/goreleaser/goreleaser/issues/1460 to gather some preliminary feedback on a potential upstream patch that could support this integration.

PR #99, opened 2020-05-13
build: Sign archives checksum

radeksimko:
This is in preparation for https://github.com/hashicorp/terraform-ls/issues/49 as we can't publish any artifacts to releases.hashicorp.com without the signature.

cc @aeschright - FYI we could also verify the signature when installing the LS in VSCode. I'm just not sure what's the best workflow there and whether to bundle the pubkey (probably more secure, less convenient for potential updates) or pull it down from somewhere on demand (theoretically less secure, but more convenient for updates).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -30,12 +30,16 @@ jobs:
         env:
           VERSION: v0
       -
+        name: Import PGP key for archive signing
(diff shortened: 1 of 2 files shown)


paultyng:
Considering our historic bundling of the HC public key in TF, its probably safe to bundle it in the VSCode extension. We can discuss further internally though what the best strategy is.

Issue #36, opened 2020-03-24
parser: Support 1st class blocks without exposed schema


radeksimko:
## Current Situation

LS currently parsers attributes and nested blocks within blocks which have their schema exposed in JSON format via `terraform providers schema -json`. The following are examples which don't fall into that category:

 - `variable`
 - `locals`
 - `output`
 - `module`
 - `terraform`

and meta parameters such as

 - provisioners
 - backends
 - `count`
 - `for_each`
 - etc.

Parsing of these blocks may therefore need to be approached differently.

## How we could tackle this

This will likely require extensive design work to ensure stability and maintainability over time. There is a number of factors to consider, e.g.

 - How will the functionality overlap with https://github.com/hashicorp/terraform/blob/master/configs/parser_config.go
   - we probably won't build a separate single library which decouples this logic yet, but should keep this option in mind as the risk of drifting from Terraform is real
 - Versioning of language parser (`internal/terraform/lang`)
   - The parser is currently constrained to `>=0.12.0` and [`FindCompatibleParser`](https://github.com/hashicorp/terraform-ls/blob/253be7a050d25dd1729183ad9bd36304d710fab0/internal/terraform/lang/parser.go#L61-L70) only finds a single parser. We can use this scaffold to reflect the reality of syntax changing between TF versions
 - How could we leverage the existing HCL parsing logic in [`internal/terraform/lang/hcl*`](https://github.com/hashicorp/terraform-ls/tree/master/internal/terraform/lang)
   - It is currently coupled with `tfjson` structs, but we could try to turn these structs into compatible interfaces and have a similar static implementation ("schema dispenser") for the above blocks
 - How do we accommodate cases which have part of the schema dynamic (provider, resource, datasource) and part static (meta parameters like `provisioner`, `count` etc.)

For reasons above I would encourage discussing the design of the possible solution(s) here before attempting to raise a PR.

PR #104, opened 2020-05-19
textDocument/complete: Complete first level keywords

### LINKED ISSUE #18: textDocument/complete: Complete supported first-class keywords
Since MVP will support provider, resource and data source completion after #9 and #10 we should also be able to complete the following first-class keywords:

 - `provider`
 - `resource`
 - `data`

and insert appropriate snippets for each - i.e. labels, braces with cursor on the block type.

Related to snippets: #13

radeksimko:
Closes #18

As described in that issue, this only adds support for the blocks which have schema provided by providers, supporting other blocks is generally tracked under #36

### UX

Screenshot 2020-05-19 at 12 26 36
Screenshot 2020-05-19 at 12 27 23

### DIFF
--- a/internal/hcl/errors.go
+++ b/internal/hcl/errors.go
@@ -22,3 +22,8 @@ type NoBlockFoundErr struct {
 func (e *NoBlockFoundErr) Error() string {
 	return fmt.Sprintf("no block found at %#v", e.AtPos)
 }
+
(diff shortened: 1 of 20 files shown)


radeksimko:
> I'm realizing its going to be harder and harder to switch to hclog if we don't soonish.

That is a good point. I'm going to create an issue for that and try to look into that for next milestone.

Issue #69, opened 2020-04-25
terraform/discovery: Re-load path to terraform on version change


radeksimko:
### Current Version
```
v0.1.0
```

### Use-cases

Allow users switching between Terraform versions whilst they have any files open in the IDE.

### Attempted Solutions

We perform discovery of the Terraform binary only once per session, so the user has to reopen the session (folder), if they change Terraform version and want to keep all functionality working and accurate.

### Proposal

We could trigger re-discovery (of the absolute path to Terraform and its new version) and re-reading of schema.

The trigger for such event could be:
 - `tfenv use`, or simply change of `.terraform-version` file
 - disappearance of the current `terraform` binary
 - byte change of the current `terraform` binary (its checksum)

radeksimko:
Based on some local experiments it appears that tfenv should work out of the box, given that it uses a shell script called `terraform` in placed in $PATH to front the real binary of any particular version.

We may just need to remove `terraform.exe` lookups for Windows, to enable similar pattern on Windows.

PR #109, opened 2020-05-21
Add ability to specify path to TF binary

paultyng:


langserver/langserver.go
```diff
@@ -12,12 +12,15 @@ import (
...
 	"github.com/creachadair/jrpc2/server"
+
+	"github.com/hashicorp/terraform-ls/internal/terraform/discovery"
 	"github.com/hashicorp/terraform-ls/langserver/session"
 )
 
 type langServer struct {
 	srvCtx     context.Context
 	logger     *log.Logger
+	tfPath     discovery.DiscoveryFunc
```
radeksimko, line 23: Nitpick, but do you mind calling this `tfDiscoFunc` or something similar that makes it more obvious that this variable is a function and not a string?

I suppose the other `tfPath` in `service.go` is a forgotten relic that should also be changed - especially once https://github.com/hashicorp/terraform-ls/issues/69 is addressed. I'm happy to address it as part of #69 though - no biggie.

langserver/session/types.go
```diff
@@ -5,12 +5,15 @@ import (
...
 	"github.com/creachadair/jrpc2"
+
+	"github.com/hashicorp/terraform-ls/internal/terraform/discovery"
 )
 
 type Session interface {
 	Assigner() (jrpc2.Assigner, error)
 	Finish(jrpc2.ServerStatus)
 	SetLogger(*log.Logger)
+	SetDiscovery(discovery.DiscoveryFunc)
```
radeksimko, line 16: Nitpick: Can we call this `SetDiscoveryFunc`, so it's more obvious that the argument is a function?

langserver/handlers/service.go
```diff
@@ -53,6 +53,10 @@ func (svc *service) SetLogger(logger *log.Logger) {
 	svc.logger = logger
 }
 
+func (svc *service) SetDiscovery(f discovery.DiscoveryFunc) {
```
radeksimko, line 56: Nitpick: Can we call this `SetDiscoveryFunc`?

paultyng:
@radeksimko made the suggested adjustments, also added some testing of the path string to make sure its a bit more valid as discussed. I cleaned up a few of the other `ui.Error` calls in another commit as well just for consistency. Let me know if its still good to merge.

commands/serve_command.go
```diff
@@ -79,18 +82,43 @@ func (c *ServeCommand) Run(args []string) int {
...
+		}
+		stat, err := os.Stat(path)
+		if err != nil {
+			c.Ui.Error(fmt.Sprintf("Unable to find Terraform binary: %s", err))
+			return 1
+		}
+		if stat.IsDir() {
+			c.Ui.Error(fmt.Sprintf("Expected a Terraform binary, got a directory: %q", path))
+			return 1
+		}
```
radeksimko, line 103: Total nitpick, but I assume the above lines could make a `validateTfExecPath(string) error` function.

paultyng:
The confusing part was they weren't on all errors just some errors, but if its the how it should be presented you could already create a fancy UI implementation that adds the separation.

I was hoping to maybe revisit/refactor this after the release anyway, clean up the hard coded exit codes and flag validation a bit. I'll just write an issue to do so if thats ok with you?

PR #115, opened 2020-05-24
Improve UX of completion items

paultyng:
Move descriptions to the documentation field, update to terraform-json 0.5.0 that has additional description support, etc.

Using the `documentation` instead of `detail` gives it a better look in VS Code at least:

image
(my misspelling aside)

### DIFF
--- a/internal/lsp/completion.go
+++ b/internal/lsp/completion.go
@@ -26,14 +26,22 @@ func CompletionList(candidates lang.CompletionCandidates, pos hcl.Pos, caps lsp.
 }
 
 func CompletionItem(candidate lang.CompletionCandidate, pos hcl.Pos, snippetSupport bool) lsp.CompletionItem {
+	// TODO: deprecated / tags?
+
+	doc := ""
+	if c := candidate.Documentation(); c != nil {
+		// TODO: markdown handling
+		doc = c.Value()
+	}
+
 	if snippetSupport {
 		pos, newText := candidate.Snippet(pos)
-
 		return lsp.CompletionItem{
 			Label:            candidate.Label(),
 			Kind:             lsp.CIKField,
 			InsertTextFormat: lsp.ITFSnippet,
 			Detail:           candidate.Detail(),
+			Documentation:    doc,
 			TextEdit: &lsp.TextEdit{
 				Range: lsp.Range{
 					Start: lsp.Position{Line: pos.Line - 1, Character: pos.Column - 1},
@@ -49,5 +57,6 @@ func CompletionItem(candidate lang.CompletionCandidate, pos hcl.Pos, snippetSupp
 		Kind:             lsp.CIKField,
 		InsertTextFormat: lsp.ITFPlainText,
 		Detail:           candidate.Detail(),
+		Documentation:    doc,
 	}
 }
--- a/internal/terraform/lang/datasource_block.go
+++ b/internal/terraform/lang/datasource_block.go
@@ -41,6 +41,12 @@ func (f *datasourceBlockFactory) BlockType() string {
 	return "data"
 }
 
+func (f *datasourceBlockFactory) Documentation() MarkupContent {
+	return PlainText("A data block requests that Terraform read from a given data source and export the result " +
+		"under the given local name. The name is used to refer to this resource from elsewhere in the same " +
+		"Terraform module, but has no significance outside of the scope of a module.")
+}
+
 type datasourceBlock struct {
 	logger *log.Logger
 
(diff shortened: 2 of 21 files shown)


paultyng:
I'm not sure why the unit test isn't passing, there is something happening in the mock server I think that isn't replicating the normal behavior, as I see the documentation data coming through in the UI.

internal/terraform/lang/datasource_block.go
```diff
@@ -41,6 +41,12 @@ func (f *datasourceBlockFactory) BlockType() string {
 	return "data"
 }
 
+func (f *datasourceBlockFactory) Documentation() string {
+	return "A data block requests that Terraform read from a given data source and export the result under the " +
```
paultyng, line 45: I just kinda grabbed this off the docs site, we can iterate on them, but wanted to put something.

paultyng:
Without markdown support in the LS protocol, we may want to convert markdown to plain text in the near term, not sure. I'll open an issue for markdown support.

internal/terraform/lang/resource_block.go
```diff
@@ -125,8 +131,9 @@ func resourceCandidates(resources []schema.Resource) []CompletionCandidate {
 	candidates := []CompletionCandidate{}
 	for _, r := range resources {
 		candidates = append(candidates, &labelCandidate{
-			label:  r.Name,
-			detail: r.Provider,
+			label:         r.Name,
+			detail:        fmt.Sprintf("Resource, %s provider", r.Provider),
```
radeksimko, line 135: Very subjective (so feel free to ignore 😄 ), but I'm thinking `"Resource (%s)", r.Provider` would provide same/similar value and reduce the length. Maybe the length is not a concern for most providers and IDEs, I'm not sure 🤷

internal/terraform/lang/schema.go
```diff
@@ -69,18 +69,21 @@ func schemaAttributeDetail(attr *tfjson.SchemaAttribute) string {
 		requiredText = "Required"
 	}
 
-	return strings.TrimSpace(fmt.Sprintf("(%s, %s) %s",
-		requiredText, attr.AttributeType.FriendlyName(), attr.Description))
+	return strings.TrimSpace(fmt.Sprintf("%s, %s",
```
radeksimko, line 72: I like the removal of braces here 👍 - it makes it look cleaner.

internal/terraform/lang/schema.go
```diff
@@ -69,18 +69,21 @@ func schemaAttributeDetail(attr *tfjson.SchemaAttribute) string {
...
+		requiredText, attr.AttributeType.FriendlyName()))
 }
 
 func schemaBlockDetail(blockType *BlockType) string {
 	blockS := blockType.Schema()
 
-	requiredText := "Required"
-	if len(blockType.BlockList) >= int(blockS.MinItems) {
-		requiredText = "Optional"
+	detail := fmt.Sprintf("Block, %s", blockS.NestingMode)
```
radeksimko, line 79: I reckon we'll iterate on the human-readable text here over time to e.g. print `list of blocks`, `single block` etc. but this is already big step forward and I like these changes 👍 💯

internal/terraform/schema/schema_storage.go
```diff
@@ -186,6 +188,7 @@ func (s *Storage) Providers() ([]string, error) {
 }
 
 func (s *Storage) ResourceSchema(rType string) (*tfjson.Schema, error) {
+	// TODO: this is going to need to use provider identities, especially in 0.13
```
radeksimko, line 191: 👍 Thank you for the note

langserver/handlers/complete_test.go
```diff
@@ -84,19 +84,22 @@ func TestCompletion_withValidData(t *testing.T) {
...
+						"detail":"Optional, string",
+						"documentation":"Desc 2",
 						"insertTextFormat":1
 					},
 					{
 						"label":"individual",
 						"kind":5,
-						"detail":"(Optional, bool) Desc 3",
+						"detail":"Optional, bool",
+						"documentation":"Desc 3",
```
radeksimko, line 102: The `testSchemaOutput` (mocking output from `providers schema -json`) will also need updating to reflect the addition of `documentation` field.

I'm assuming it's OK to _just_ update it as both `terraform-json` & protocol changes are additive and the missing fields would just become empty strings and cause no harm anywhere? i.e. users with older version of Terraform prior to https://github.com/hashicorp/terraform/pull/22583 still get a decent UX?

internal/lsp/completion.go
```diff
@@ -34,6 +34,9 @@ func CompletionItem(candidate lang.CompletionCandidate, pos hcl.Pos, snippetSupp
 			Kind:             lsp.CIKField,
 			InsertTextFormat: lsp.ITFSnippet,
 			Detail:           candidate.Detail(),
+			// TODO: deprecated / tags?
+			// TODO: markdown handling
```
radeksimko, line 38: I assume it's safe to say this will always be plaintext until SDK actually allows providers to declare anything else, right?

radeksimko: I think the markdown support should be RFCd properly, discussed with the SDK team and it may require further changes in the protocol as mentioned.

internal/terraform/lang/schema.go
paultyng: Agree, just wanted to move closer to core concepts.

langserver/handlers/complete_test.go
paultyng: Yeah, its all additive in the JSON

internal/terraform/lang/resource_block.go
paultyng:
```suggestion
			detail:        fmt.Sprintf("Resource (%s)", r.Provider),
```

paultyng:
@radeksimko I reworked the PR to surface the distinction more cleanly in line with what the LSP protocol should be in the future. Let me know your thoughts.

Issue #12, opened 2020-02-21
textDocument/complete: Use "isIncomplete" property


radeksimko:
Based on some anecdotal evidence and common sense the server should not be sending more than ~100 suggestions for completion in one response - instead it should just send the first 100 best matches and set `isIncomplete` to `true` and let the client recompute (effectively force the user to keep typing to filter suggestions).

The LSP doesn't define a workflow for passing the letters for further filtering of the candidate list, so we will need to read it from the config, basically read anything on the left side of the position until "enclosing character". It may not be ideal in terms of performance, but probably better for simplicity to just leverage HCL parser there to understand where does the content end.

radeksimko:
As mentioned above, the server is responsible for re-inspecting the document after new `complete` request is sent with new position (line:col). The client (per LSP) does not send these further updates with any context that is needed to further filter completion candidates.

Partial parsing would be fairly easy for labels (provider names, resource types, data source types) due to the fact that even config with incomplete labels is still valid HCL:

```hcl
data "aws_" "" { }
```

Implementing this for body completion is more difficult and the main culprit is that the HCL parser ignores unknown tokens (which is probably by design and intended behaviour), such as `na` below:

```hcl
data "aws_ami" "name" {
  na
}
```
so we may need to add some helper methods to HCL, e.g. `TokenAtPos(pos hcl.Pos)` which would allow us to read `na` based on `hcl.Pos`.

PR #125, opened 2020-05-28
refactoring(parser): Pass around tokens instead of blocks

radeksimko:
Previously we were passing around (either generic `hcl` or `hclsyntax`) blocks which don't carry sufficient details about the block when it comes to completion of incomplete configs, as described in https://github.com/hashicorp/terraform-ls/issues/12#issuecomment-619157850

This PR refactors the parsing logic to pass around `hclsyntax.Tokens` instead as that carries the right amount of details we need.

The PR should however be a no-op from user's perspective.


internal/terraform/lang/hcl_parser.go
```diff
@@ -10,7 +10,13 @@ import (
 
 // ParseBlock parses HCL configuration based on tfjson's SchemaBlock
 // and keeps hold of all tfjson schema details on block or attribute level
-func ParseBlock(block *hclsyntax.Block, labels []*ParsedLabel, schema *tfjson.SchemaBlock) Block {
+func ParseBlock(tokens hclsyntax.Tokens, labels []*ParsedLabel, schema *tfjson.SchemaBlock) (Block, error) {
+	hclBlock, _ := hclsyntax.ParseBlockFromTokens(tokens)
```
paultyng, line 14: Could you add a comment here as to why you are ignoring the error?

internal/terraform/lang/hcl_parser.go
```diff
@@ -33,6 +39,29 @@ func ParseBlock(block *hclsyntax.Block, labels []*ParsedLabel, schema *tfjson.Sc
 	return b
 }
 
+func ParseLabels(tokens hclsyntax.Tokens, schema LabelSchema) ([]*ParsedLabel, error) {
+	hclBlock, _ := hclsyntax.ParseBlockFromTokens(tokens)
```
paultyng, line 43: Same as above, could you add a comment here?

internal/terraform/lang/parser.go
```diff
@@ -131,26 +132,28 @@ func (bt *completableBlockType) Documentation() MarkupContent {
...
-	if block == nil {
+func (p *parser) ParseBlockFromTokens(tokens hclsyntax.Tokens) (ConfigBlock, error) {
+	if len(tokens) == 0 {
 		return nil, EmptyConfigErr
 	}
 
+	// It is probably excessive to be parsing the whole block just for type
+	// but there is no avoiding it without refactoring the upstream HCL parser
+	// and it should not hurt the performance too much
+	block, _ := hclsyntax.ParseBlockFromTokens(tokens)
```
paultyng, line 143: Same as above, could you expand the comment here about why ignoring the error?

internal/terraform/lang/provider_block.go
```diff
@@ -70,7 +70,9 @@ func (p *providerBlock) Labels() []*ParsedLabel {
 	if p.labels != nil {
 		return p.labels
 	}
-	p.labels = parseLabels(p.BlockType(), p.labelSchema, p.hclBlock.Labels)
+	labels, _ := ParseLabels(p.tokens, p.labelSchema)
```
paultyng, line 73: Same as above, could you add a comment here?

internal/terraform/lang/resource_block.go
```diff
@@ -81,7 +81,9 @@ func (r *resourceBlock) Labels() []*ParsedLabel {
 	if r.labels != nil {
 		return r.labels
 	}
-	r.labels = parseLabels(r.BlockType(), r.labelSchema, r.hclBlock.Labels)
+	labels, _ := ParseLabels(r.tokens, r.labelSchema)
```
paultyng, line 84: Same as above, could you add a comment here?

radeksimko: Given that `ParseLabels` doesn't actually need to return any error anymore I will just remove the err from the return values.

internal/terraform/lang/provider_block.go
radeksimko: Given that `ParseLabels` doesn't actually need to return any error anymore I will just remove the err from the return values.

PR #132, opened 2020-06-02
textDocument/complete: Use isIncomplete for >100 items

### LINKED ISSUE #12: textDocument/complete: Use "isIncomplete" property

radeksimko:
Closes #12 

Ideally we should probably be limiting the number of items we pull out from whatever data storage, but the data is currently just schema and the schema is currently stored in memory, so it's cheap to just pull out everything and then cut it while constructing the completion list 🤷 

We can revisit later.

### DIFF
--- a/internal/terraform/lang/parser.go
+++ b/internal/terraform/lang/parser.go
@@ -14,6 +14,10 @@ import (
 	"github.com/hashicorp/terraform-ls/internal/terraform/schema"
 )
 
+// defaultMaxCompletionCandidates is the maximum number of candidates
+// to send in one completion response (with isIncomplete = true)
+var defaultMaxCompletionCandidates = 100
+
 // 0.12.0 first introduced HCL2 which provides
 // more convenient/cleaner parsing
 //
@@ -28,6 +32,7 @@ const parserVersionConstraint = ">= 0.12.0"
 type parser struct {
 	logger *log.Logger
 
+	maxCandidates int
 	schemaReader schema.Reader
 }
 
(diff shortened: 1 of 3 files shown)


internal/terraform/lang/parser.go
```diff
@@ -115,12 +121,16 @@ func (p *parser) CompletionCandidatesAtPos(file ihcl.TokenizedFile, pos hcl.Pos)
...
 	bTypes := p.blockTypes()
 
-	list := &completeList{
+	list := &candidateList{
 		candidates: make([]CompletionCandidate, 0),
 	}
 
 	prefix := prefixAtPos(file, pos)
 	for name, t := range bTypes {
+		if len(list.candidates) >= p.maxCandidates {
```
paultyng, line 130: if the list was exactly max candidates, I think this is a minor bug?

radeksimko: I think that's correct, it's just probably slightly more difficult to read/understand, because it's checking the length with the addition from the previous iteration.
If the length is equal to max then this conditional doesn't even get executed, because we reached the end of the slice. I will add one more test to cover this scenario.

PR #133, opened 2020-06-03
textDocument/complete: Pass TextEdit instead of static text

radeksimko:
Depends on #132 

--- 

This seemingly small change has a positive effect on the UX while completing in the middle of an existing token, where clients can correctly highlight parts of completion candidates.

### Before

Screenshot 2020-06-03 at 09 08 42

Screenshot 2020-06-03 at 09 14 06

### After

Screenshot 2020-06-03 at 09 08 06

Screenshot 2020-06-03 at 09 13 34

--- 

As mentioned in the comment in code, this doesn't work (yet) for block types (e.g. `resource`, `data`, `provider`). It probably should, but some refactoring is necessary to make that work and realistically changing of block type is far less likely to occur in real scenarios, so it's ok to take on that tech debt.

### DIFF
--- a/internal/terraform/lang/types.go
+++ b/internal/terraform/lang/types.go
@@ -76,8 +76,26 @@ type CompletionCandidate interface {
 	Label() string
 	Detail() string
 	Documentation() MarkupContent
-	Snippet() string
-	PlainText() string
+	Snippet() TextEdit
+	PlainText() TextEdit
+}
+
+type TextEdit interface {
+	Range() *hcl.Range
+	NewText() string
+}
+
+type textEdit struct {
+	newText string
+	rng     *hcl.Range
+}
+
+func (te *textEdit) Range() *hcl.Range {
+	return te.rng
+}
+
+func (te *textEdit) NewText() string {
+	return te.newText
 }
 
 // MarkupContent reflects lsp.MarkupContent
--- a/internal/terraform/lang/utils.go
+++ b/internal/terraform/lang/utils.go
@@ -18,18 +18,18 @@ func PosInLabels(b *hclsyntax.Block, pos hcl.Pos) bool {
 	return false
 }
 
-func prefixAtPos(looker TokenAtPosLooker, pos hcl.Pos) string {
+func prefixAtPos(looker TokenAtPosLooker, pos hcl.Pos) (string, *hcl.Range) {
 	token, err := looker.TokenAtPosition(pos)
 	if err != nil {
-		return ""
+		return "", nil
 	}
 
 	switch token.Type {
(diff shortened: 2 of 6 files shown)


internal/terraform/lang/utils.go
```diff
@@ -18,18 +18,18 @@ func PosInLabels(b *hclsyntax.Block, pos hcl.Pos) bool {
...
 	token, err := looker.TokenAtPosition(pos)
 	if err != nil {
-		return ""
+		return "", nil
 	}
 
 	switch token.Type {
 	case hclsyntax.TokenIdent, hclsyntax.TokenQuotedLit, hclsyntax.TokenStringLit:
-		return string(token.Bytes[:pos.Byte-token.Range.Start.Byte])
+		return string(token.Bytes[:pos.Byte-token.Range.Start.Byte]), token.Range.Ptr()
```
njuCZ, line 29: this is the range for prefix, so we should set the right end pos

njuCZ: maybe it depends on how we handle an attribute re-completion, it's OK to override the whole word

radeksimko: Yeah, that's a fair point. This is actually by design, it's just confusing that we're mixing prefix (left part of the token) and range of the whole token in one function.

I'll have a think about it.

internal/terraform/lang/types.go
```diff
@@ -76,8 +76,26 @@ type CompletionCandidate interface {
 	Label() string
 	Detail() string
 	Documentation() MarkupContent
-	Snippet() string
-	PlainText() string
+	Snippet() TextEdit
```
paultyng, line 79: I think this is fine for now, but probably a lot easier to just make this function `TextEdit(snippetSupport bool)` and then you can take the branch out downstream.

radeksimko: Hmm, I see your point, but also I feel that the two methods communicate the situation better than one with a boolean argument. It seems fairly cheap here to enforce the [single responsibility principle](https://blog.gfader.com/2010/06/avoid-boolean-parameters-in-method.html) compared to maintaining if statements in multiple places.

In theory compiling the snippets may become more expensive/complex for some blocks and complex code deserves its own function to retain some readability. 😉 

Maybe it's premature optimization, time will tell 🤷‍♂️

internal/terraform/lang/utils.go
radeksimko: I can't come up with a better pattern, so I'm going to merge it as is, but I'm open to ideas and PRs to address this in the future.

PR #147, opened 2020-06-08
add support for "prefill all required fields"

### LINKED ISSUE #96: feat: required arguments pre-filled when creating resource
It would be awesome to have pre-filled  required arguments when created new ressource.

njuCZ:
fix #96 

- add support for prefill all required fields. (only prefix is empty will trigger this)
- support for sortText and a little refactor of sort logic(to ensure the `prefill all fields` item could place first)

something maybe not good: 
- In original design, `CompletionCandidate` is immutable, to avoid adding a `setSortText` func, I add some functions in `sort.go`, but in such way, everytime we add a new candidate type, we should remember to add revelant code in this file.

image

### DIFF
--- a/internal/lsp/completion.go
+++ b/internal/lsp/completion.go
@@ -37,21 +37,23 @@ func CompletionItem(candidate lang.CompletionCandidate, pos hcl.Pos, snippetSupp
 	if snippetSupport {
 		return lsp.CompletionItem{
 			Label:            candidate.Label(),
-			Kind:             lsp.CIKField,
+			Kind:             lsp.CompletionItemKind(candidate.CompletionItemKind()),
 			InsertTextFormat: lsp.ITFSnippet,
 			Detail:           candidate.Detail(),
 			Documentation:    doc,
 			TextEdit:         textEdit(candidate.Snippet(), pos),
+			SortText:         candidate.SortText(),
 		}
 	}
 
 	return lsp.CompletionItem{
 		Label:            candidate.Label(),
-		Kind:             lsp.CIKField,
+		Kind:             lsp.CompletionItemKind(candidate.CompletionItemKind()),
 		InsertTextFormat: lsp.ITFPlainText,
 		Detail:           candidate.Detail(),
 		Documentation:    doc,
 		TextEdit:         textEdit(candidate.PlainText(), pos),
+		SortText:         candidate.SortText(),
 	}
 }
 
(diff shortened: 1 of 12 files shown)


njuCZ:
@radeksimko sorry to say I don't see smiliar features in other language server. I mainly refer to the UX of Goland.
image

Personally, I think both ways are OK. I just think if users copy paste the resource type, they could also benefit from this feature

radeksimko:
I couldn't find any other precedents myself, but I have opened https://github.com/microsoft/language-server-protocol/issues/1057 to see what the LSP community has to say about this.

I am also curious if Goland's API has particular name for the "kind" of this completion meta items? I understand that this kind of data doesn't come from any language server, but they should have a name for it, which we could correlate with LSP's `CompletionItemKind`?

ahmednrana:
This would be highly useful.

radeksimko:
I do agree this would be useful, but given that UX/UI is the only form of contract that we have with the user, I'd like it to be right as it's something we ideally shouldn't "break". Implementation details can always be changed behind the scenes without user noticing, but UI cannot (as easily) unfortunately once we commit to something.

I did [ask the LSP community](https://github.com/microsoft/language-server-protocol/issues/1057) as mentioned and the conclusion so far seems that this likely isn't the way we should implement it.

[I'm still waiting for last confirmation](https://github.com/microsoft/language-server-protocol/issues/1057#issuecomment-668122950), but if at all this should be implemented via `textDocument/completion`, then it should be probably closer to the UI originally discussed in #96 but using `additionalTextEdits`. Implementing it the way this PR implements would likely be seen as an abuse of the protocol and as such I would like to avoid merging it in the current form.

With that said I do appreciate all the effort you invested into this @njuCZ and it's useful to start the conversation.

radeksimko:
As per conversation in https://github.com/microsoft/language-server-protocol/issues/1057#issuecomment-668122950 I'd prefer this to be solved via AdditionalTextEdits, rather than a dedicated "meta-candidate".

Just FYI - as of recently the completion logic has been generalized (to work for any HCL2 language server) and decoupled into a separate library and so changes would need to be done there.

Here's probably where I would start if I was to implement it:
https://github.com/hashicorp/hcl-lang/blob/4e412924f52b30a55243e9a754cbd313c544c60b/decoder/label_candidates.go#L44
and somewhere we could also implement the same functionality for nested required fields (i.e. optional nested block with required attributes):
https://github.com/hashicorp/hcl-lang/blob/main/decoder/block_candidates.go

btw. the solution should probably also reflect "required" blocks (those which have `MinItems >= 1`).

As for dealing with the edge case where the block may already have existing attributes in the body I would suggest we just replace it, i.e. the range for the text edit can be the whole body of a block. It looks like the latest VSCode version at least gives user a way of overriding that behaviour anyway:
https://code.visualstudio.com/updates/v1_51#_status-bar-for-suggestions

PR #149, opened 2020-06-09
terraform: Support 0.13 version

### LINKED ISSUE #111: Make LS compatible with Terraform v0.13
### Current Version

```
v0.2.1
```

### Use-cases

Terraform 0.13 will be released soon and users should be able to use LS with that version.

### Proposal

One known issue is different location of the lock file, which is currently set here

https://github.com/hashicorp/terraform-ls/blob/b0f90d054aae8837759ab2fb740f4dd93849f738/internal/terraform/schema/watcher.go#L46-L52

the new location is apparently `.terraform/plugins/selections.json` - which seems platform-agnostic, so that's positive, but we may not be able to just automatically look for that new path based on Terraform version, because `v0.13` doesn't seem to be moving the old lock file to the new location.

radeksimko:
This relaxes the version constraints as we now know where to find the lock file in 0.13+.

This also enables support for any 0.12 prereleases.

Closes #111 

I removed the upper bound constraint from the watcher as I reckon/hope that the lock file location won't change - and if it does, then the worst case scenario is that we will report the workspace as uninitialised.

I assume we may start constraining the parser at some point, but that's a problem for another day.

### DIFF
--- a/internal/terraform/lang/parser.go
+++ b/internal/terraform/lang/parser.go
@@ -37,10 +37,19 @@ type parser struct {
 }
 
 func ParserSupportsTerraform(v string) error {
-	tfVersion, err := version.NewVersion(v)
+	rawVer, err := version.NewVersion(v)
 	if err != nil {
 		return err
 	}
+
+	// Assume that alpha/beta/rc prereleases have the same compatibility
+	segments := rawVer.Segments64()
+	segmentsOnly := fmt.Sprintf("%d.%d.%d", segments[0], segments[1], segments[2])
+	tfVersion, err := version.NewVersion(segmentsOnly)
+	if err != nil {
+		return fmt.Errorf("failed to parse stripped version: %w", err)
+	}
+
 	c, err := version.NewConstraint(parserVersionConstraint)
 	if err != nil {
 		return err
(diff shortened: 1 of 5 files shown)


internal/terraform/lang/parser.go
```diff
@@ -37,10 +37,19 @@ type parser struct {
...
 func ParserSupportsTerraform(v string) error {
-	tfVersion, err := version.NewVersion(v)
+	rawVer, err := version.NewVersion(v)
 	if err != nil {
 		return err
 	}
+
+	// Assume that alpha/beta/rc prereleases have the same compatibility
+	segments := rawVer.Segments64()
+	segmentsOnly := fmt.Sprintf("%d.%d.%d", segments[0], segments[1], segments[2])
```
radeksimko, line 47: Perhaps it would be useful to upstream these two lines into `go-version` as `Version.RawVersion()` or something like that. Not something I'd want to block this PR on, but thinking out loud.

paultyng: Yeah, I think we should just start implementing this in terraform-exec and see what feels right there. I'd rather not have to have a lot of this stuff in different packages we need to think/worry about and terraform-exec will handle schema dumping, we could also add lock file watching to it, etc.

Issue #106, opened 2020-05-21
Ask to `terraform init` an uninitialized workspace


paultyng:
https://microsoft.github.io/language-server-protocol/specification#window_showMessageRequest

> The show message request is sent from a server to a client to ask the client to display a particular message in the user interface. In addition to the show message notification the request allows to pass actions and to wait for an answer from the client.

Not entirely sure about the UX on this, but may be worth exploring.

paultyng:
We should probably ask to initialize entirely uninitialized workspaces or when we get that error from the Terraform binary. But we may be able to be a bit proactive about it, when there are providers in source that are not initialized or maybe the local mirror directories change, etc.

radeksimko:
I think this might require _some_ help from the language server in relation to https://github.com/hashicorp/terraform-ls/issues/32 because the server will likely be the one that knows better/best _where_ it's suitable to `terraform init` a particular directory.

BertelBB:
Is it possible to disable the initialization?
It is breaking the _hashicorp.terraform_ extension for me because I am opening a workspace in VSCode with multiple Terraform projects and the extension is trying to perform the initialization in the root directory of the workspace, which is not a Terraform project.

radeksimko:
The error will not be reported as per https://github.com/hashicorp/terraform-ls/pull/171 from the next release (0.4.0; to be released in coming days). The schema is still necessary for most functionality (pretty much all except formatting), but we need to better understand where in the hierarchy could a folder be potentially `init`d, so we can hint this to the user - as opposed to just expect it in any directory that is being opened - that's what this issue is mainly about.

Supporting monorepo and other more complex hierarchies is something we plan to support as part of #32

radeksimko:
As for the implementation details of this one, I realized that [the JSON RPC library](https://github.com/creachadair/jrpc2) we use doesn't seem to support server->client requests. It can only send notifications (without waiting for response), such as [`window/showMessage`](https://microsoft.github.io/language-server-protocol/specification#window_showMessage) via [`jrpc2.ServerPush()`](https://pkg.go.dev/github.com/creachadair/jrpc2?tab=doc#ServerPush).

Issue #83, opened 2020-05-06
Treat schema availability as not essential


radeksimko:
### Current Version

```
0.1.0
```

### Use-cases

Sometimes the user may open a directory which has not been `terraform init`d - which is then surfaced as `initialize` error and the server doesn't start at all as unavailability of schema is treated as essential for its functioning.

We could still provide limited functionality that doesn't require schema - e.g. formatting and make schema watching/obtaining non-essential.

### Proposal

Provide limited functionality if schema isn't available and inform the user of the limited functionality via diagnostics.

We should also figure out how/whether we can detect when user runs `terraform init` for the first time, so they don't have to reopen the whole folder/IDE after doing so. This practically depends on whether we can watch file/dir creations in a given path.

### References

https://github.com/hashicorp/terraform-ls/pull/84

paultyng:
Should we just init with `-backend=false` for them?

radeksimko:
As per our offline discussion - I believe that:

1. if any side executes `init` it should be client, through [`workspace/executeCommand`](https://github.com/hashicorp/terraform-ls/issues/81), not the server directly
2. a [good client](https://knowyourmeme.com/memes/theyre-good-dogs-brent) should not be downloading hundreds of MB on the background _without the user knowing_
   - I'd be ok with a client config option which turns auto-`init` on and off, but that should certainly ship as off by default

radeksimko:
I just noticed that this is also important for when the client for any reason triggers `initialize` for a directory that isn't terraform at all - apparently happens with our private build of the VSCode extension currently.

It will just `initialize` whatever project is currently active in VSCode, which it probably shouldn't, but the server also shouldn't just error out in such case.

paultyng:
Not sure, but this is possibly a duplicate of or at least related to #106

devTechi:
Hi,
for me this extensions stopped somehow working due to this issue (and others).

We aren't using `terraform` locally at all. So no possibility to do a local `terraform init`. And `terraform init -backend=false` gave me errors since we are still using on version 0.11.* in some repositories.

radeksimko:
Hi @devTechi 
I understand it can be hard to understand what functionality is/was provided by the client (e.g. VSCode extension) or the server, but when it comes to VSCode settings, that's most likely client side.

[...]

Issue #32, opened 2020-03-22
Support single files and parent (non-Terraform) folders


radeksimko:
## Single files

A restriction is currently in place that prevents the Language Server from communicating with an IDE (client) that opened a single file. As a result it enforces the user to open whole folders.

This is done by failing initialization with the following error:

```
Editing a single file is not yet supported. Please open a directory.
```

## Why

The restriction prevents the LS from having to deal with the potentially increased complexity in any logic that concerns the relationship between a file and its plugins (providers).

The LS has responsibility for:

 - finding compatible Terraform binary (happens during `initialize` via `$PATH`)
 - finding and storing Terraform version (happens during `initialize`)
 - retrieving and caching schema for `init`ed providers (happens during `initialize`)
 - invalidating the cache and retrieving it again when providers change (happens any time via watching the lock file in a plugin folder)

The `initialize` method is called only **once** for every opened folder, which is mainly where the lower complexity comes from.

## Parent (non-Terraform) folders

This also affects users who wish to open a folder higher up the filesystem hierarchy, e.g. a folder with all Terraform workspaces, as opposed to opening workspaces individually. We do not support such case today either, but don't actually raise any error in this case (yet?).

The same reasoning and complexity scope applies here too though and it's likely that both use cases have the same (or very similar) solution.

### Local Modules

Users with locally stored modules (where `module`'s `source` is path in a local filesystem) may fall into this category as such modules tend to be _not_ `init`ed within their own folder (i.e. such modules often don't have their own `.terraform` folder). Supporting them will rely heavily on how their provider inheritance is set up - i.e. we need to understand where can we get the provider schemas from for every module.

## Future

If we support single files, we need to consider the above happening at different times (probably during `didOpen`) and significantly more often. Hence we need to ensure this scales well in terms of resource consumption - e.g. by ensuring that we never watch/refresh the same plugin directory twice within the context of a running server process. We may also need to decouple the Terraform binary discovery and version handling, to ensure we only do it once per workspace.

## Related

This limitation also helps us avoid bugs similar to https://github.com/juliosueiras/terraform-lsp/issues/58

## Proposal

 - walker: introduce more limited walking mode, such that it doesn't descend into lower directories (to prevent unexpected outcomes if user opens `~` home dir)
 - initialize handler: avoid indexing if `rootUri` is empty (single file was open) and store a flag to say "LS is in single file mode"
[...]

Issue #75, opened 2020-04-29
Support different OS/paths between client and server


radeksimko:
### Current Version

```
0.1.0
```

### Use-cases

Currently the server makes assumptions for Windows based on build flags, which doesn't account for situations where client may be running a Unix-based filesystem for any reason and send Unix paths and vice versa.

The server will in such case likely misinterpret file paths.

### Proposal

Detect Windows-style paths regardless of where the server is running.

### References

https://github.com/hashicorp/terraform-ls/pull/73

PR #167, opened 2020-06-16
refactoring: Introduce root module (manager) abstraction

radeksimko:
Sorry for the bigger PR! 😅 

--- 

The main theme of the refactoring is to decouple a few concepts:
 - watcher
 - schema storage
 - terraform executable discovery
 - terraform executor

and introduce a new concept of root module manager which better reflects the reality where each root module (directory) can have its own schema/plugin cache, its own terraform version/executable and its own module cache.

This also introduces module cache parser and implements cache invalidation when the module cache changes.

This doesn't yet solve, but introduces a more solid foundation for solving the following issues:

 - https://github.com/hashicorp/terraform-ls/issues/32
 - https://github.com/hashicorp/terraform-ls/issues/83
 - https://github.com/hashicorp/terraform-ls/issues/106
 - https://github.com/hashicorp/vscode-terraform/issues/396

The immediate end-user effect of this PR is the following:
 - if an `init`d directory is opened as root by client, the server obtains the schema and makes it available for any referenced local modules
 - if client opens any referenced module, it gets schema data from the root module
 - if client opens module which is not referenced, the completion errors out with `"finding compatible parser failed: root module not found for .../nested-path"`

Technically this scenario seemed to work before too, but that was rather a side effect of the implementation where we had a single global plugin/schema cache and we would (wrongly) also provide this data to modules which were _not_ referenced. So this PR effectively makes it _unavailable_ for modules which are not actually referenced.

The aim of this initial refactoring PR is to keep the functional changes minimal, but follow-up with more smaller PRs afterwards which can:

 - make initialization of plugin cache optional (that will require some changes in `langserver/handlers` mocks - we'll probably need to add some real test data in form of folders with `.terraform`)
 - walk the hierarchy to reflect monorepo with multiple root modules
 - decide how to best match these multiple schema/plugin caches with module paths


go.sum
```diff
@@ -32,6 +32,8 @@ github.com/hashicorp/errwrap v1.0.0 h1:hLrqtEDnRye3+sgx6z4qVLNuviH3MR5aQ0ykNJa/U
 github.com/hashicorp/errwrap v1.0.0/go.mod h1:YH+1FKiLXxHSkmPseP+kNlulaMuP3n2brvKWEqk/Jc4=
 github.com/hashicorp/go-multierror v1.0.0 h1:iVjPR7a6H0tWELX5NxNe7bYopibicUzc7uPribsnS6o=
 github.com/hashicorp/go-multierror v1.0.0/go.mod h1:dHtQlpGsu+cZNNAkkCN/P3hoUDHhCYQXV3UM06sGGrk=
```
paultyng, line 34: This looks like you need a `tidy`

paultyng:
Just a general observation as I read through, you may want to append an `_test` to all the mock files, to ensure they aren't included at runtime.

internal/terraform/workspace/workspace.go
```diff
@@ -0,0 +1,284 @@
...
+
+	"github.com/hashicorp/terraform-ls/internal/terraform/discovery"
+	tferr "github.com/hashicorp/terraform-ls/internal/terraform/errors"
+	"github.com/hashicorp/terraform-ls/internal/terraform/exec"
+	"github.com/hashicorp/terraform-ls/internal/terraform/lang"
+	"github.com/hashicorp/terraform-ls/internal/terraform/schema"
+)
+
+type workspace struct {
+	ctx            context.Context
```
paultyng, line 23: this seems like an issue to attach a context to a struct, you shouldn't need to

internal/terraform/workspace/workspace.go
```diff
@@ -0,0 +1,284 @@
...
+	if w.moduleCache == nil {
+		return false
+	}
+
+	for _, m := range w.moduleCache.Modules {
+		if m.Key == "" {
+			// skip root module, as that's tracked separately
+			continue
+		}
+		if !strings.HasPrefix(m.Source, "."+string(os.PathSeparator)) {
```
paultyng, line 219: Not sure if this is accurate as I'm not sure what the non-relative paths look like in all cases, but this may be a better check:

```suggestion
		if filepath.IsAbs(m.Source) {
```

internal/terraform/workspace/workspace.go
```diff
@@ -0,0 +1,284 @@
...
+			// skip external modules, for now
+			continue
+		}
+		absPath := filepath.Join(w.moduleCache.rootDir, m.Dir)
+		w.logger.Printf("checking if %q == %q", absPath, path)
+		if absPath == path {
```
paultyng, line 225: If you are going to do a string comparison here, you probably want to use `filepath.Clean`. Not sure if this still leaves an edge for symlinks though.

internal/terraform/workspace/workspace_manager.go
```diff
@@ -0,0 +1,161 @@
...
+	"time"
+
+	"github.com/hashicorp/terraform-ls/internal/terraform/discovery"
+	"github.com/hashicorp/terraform-ls/internal/terraform/exec"
+	"github.com/hashicorp/terraform-ls/internal/terraform/lang"
+	"github.com/hashicorp/terraform-ls/internal/terraform/schema"
+)
+
+type workspaceManager struct {
+	ctx           context.Context
```
paultyng, line 18: Shouldn't attach `context.Context` to structs

internal/terraform/workspace/workspace_manager.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	return w.TerraformExecutor(), nil
+}
+
+// workspaceDirFromPath strips known lock file paths and filenames
+// to get the directory path of the relevant workspace
+func workspaceDirFromFilePath(filePath string) string {
+	for _, p := range pluginLockFilePaths("/") {
+		if strings.HasSuffix(filePath, p) {
```
paultyng, line 139: `filepath.Rel`

paultyng, line 139: You may want to `filepath.Clean` here prior to doing a string comparison

internal/terraform/workspace/workspace_manager.go
```diff
@@ -0,0 +1,161 @@
...
+			return strings.TrimSuffix(filePath, p)
+		}
+	}
+
+	mlfp := moduleLockFilePath("/")
+	if strings.HasSuffix(filePath, mlfp) {
```
paultyng, line 145: Same as above, may need `filepath.Clean`

internal/terraform/workspace/workspace_manager.go
```diff
@@ -0,0 +1,161 @@
...
+		return err
+	}
+
+	wc.ws[dir] = w
+
+	return nil
+}
+
+func (wc *workspaceManager) WorkspaceByPath(path string) (Workspace, error) {
+	if w, ok := wc.ws[path]; ok {
```
paultyng, line 96: You may want to normalize the keys with a `filepath.ToSlash` or something here.

go.sum
radeksimko: it's actually forgotten `go mod vendor` 🙈

radeksimko:
> Just a general observation as I read through, you may want to append an `_test` to all the mock files, to ensure they aren't included at runtime.

I was thinking about it ☝️ , but then I also realized that the mock logic needs to be usable outside of the package and AFAIK anything in tests isn't exported. On the ~positive~ other side none of these mock functions/structs import `testing`, so they should be harmless to runtime.

internal/terraform/workspace/workspace.go
radeksimko: The reason for the persistence here is to allow watcher to ask for schema refresh without having to pass its own context. The watcher remains global (to help centralising paths we need to watch across the whole hierarchy) so it is designed to outlive a workspace, but operations on the workspace should be cancellable when e.g. workspace cache is removed (not possible today, but I don't see why not implement it later).

radeksimko: That said I don't like persisting it either, I just couldn't come up with a better solution.

internal/terraform/workspace/workspace_manager.go
radeksimko: Fair - I think I can just create a fresh context for new workspaces instead.

internal/terraform/workspace/workspace_manager.go
radeksimko: Good point, but I think it's better to do it higher up in `WorkspaceByPath`. The `pluginLockFilePaths` shouldn't need cleaning up as it's a static path we have control over, so messy path in form of double `//` etc. can't really appear there.

internal/terraform/workspace/workspace_manager.go
radeksimko: As mentioned above - probably best to clean `filePath` once before even entering this function, instead of "re-cleaning" it downstream. The `moduleLockFilePath` also shouldn't require cleaning for the same reason as mentioned above (it's entirely static path just with OS-specific separators).

internal/terraform/workspace/workspace_manager.go
radeksimko: 🤔   I don't disagree, but also I wonder when would that matter? Are you thinking of https://github.com/hashicorp/terraform-ls/issues/75 or https://github.com/hashicorp/terraform-ls/issues/74 ?

internal/terraform/workspace/workspace.go
radeksimko: True, I think the right side (`path`) can be cleaned up in the manager's `WorkspaceByPath`, but no harm in cleaning it up here again I guess.

The left part certainly needs cleaning here though. I need to better understand what paths we may expect here first. I'm not sure if Terraform itself does any cleaning and/or symlink eval before storing these paths in the cache.

internal/terraform/workspace/workspace.go
radeksimko: Yeah - that's a good question - I need to try out a few scenarios to see what paths to expect in the cache.

internal/terraform/workspace/workspace_manager.go
radeksimko: It turns out module paths are [always persisted with Unix-style separators](https://github.com/hashicorp/terraform/blob/bc386234d528747ad244574f72f1fc30a486d5c2/internal/modsdir/manifest.go#L131-L134) so good catch - this probably wouldn't work on Windows! 🙈

internal/terraform/workspace/workspace.go
radeksimko: I added more explicit `IsExternal()` to the module record instead as that is cleaner and the logic there is also slightly different as this one was not as reliable - `Source` can also have `../` for modules which are local to some other modules.

internal/terraform/workspace/workspace_manager.go
radeksimko: I added some logic to account for module paths being stored with forward slashes, but I think I'd rather avoid converting it back and forth when comparing the paths, for now.

That said I'm not opposed to implementing it later - that is also why the `map[string]*workspace` is an implementation detail, rather than something directly exposed.

internal/terraform/workspace/workspace_manager.go
radeksimko: I added `filepath.Clean()` to both `WorkspaceByPath` and `AddWorkspace` and also to `TrackedFile` constructor, so that the path is cleaned on entry _and_ on lookup. It does _not_ follow symlinks yet though - as `filepath.EvalSymlink` requires real directories/files as test data, so fabricated paths are no longer good enough in tests.

internal/terraform/workspace/workspace.go
radeksimko: `w.moduleCache.rootDir` is cleaned as part of `TrackedFile` constructor, `m.Dir` as part of manifest parser (although Terraform should produce clean paths in general) and the `path` is cleaned in `WorkspaceByPath` and `AddWorkspace`.

internal/lsp/file_handler.go
```diff
@@ -60,6 +95,13 @@ func (fh *versionedFileHandler) Version() int {
...
 
-func FileHandlerFromPath(path string) FileHandler {
-	return FileHandler(filesystem.URIFromPath(path))
+func FileHandlerFromPath(path string) *fileHandler {
+	return &fileHandler{uri: filesystem.URIFromPath(path)}
+}
+
+func FileHandlerFromDirPath(dirPath string) *fileHandler {
+	// Dir URIs are usually without trailing separator in LSP
+	dirPath = strings.TrimSuffix(dirPath, string(os.PathSeparator))
```
paultyng, line 104: I think you could just simplify this as `filepath.Clean` unless you explicitly don't want it to resolve `.` and `..` references?

```suggestion
	dirPath = filepath.Clean(dirPath)
```

PR #171, opened 2020-06-18
Treat schema availability as not essential

### LINKED ISSUE #83: Treat schema availability as not essential

radeksimko:
Closes #83 

This does *not* yet introduce support for more complex hierarchies
but enables formatting even for files in directories
which do not have any plugin cache.

The lack of plugin cache is now emitted just as a log message.

Informing the user at this point wouldn't be very helpful
because we don't know where the module usually gets init'd from yet,
i.e. whether the plugin cache is even expected *there*.

Complex hierarchies will be addressed in other PRs in iterative steps as part of #32

### DIFF
--- a/internal/terraform/rootmodule/root_module_manager.go
+++ b/internal/terraform/rootmodule/root_module_manager.go
@@ -39,20 +39,20 @@ func newRootModuleManager(ctx context.Context) *rootModuleManager {
 }
 
 func (rmm *rootModuleManager) defaultRootModuleFactory(ctx context.Context, dir string) (*rootModule, error) {
-	w := newRootModule(ctx)
+	rm := newRootModule(ctx)
 
-	w.SetLogger(rmm.logger)
+	rm.SetLogger(rmm.logger)
 
 	d := &discovery.Discovery{}
-	w.tfDiscoFunc = d.LookPath
-	w.tfNewExecutor = exec.NewExecutor
-	w.newSchemaStorage = schema.NewStorage
+	rm.tfDiscoFunc = d.LookPath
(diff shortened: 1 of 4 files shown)


internal/terraform/rootmodule/root_module_manager.go
```diff
@@ -78,15 +78,15 @@ func (rmm *rootModuleManager) AddRootModule(dir string) error {
 
 	_, exists := rmm.rms[dir]
 	if exists {
-		return fmt.Errorf("rootModule %s was already added", dir)
+		return fmt.Errorf("root module %s was already added", dir)
```
paultyng, line 81: Not sure if it matters, but should this just be a noop instead of an error?

radeksimko: I suppose it may come down to it eventually, but I think we do not ever add the same module twice, and if we are then I'd like to understand why first, before allowing it.

PR #176, opened 2020-06-22
Walk hierarchy to add root modules

radeksimko:
This aims to solve some scenarios outlined in #32 

--- 

Because detection of the relevant root module for a file/dir is a non trivial task, I also introduce candidate finder which can report all potential candidates for root modules in the context of a directory being initialized/opened.

The following factors may play role in deciding about the right root module:

### Whether the root module was initialized at all (REQUIREMENT)

This is a hard requirement, at least for now. i.e. Root module has to be `init`d before it's even considered as a candidate (i.e. it has to have `.terraform`). The module cache provides enough details about module hierarchy in a very efficient and simple way.

### Plugin Version Proximity (NOT IMPLEMENTED)

https://github.com/hashicorp/terraform-ls/issues/180

### Hierarchy Proximity (NOT IMPLEMENTED)

https://github.com/hashicorp/terraform-ls/issues/179

--- 

Before the above detection is implemented we just return the first positive match (in whatever order `filepath.Walk` finds it) and display a warning in case there is more than 1 match.

The only awkward thing is that every IDE implements `windows/showMessage` differently. I assume they just work within the constraints of the client API 🤷 

### VSCode

Screenshot 2020-06-24 at 07 39 38

### Sublime Text

Screenshot 2020-06-24 at 07 47 37

### NeoVim

Screenshot 2020-06-24 at 07 47 04


FernandoMiguel:
not all my projects have a `main.tf` file. 
but any tf project will have 
```hcl
terraform {
  backend "s3" {}
}

```
block

radeksimko:
@FernandoMiguel Just to be clear the example I mentioned above isn't assuming any particular filenames. It's just an example.

I will lock this thread temporarily until we iron out the implementation details as it's not quite ready yet for wider public consumption.

Generally though feedback related to different hierarchies is welcomed in #32 

Thanks.

radeksimko:
This should provide some details about what kind of hierarchies will be supported when this PR is merged:
https://github.com/hashicorp/terraform-ls/tree/f-root-module-discovery/internal/terraform/rootmodule/testdata

internal/terraform/rootmodule/types.go
```diff
@@ -29,6 +33,7 @@ type RootModuleManager interface {
 	AddRootModule(dir string) error
 	PathsToWatch() []string
 	RootModuleByPath(path string) (RootModule, error)
+	RootModuleCandidatesByPath(path string) []string
```
paultyng, line 36: You could probably just embed the other interface to keep these in sync I think?

internal/terraform/schema/schema_storage.go
```diff
@@ -140,10 +140,10 @@ func (s *Storage) obtainSchemasForModule(tf *exec.Executor, dir string) error {
 	start := time.Now()
 	ps, err := tf.ProviderSchemas()
 	if err != nil {
-		return fmt.Errorf("Unable to retrieve schemas: %s", err)
+		return fmt.Errorf("Unable to retrieve schemas for %q: %s", dir, err)
```
paultyng, line 143:
```suggestion
		return fmt.Errorf("Unable to retrieve schemas for %q: %w", dir, err)
```

langserver/handlers/did_open.go
```diff
@@ -15,5 +19,51 @@ func TextDocumentDidOpen(ctx context.Context, params lsp.DidOpenTextDocumentPara
...
+func renderCandidates(rootDir string, candidatePaths []string) string {
+	for i, p := range candidatePaths {
+		// This helps displaying shorter, but still relevant paths
+		candidatePaths[i] = renderCandidate(rootDir, p)
+	}
+	return strings.Join(candidatePaths, ", ")
+}
+
+func renderCandidate(rootDir, path string) string {
+	return strings.TrimPrefix(
```
paultyng, line 67: I wonder if there is a better `filepath` func to do this to get the relative path? String manipulation feels wrong but maybe its fine?

radeksimko: Yeah, it does not feel quite right, but I couldn't find `filepath.TrimPrefix` or anything similar.

PR #178, opened 2020-06-24
Make formatting work regardless of initialization state

### LINKED ISSUE #165: Formatting should work regardless of workspace/initialization state
### Expected Behavior

Document and range formatting should work regardless of initialization state or presence of TF files in the root directory.

### Actual Behavior

Formatting does not function in those cases.

radeksimko:
This _actually_ fixes #165

### DIFF
--- a/internal/terraform/rootmodule/errors.go
+++ b/internal/terraform/rootmodule/errors.go
@@ -14,3 +14,11 @@ func (e *RootModuleNotFoundErr) Error() string {
 	}
 	return "root module not found"
 }
+
+func IsRootModuleNotFound(err error) bool {
+	if err == nil {
+		return false
+	}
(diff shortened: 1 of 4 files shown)


internal/terraform/rootmodule/errors.go
```diff
@@ -14,3 +14,11 @@ func (e *RootModuleNotFoundErr) Error() string {
 	}
 	return "root module not found"
 }
+
+func IsRootModuleNotFound(err error) bool {
```
paultyng, line 18: These should either wrap or just be replaced by [`errors.Is`](https://godoc.org/errors#Is) but doesn't need to happen in this PR.

radeksimko: Happy to address that in a follow-up PR.

PR #181, opened 2020-06-24
Add verbose output and cover info in CI tests

paultyng:


paultyng:
Yeah I agree, the change is the valuable bit, but can at least manually review it in the meantime.

PR #187, opened 2020-06-25
Surface root module init errors on didOpen

radeksimko:
I am marking this as draft for now, because I the format of errors it would produce is ... _suboptimal_ due to the fact that error output from `terraform providers schema -json` is just not designed to be displayed anywhere else outside of the terminal (even after stripping ANSI colour codes).

Screenshot 2020-06-25 at 10 11 08

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/hashicorp/terraform-ls
 go 1.13
 
 require (
+	github.com/acarl005/stripansi v0.0.0-20180116102854-5a71ef0e047d
 	github.com/apparentlymart/go-textseg v1.0.0
 	github.com/creachadair/jrpc2 v0.8.1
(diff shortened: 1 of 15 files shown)


aeschright:
Oh yeah, that's a lot to read in a popup window. Are you thinking we could do some cleanup on the language server side of this?

radeksimko:
Yes, I talked to @kmoe about this too and she mentioned she's tackling similar problem in https://github.com/hashicorp/terraform-exec by essentially regexp-ing known error messages and re-interpreting them as custom error types which have more "consumable" `String()`.

I think that might be good enough solution in the short run, until Terraform exposes this in a more friendly (JSON) format.

radeksimko:
Closing this in favour of https://github.com/hashicorp/terraform-ls/issues/218

Issue #195, opened 2020-06-30
Cancel root module operations on shutdown


radeksimko:
### Server Version

```
0.4.0
```

### Expected Behavior

On shutdown of the server (e.g. initiated by quitting the IDE) it should cancel any operations running on the root module, such as:
 - `terraform version`
 - obtaining schema via `terraform providers schema -json`
 - parsing module manifest

### Actual Behavior

Terraform executions will get cancelled either way, because they either inherit the session context or watcher context, both of which get cancelled as part of the shutdown:

https://github.com/hashicorp/terraform-ls/blob/c3bdf230ac89886a078767de545cd9aeaff94331/langserver/handlers/service.go#L224-L230

A new challenge arises [once we make walking asynchronous](https://github.com/hashicorp/terraform-ls/issues/188).
While the walker itself will be cancellable, the operations launched by the walker such as adding new root modules (and then performing any of the above) will retain context of the session and therefore can't be cancelled as part of walker cancellation.

This may result in multiple seconds of delay on shutdown while waiting for terraform to finish executing.

### Further Context

Each root module gets its own non-cancellable context at the moment:

https://github.com/hashicorp/terraform-ls/blob/c3bdf230ac89886a078767de545cd9aeaff94331/internal/terraform/rootmodule/root_module_manager.go#L83

Generally the context-related logic in `rootmodule` needs some overhaul to avoid persisting context.

This may require some more explicit workflow management, explicit cancellation of each individual `RootModule`, and perhaps explicit cancellation of `RootModuleManager` (which will in turn cancel and remove any tracked root modules).

In other words we should only be persisting `context.CancelFunc` for each root module, but not the full context.

PR #196, opened 2020-06-30
terraform/rootmodule: Make walker async by default

### LINKED ISSUE #188: Walk hierarchy asynchronously
### Current Version

```
v0.4.0
```

### Use-cases

Currently the whole hierarchy is walked through to discover root modules on initialization synchronously:

https://github.com/hashicorp/terraform-ls/blob/e40f7c5b2f411caa1aeec214e88cbc888d667ef0/langserver/handlers/initialize.go#L58-L66

This can result in:

1. delay in adding paths to the watcher
2. delayed initial response to the client, which means the client may refuse to provide any functionality until `initialize` request is responded to

### Proposal

Run walker asynchronously.

It should still be cancellable per session though.

radeksimko:
Closes #188 

There is some more work needed around cancellation (some tech debt is now a little more exposed as a result of this PR 😄 ), but that's resolvable separately as part of https://github.com/hashicorp/terraform-ls/issues/195

### DIFF
--- a/internal/context/context.go
+++ b/internal/context/context.go
@@ -30,6 +30,7 @@ var (
 	ctxParserFinder     = &contextKey{"parser finder"}
 	ctxTfExecFinder     = &contextKey{"terraform exec finder"}
 	ctxRootModuleCaFi   = &contextKey{"root module candidate finder"}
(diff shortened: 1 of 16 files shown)


radeksimko:
Agreed about `-race`, I have opened #206

PR #198, opened 2020-07-01
settings: Make root modules configurable

### LINKED ISSUE #189: Allow user to override discovery of root modules
### Current Version

```
0.4.0
```

### Use-cases

Currently the language server will walk through the hierarchy on `initialize` and attempt to discover all initialized root modules and obtain the schema from all, so it can later be matched to files being opened in the editor.

There are scenarios where this may not be sufficient:

 - User has multiple matching root modules and the one being automatically matched is not the one user wants to be matched
   - We intend to make the detection more accurate as per https://github.com/hashicorp/terraform-ls/issues/179 and https://github.com/hashicorp/terraform-ls/issues/180 but getting it right may take time
 - The root module in question is above the folder opened in the editor as root

### Proposal

Allow the user to override the discovery mechanism by defining a list of paths to root modules.

This could be provided in the form of:

 - CLI flags (e.g. `-root-module=/path/one -root-module=/path/two` etc.)
 - LSP settings - this has an added benefit of reacting to changes at runtime (adding/removing root module without having to restart the server)

### Related LSP methods

https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workspace_configuration
https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workspace_didChangeConfiguration

### References
https://github.com/hashicorp/vscode-terraform/issues/396

radeksimko:
Depends on #196

Closes #189

Related: https://github.com/hashicorp/vscode-terraform/pull/423


docs/SETTINGS.md
```diff
@@ -0,0 +1,44 @@
...
+				"rootModulePaths": ["/any/path"]
+			},
+		}
+	}
+}
+```
+
+### VS Code
+
+Use `terraform-ls.options`, e.g.
```
radeksimko, line 36: This may need to be tweaked depending on the name we actually decide upon and ship, but if I had to choose between `configuration`, `options` or `settings`, I would choose `options`, because that reflects the naming convention in the LSP and also leaves some space in case we decide to add some more settings under `terraform-ls` which would _not_ be passed down directly to the server, but affect how it's launched or installed (e.g. version pinning comes to mind, opting into prereleases etc.).

radeksimko:
I just realized that this will add desired root modules synchronously and so if there's too many or any module has a lot of provider schemas to crunch, it may take time until completion etc. is available.

I'll look into making it asynchronous.

docs/SETTINGS.md
aeschright: I was thinking it might be easier for the user to manage if the setting isn't nested, i.e. `terraform-ls.rootModules` -- but I don't know how many options we would anticipate adding later.

radeksimko: > but I don't know how many options we would anticipate adding later.

I was thinking that anything version/installation/update related could go under this `terraform-ls` namespace? That way it's also unlikely to conflict with any terraform-lsp related settings, if we do decide to add any.

Also knowing how difficult/awkward it is at the moment to change/deprecate supported settings in VSCode and remembering the confusion from v1/v2 settings changes, I was thinking of applying extra caution here.

radeksimko:
The asynchronicity is being addressed in https://github.com/hashicorp/terraform-ls/pull/219 - this PR will just need to be rebased and tuned accordingly once #219 is merged.

docs/SETTINGS.md
radeksimko: Changed this to just `terraform-ls` as per our discussion earlier this week about how this makes it easier to expose the configuration in the UI.

internal/context/context.go
```diff
@@ -198,3 +199,15 @@ func RootModuleWalker(ctx context.Context) (*rootmodule.Walker, error) {
 	}
 	return w, nil
 }
+
+func WithRootModuleLoader(rml rootmodule.RootModuleLoader, ctx context.Context) context.Context {
```
appilon, line 203: Nit, make `ctx` first argument

radeksimko: Yeah, I don't disagree, but I think I'd prefer to change all of the `With*` functions at the same time and perhaps do that in a separate PR, if that's ok and follow the same pattern throughout the package for now?

PR #199, opened 2020-07-02
fix volume case issue in windows platform

njuCZ:
In windows platform, when initializing, the vscode sends root uri in this format `file:///c%3A/Users/user/workspace`, the volumn `c` is in lowercase. The volume in all root module path are in lowercase. 

But when there is any change for `pluginlockfile` or `modulemanifestfile`, the file path starts with upper case `C`, which will cause path not match using `==`, the language server could not get the latest  schema or modules info.

this is a similiar issue https://github.com/golang/go/commit/935faf3be200d6f469cccdbec27cee3741a97350. It seems windows path are case insentitive and linux is case sensitive

I think for now it's ok to only igore the volumn case, there seems no need to ignore case for the whole path.


njuCZ:
@radeksimko thanks for your suggestions, I have added the test case, please have a look again

internal/terraform/rootmodule/path.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"path/filepath"
+	"strings"
+)
+
+func pathEquals(path1, path2 string) bool {
+	volume1 := filepath.VolumeName(path1)
+	volume2 := filepath.VolumeName(path2)
+	return strings.EqualFold(volume1, volume2) && path1[len(volume1):] == path2[len(volume2):]
```
radeksimko, line 11: I don't know enough about filesystems on Windows myself to decide, but I suspect we may need to compare of the other part of the path in case-insensitive way too at some point. It can be addressed as and when needed though.

PR #202, opened 2020-07-03
Usage: Update VS Code extension instructions

radeksimko:


docs/USAGE.md
```diff
@@ -109,4 +109,11 @@ endif
...
 
-See our [Terraform VS Code Extension](https://github.com/hashicorp/vscode-terraform) repository.
+ - Install [Terraform VS Code Extension](https://marketplace.visualstudio.com/items?itemName=hashicorp.terraform) `>=2.0.0`
+ - Latest version of the language server should be installed automatically on VS Code launch
+ - In case you are upgrading from v1 you may need to remove `terraform.languageServer`
+ 	settings or just make it empty as shown below
+
+```json
+"terraform.languageServer": {}
+```
```
aeschright, line 119: I think they'll have to do that before upgrading, or disable and reenable the extension. The extension does do some cleanup if they have the previous default language server config.

radeksimko: AFAIK the extension prompts for reload when the settings is changed, so that would take care of it, wouldn't it?

aeschright: Yes, it will prompt to reload. What I'm not sure about is if/when the default will get reapplied.

radeksimko: Ah, I see - I can list the default settings there then. 😃

radeksimko: Updated the PR accordingly, PTAL.

Issue #128, opened 2020-06-01
v2.0.0 - aws authentication issue


garry-harthill-cko:
I believe the language server is reading the remote state. Remote state requires authentication.

I have a more elaborate method of authenticating against AWS which is a wrapper that rotates keys and sets environment variables, etc.

https://github.com/99designs/aws-vault

It would be a good feature if we would wrap the language servers execution in this program. Maybe a "terraform-ls.command" option so we can define the full terraform-ls command to be executed. eg:
`/usr/local/bin/aws-vault exec myProfile -- terraform-ls`

radeksimko:
One more question - what Terraform version do you use for this configuration?

radeksimko:
Also could you please try to run the following (in the mentioned workspace):

```
terraform providers schema -json
```
with and without your auth wrapper?

garry-harthill-cko:
> Also could you please try to run the following (in the mentioned workspace):
> 
> ```
> terraform providers schema -json
> ```
> 
> with and without your auth wrapper?
```
terraform -v
Terraform v0.12.25
```

Without the auth wrapper I get the error I described before:
```
Error: No valid credential sources found for AWS Provider.
        Please see https://terraform.io/docs/providers/aws/index.html for more information on
        providing credentials for the AWS Provider
```

With the auth wrapper it appears to work and returns a very large json document:
```
{"format_version":"0.1","provider_schemas":{"aws":{"provider":{"version":0,"block":{"attributes":{"access_key":{"type":"string","description":"The access key for API operations. You can retrieve this\nfrom the 'Security \u0026 Credentials' section of the AWS console.","optional":true},
...
...
}}}}}}}}
```

garry-harthill-cko:
Can this schema document be downloaded separately and placed in a directory manually for use by the extension?

radeksimko:
Thank you 🙏 

It appears that `terraform providers schema -json` does in fact require state authentication, for some reason, which may be related to https://github.com/hashicorp/terraform-ls/issues/56 🤔 

You may be able to override the path to Terraform in the meantime via the following VSCode configuration (`settings.json`):

```json
{
    "terraform.languageServer.args": [
        "serve",
        "-tf-exec=/auth/wrapper terraform"
    ]
}
```
We could also make this a more 1st-class settings in VSCode in the future.

> Can this schema document be downloaded separately and placed in a directory manually for use by the extension?

[...]

PR #205, opened 2020-07-03
textDocument/didOpen: Enrich warning message

radeksimko:
This is in reaction to https://discuss.hashicorp.com/t/no-root-module-found/10944

It's likely this will generate more pressure to resolving https://github.com/hashicorp/terraform-ls/issues/128 but it would at least get people closer to the point where it works (assuming they setup auth wrapper or use local state) and make the message more actionable.

### DIFF
--- a/langserver/handlers/did_open.go
+++ b/langserver/handlers/did_open.go
@@ -33,8 +33,11 @@ func TextDocumentDidOpen(ctx context.Context, params lsp.DidOpenTextDocumentPara
 
 	candidates := cf.RootModuleCandidatesByPath(f.Dir())
 	if len(candidates) == 0 {
-		msg := fmt.Sprintf("No root module found for %s"+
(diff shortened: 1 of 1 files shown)


eskp:
Can we disable this message? There might be no need for `terraform init`

radeksimko:
I think that adding an option to opt-in to ignoring these messages would be fine, assuming that only the users who know what they're doing actually opt in, but I'd like to be sure that everyone understand the implications before doing it.

The server cannot do much (basically nothing other than auto-formatting) without initialized root module. It can't access the schema, hence it can't use the schema in completion, later hovering, or any other future features.

Additionally since 0.4.0 the server does attempt to find the root module within the whole hierarchy and already reflects the fact that modules may not be `init`-ed on their own, but e.g. from a parent directory, which references that module. So this message in particular should rarely be a false negative and if it is then we'd like to know about it.

PR #210, opened 2020-07-04
Read go version from .go-version on GitHub Actions

jojo43:
As written in the comment, I fixed it to read version from `.go-version` instead of hard-coding.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,8 @@ jobs:
         name: Set up Go
         uses: actions/setup-go@v1
         with:
+          # TODO: Replace with go-version-from-file when it is supported
+          # https://github.com/actions/setup-go/pull/62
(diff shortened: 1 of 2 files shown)


radeksimko:
Hey @jojo43 
Thanks for the contribution. I'll read through more thoroughly next week, but just to be honest and transparent I was more hoping to see this solved in the context of the existing Go action as per https://github.com/actions/setup-go/issues/23#issuecomment-615233003

I think this is a decent temporary solution too, but have you considered contributing to that repository? I think that native support would significantly simplify the configuration and help avoiding the extra step.

jojo43:
> I was more hoping to see this solved in the context of the existing Go action

I totally agree you. 👍 

> have you considered contributing to that repository?

I've never, but going to look into it!

jojo43:
I opened a PR: https://github.com/actions/setup-go/pull/62

Issue #211, opened 2020-07-06
terraform-ls performs shutdown if any terraform 11 folder found in project


masseyc:
### Server Version

```
0.4.1
```

### Terraform Version

```
0.12.28 and 0.11.10
```

### Client Version

```
VSCode Terraform Plugin 2.0.2
```

### Terraform Configuration Files

```hcl
see attached
```

### Log Output

https://gist.github.com/masseyc/164b66092f351d12b008296dac3f2f62

### Expected Behavior

terraform-ls continues within terraform 12 folder/project

### Actual Behavior

terraform-ls detects folder with terraform 11 project and performs shutdown, does not provide completion etc to to other parts of project.

### Steps to Reproduce

Create VSCode project with 2 folders, each with a terraform project/folder, one based on 0.11, one 0.12 (as common in large projects using terraform version managers e.g. tfenv) terraform-ls will detect the 0.11 project and exit with error.

See attached simple example. With tfenv as terraform install, run terraform init in each folder then attempt to open in VSCode with terraform-ls enabled.
[terraform-ls-tfenv.tar.gz](https://github.com/hashicorp/terraform-ls/files/4881033/terraform-ls-tfenv.tar.gz)

Issue #164, opened 2020-06-15
Support provider aliases and 0.13 identities


paultyng:
### Terraform Version

This affects 0.13 mostly, but to a lesser degree 0.12

### Terraform Configuration Files

```hcl
terraform {
  required_providers {
    null = {
      source = "hashicorp/null"
      version = "2.1.2"
    }
    null2 = {
      source = "paultyng/null"
      version = "0.1.0"
    }
  }
}

provider "null2" {
    alias = "null3"
}

resource "null_resource" "foo" {

}

resource "null_resource" "bar" {
    provider = null2
}

resource "null_resource" "baz" {
    provider = null2.null3
}
```

### Expected Behavior

Auto-completion with multiple providers that have conflicting resource names should include the addition of the `provider = ` attribute.

### Actual Behavior
I believe resources with the same type names currently are just overwritten by the last one in. This can also probably be observed in 0.12 with `google` and `google-beta`.

paultyng:
https://github.com/hashicorp/terraform-config-inspect does not yet support listing out possible aliases, only ones in used on resources, so that will need to be addressed upstream (or parsed out manually).

radeksimko:
Just realized that provider block completion doesn't work at all with 0.13, since we're looking e.g. for `aws`, but 0.13 indexes that as `registry.terraform.io/hashicorp/aws`.

Also provider labels are completed with the FQN too, which is wrong:

![Screenshot 2020-06-30 at 19 42 54](https://user-images.githubusercontent.com/287584/86164801-429eec80-bb0a-11ea-8a53-15d07b147ee4.png)

Issue #218, opened 2020-07-08
Improve UX for async edge cases


radeksimko:
Once https://github.com/hashicorp/terraform-ls/issues/195 & https://github.com/hashicorp/terraform-ls/pull/219 is resolved it's likely that a number of edge cases will emerge.

Error messages may not be as helpful as we'd like them to be and the UX overall may need some enhancements, but it's not clear yet how to best address all these edge cases, so I'm creating this issue to track it and brain dump some known unknowns.

When we add a new root module, we always discover module manifest and plugin lock file locations, just so they can be tracked by the watcher.

There is however a number of operations which run asynchronously after adding the module roughly in the following order

1. parsing module manifest
2. discovering terraform executor
3. obtaining terraform version
4. finding compatible language parser (there's just one for now, but it's gated on 0.12+)
5. obtaining schema (to be gated on version too as per https://github.com/hashicorp/terraform-ls/issues/164)

Any of the above can either be in progress of loading - i.e. not done _yet_ when the user is e.g.

 - requesting completion
 - requesting formatting

or any of these can fail, which makes any functionality depending on it to error out, e.g.

 - unreadable module manifest
 - no terraform found in `$PATH`
 - no terraform version found (e.g. because of `tfenv` unparseable output, or binary being un-executable)
 - no compatible parser found (e.g. because of too old TF version)
 - no compatible schema storage found (e.g. because of TF 0.11 and older)
 - failed to obtain schema (e.g. because of missing backend auth, invalid config or other reasons)

The question is how should these edge cases be communicated, if at all to the user and when.

There was an attempt to surface some of this in https://github.com/hashicorp/terraform-ls/pull/187 but surfacing TF errors will generally require some better handling. Also we should aim to keep the messaging relevant in time or per event (see https://github.com/hashicorp/terraform-ls/issues/212)

radeksimko:
As discussed with @paultyng we can just silently log most of these errors for now, as opposed to showing any notifications.

The easiest way to do this would probably be just logging a summary of errors after finishing the load of each module here:

https://github.com/hashicorp/terraform-ls/blob/fe3b6a04b8a26f7e096fc62f610877ea34d931ea/internal/terraform/rootmodule/root_module.go#L198-L199

PR #219, opened 2020-07-08
refactor: asynchronous loading of root module parts

### LINKED ISSUE #195: Cancel root module operations on shutdown

radeksimko:
Closes #195 

Closes #219

I also believe this solves https://github.com/hashicorp/terraform-ls/issues/211 but I would like to gate the `fmt` availability on version, or at least better understand what happens when user has an ancient version of Terraform which doesn't have `fmt`, or if it doesn't support STDIN input.

This should also make it slightly easier to solve https://github.com/hashicorp/terraform-ls/issues/164


internal/terraform/rootmodule/root_module.go
```diff
@@ -53,81 +69,146 @@ func newRootModule(ctx context.Context, dir string) *rootModule {
...
+	rm.isLoading = false
 }
 
-func (rm *rootModule) initTfExecutor(dir string) (*exec.Executor, error) {
+func (rm *rootModule) load(ctx context.Context) error {
+	var errs *multierror.Error
+	defer rm.CancelLoading()
+
+	// reset internal loading state
+	rm.isLoading = true
```
appilon, line 163: data race between this thread and someone potentially calling `CancelLoading`

internal/terraform/rootmodule/root_module.go
```diff
@@ -53,81 +69,146 @@ func newRootModule(ctx context.Context, dir string) *rootModule {
...
+	errs = multierror.Append(errs, err)
+
+	err = rm.UpdateSchemaCache(ctx, rm.pluginLockFile)
+	errs = multierror.Append(errs, err)
+
+	return errs.ErrorOrNil()
+}
+
+func (rm *rootModule) IsLoadingDone() bool {
+	return !rm.isLoading
```
appilon, line 192: Again likely needs lock

internal/terraform/rootmodule/root_module.go
```diff
@@ -138,81 +219,66 @@ func (rm *rootModule) initTfExecutor(dir string) (*exec.Executor, error) {
...
-				return nil
-			}
+	return nil
+}
 
-			return fmt.Errorf("unable to calculate hash: %w", err)
-		}
+func (rm *rootModule) discoverTerraformVersion(ctx context.Context) error {
+	if rm.tfExec == nil {
+		return nil
```
appilon, line 229: might be more appropriate to return error since you have attempted to call `discoverTerraformVersion` before some critical setup func

internal/terraform/rootmodule/root_module.go
```diff
@@ -138,81 +219,66 @@ func (rm *rootModule) initTfExecutor(dir string) (*exec.Executor, error) {
...
-	for _, path := range paths {
-		lf, err = newFile(path)
-		if err == nil {
-			return lf, nil
-		}
-		if !os.IsNotExist(err) {
-			return nil, err
-		}
+func (rm *rootModule) findCompatibleStateStorage() error {
+	if rm.tfVersion == "" {
```
appilon, line 243: same as above

internal/terraform/rootmodule/root_module.go
```diff
@@ -138,81 +219,66 @@ func (rm *rootModule) initTfExecutor(dir string) (*exec.Executor, error) {
...
-}
+func (rm *rootModule) findCompatibleLangParser() error {
+	defer func() {
+		rm.parserLoaded = true
+	}()
 
-func (f *file) Path() string {
-	return f.path
-}
+	if rm.tfVersion == "" {
```
appilon, line 261: same as above

internal/terraform/rootmodule/root_module.go
```diff
@@ -138,81 +219,66 @@ func (rm *rootModule) initTfExecutor(dir string) (*exec.Executor, error) {
...
+	rm.parser = p
 
-		return fmt.Errorf("unable to calculate hash: %w", err)
-	}
+	return nil
+}
 
-	return rm.UpdateModuleManifest(lf)
+func (rm *rootModule) LoadError() error {
+	return rm.loadErr
```
appilon, line 281: Could need locker, not sure

internal/terraform/rootmodule/root_module.go
```diff
@@ -232,15 +302,32 @@ func (rm *rootModule) UpdateModuleManifest(lockFile File) error {
...
+
+	if rm.parser == nil {
+		return nil, fmt.Errorf("no parser available")
+	}
+
+	return rm.parser, nil
+}
+
+func (rm *rootModule) IsParserLoaded() bool {
+	return rm.parserLoaded
```
appilon, line 326: could need locker

internal/terraform/rootmodule/root_module.go
```diff
@@ -232,15 +302,32 @@ func (rm *rootModule) UpdateModuleManifest(lockFile File) error {
...
+}
+
+func (rm *rootModule) IsSchemaLoaded() bool {
+	return rm.schemaLoaded
```
appilon, line 330: could need locker

internal/terraform/rootmodule/root_module.go
```diff
@@ -267,22 +354,52 @@ func (rm *rootModule) ReferencesModulePath(path string) bool {
...
+
+	if rm.tfExec == nil {
+		return nil, fmt.Errorf("no terraform executor available")
+	}
+
+	return rm.tfExec, nil
+}
+
+func (rm *rootModule) IsTerraformLoaded() bool {
+	return rm.tfLoaded
```
appilon, line 370: could need locker

internal/terraform/rootmodule/root_module.go
```diff
@@ -267,22 +354,52 @@ func (rm *rootModule) ReferencesModulePath(path string) bool {
...
 }
 
-func (rm *rootModule) UpdatePluginCache(lockFile File) error {
+func (rm *rootModule) UpdateSchemaCache(ctx context.Context, lockFile File) error {
 	rm.pluginMu.Lock()
 	defer rm.pluginMu.Unlock()
 
+	if !rm.tfLoaded {
```
appilon, line 377: could be data race

internal/terraform/rootmodule/root_module_manager.go
```diff
@@ -87,49 +97,48 @@ func (rmm *rootModuleManager) AddRootModule(dir string) (RootModule, error) {
...
-			return true
-		}
+	if rmm.syncLoading {
+		rmm.logger.Printf("synchronously loading root module %s", dir)
+		return rm, rm.load(ctx)
 	}
-	return false
+
+	rmm.logger.Printf("asynchronously loading root module %s", dir)
+	go rm.StartLoading()
```
appilon, line 106: seems like `StartLoading()` also spawns a goroutine? this function can be called synchronously I think

internal/terraform/rootmodule/root_module_manager_mock_test.go
```diff
@@ -21,24 +21,25 @@ func TestNewRootModuleManagerMock_noMocks(t *testing.T) {
...
+			tmpDir: {
+				TerraformExecQueue: &exec.MockQueue{
+					Q: []*exec.MockItem{
+						{
+							Args:   []string{"version"},
+							Stdout: "Terraform v0.12.0\n",
+						},
+						{
+							Args:   []string{"providers", "schema", "-json"},
+							Stdout: "{\"format_version\":\"0.1\"}\n",
```
appilon, line 35: Consider backticks for bit gross escaped strings

internal/terraform/rootmodule/root_module_manager.go
radeksimko: The main point of the PR was to make it asynchronous as the title suggests 😉

appilon: of course! I meant that `StartLoading` already calles `go func() { rm.load() }()` so starting `StartLoading()` in a goroutine is redundant, `StartLoading` will return almost immediately its not blocked.

internal/terraform/rootmodule/root_module.go
radeksimko: Yeah, I suppose it could return an error, but all errors from `load()` are just logged silently anyway, given the asynchronous nature of the loading.

More importantly it's not clear to me yet how, when and whether surface these errors to the user, so I created https://github.com/hashicorp/terraform-ls/issues/218 to revisit that problem later.

internal/terraform/rootmodule/root_module_manager.go
radeksimko: Oh I see - good catch - I will fix that!

radeksimko:
@appilon I believe I addressed all of your valuable feedback. PTAL.

Issue #217, opened 2020-07-08
Very High CPU consumption


jamesskinner:
### Server Version

```
0.4.1
```

### Terraform Version

```
0.12.24
```

### Client Version

```
neovim (v0.4.3) coc plugin
```

### Log Output

https://gist.github.com/jamesskinner/b759199add68935349926b9e052d9bbd

### Expected Behaviour

minimal cpu usage

### Actual Behaviour

99.9% cpu usage

### Steps to Reproduce

I don't see a specific scenario. I do work on a codebase containing multiple root terraform repositories in subfolders. I am on MacOS v10.15.4

radeksimko:
Thank you for sharing the log @jamesskinner and filing the issue.

I can see from the log that you have a folder with at least 16 root modules in it which the server is attempting to load schema for.

There may be some ways we can tackle this problem, but before jumping to workarounds and solutions I'd like to first understand the full scale of your setup. Do you mind sharing how many root modules _in total_ would you typically open in your editor within the context of a single folder "hierarchy"? i.e. can you count the number of `.terraform` directories in the whole hierarchy?

jamesskinner:
@radeksimko yes; we use terraform cloud to provision an environment which is made up of (at the moment) 11 workspaces. Each one maps to a root terraform directory which we keep in a single git repository.

The root of the repo looks roughly like this:

```
.
├── account
│   ├── main.tf
│   └── variables.tf
├── components
│   ├── a
│   │   ├── main.tf
│   │   └── variables.tf
│   ├── b
│   │   ├── main.tf
│   │   └── variables.tf
│   ├── c
│   │   ├── main.tf
│   │   └── variables.tf
│   └── d,e,f...
└── tfe-workspaces
    ├── main.tf
    └── variables.tf
```
Then each of the subdirectories under `components/`, as well as the `account` and `tfe-workspaces` direcories are their own root terraform modules.

typically I would open my editor at the root of the **repository** and have buffers open from multiple "component" subdirectories. Some of the sub-directories, I will have run `terraform init` inside, some I will not have done

Hope that helps :-)

radeksimko:
Seeing the hierarchy is useful, and 11 root modules (reflected as TFC's workspaces) doesn't sound _too_ high. I can see that the schema was obtained for the first 5 of those within 1-2 seconds, which also suggests you don't have too complex provider setup (not too many providers per root module or no providers with large schemas).

[...]

PR #223, opened 2020-07-10
Introduce CPU & memory profiling

radeksimko:
This aims to help with debugging https://github.com/hashicorp/terraform-ls/issues/217

### DIFF
--- a/commands/serve_command.go
+++ b/commands/serve_command.go
@@ -7,6 +7,8 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"runtime"
+	"runtime/pprof"
 	"strings"
 	"syscall"
 	"time"
@@ -28,6 +30,8 @@ type ServeCommand struct {
 	tfExecPath    string
 	tfExecLogPath string
 	tfExecTimeout string
+	cpuProfile    string
+	memProfile    string
 }
 
 func (c *ServeCommand) flags() *flag.FlagSet {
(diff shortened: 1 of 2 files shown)


commands/serve_command.go
```diff
@@ -54,6 +60,19 @@ func (c *ServeCommand) Run(args []string) int {
...
+		stop, err := writeCpuProfileInto(c.cpuProfile)
+		if err != nil {
+			c.Ui.Error(err.Error())
+			return 1
+		}
+		defer stop()
+	}
+
+	if c.memProfile != "" {
+		defer writeMemoryProfileInto(c.memProfile)
```
appilon, line 73: I'm unsure how memory profiling works? It seems to me like on exit we will simply write out a memory snapshot at that moment (which is as the system is exiting) is this an accurate profile? I envision profiling to be something that is written out over time asynchronously? Perhaps a goroutine that writes out on a ticker?

I have never done this so if this is "how it's done in Go" then awesome TIL!

commands/serve_command.go
```diff
@@ -54,6 +60,19 @@ func (c *ServeCommand) Run(args []string) int {
 		return 1
 	}
 
+	if c.cpuProfile != "" {
+		stop, err := writeCpuProfileInto(c.cpuProfile)
+		if err != nil {
+			c.Ui.Error(err.Error())
+			return 1
+		}
+		defer stop()
```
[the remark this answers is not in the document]
radeksimko: Yeah, I think you're right - this way I'd be leaving a file descriptor open when starting the profiling fails. I'll move it up.

commands/serve_command.go
radeksimko: I think that is mostly accurate description - i.e. yes - we would profile whatever allocated memory is left after the explicit garbage collection. It's not going to be a silver bullet for all memory-related issues, but I think it's a good start and it follows what gopls does and also what the docs suggest:

 - https://github.com/golang/tools/blob/eb789aa7ce5077c5b70da4ebd0bd36e2facf2107/go/analysis/internal/checker/checker.go#L105-L118
 - https://golang.org/pkg/runtime/pprof/

Personally I think this will likely be sufficient for the foreseeable future, because we should be far more concerned about memory leaks and "cold memory" which is still left allocated on shutdown, as opposed to short-term peaks.

I certainly not opposed to adding something more fancy as we start to care more about the runtime memory usage, but as you described this would require some more complex instrumentation and I'd like to ship something that solves the first category of problems and iterate on it later.

PR #234, opened 2020-07-14
Fix some issues observed via staticcheck

paultyng:


internal/terraform/exec/exec_mock.go
```diff
@@ -56,7 +56,7 @@ func (mc *MockCall) UnmarshalJSON(b []byte) error {
 		return err
 	}
 
-	mc = (*MockCall)(q.Q[0])
+	*mc = *(*MockCall)(q.Q[0])
```
paultyng, line 59: I'm not entirely sure about this one, but just normal assignment here didn't surface this value out of `UnmarshalJSON` so was broken already.

internal/terraform/exec/exec.go
```diff
@@ -268,7 +266,7 @@ func (e *Executor) Version(ctx context.Context) (string, error) {
 	if len(lines) < 1 {
 		return "", fmt.Errorf("unexpected version output: %q", outString)
 	}
-	version := strings.TrimLeft(lines[0], "Terraform v")
+	version := strings.TrimPrefix(lines[0], "Terraform v")
```
radeksimko, line 269: 🙈

internal/terraform/exec/exec_mock.go
radeksimko: If this is broken then I wonder why it doesn't come up in tests 🤔

radeksimko: This is because the unmarshaler is never called since we always explicitly unmarshal into `MockItem`:
https://github.com/hashicorp/terraform-ls/blob/592480614510963caf095b7b7e27c6f368b6bfc8/internal/terraform/exec/exec_mock.go#L110-L118

so I think it's better to just remove the unmarshaler entirely.

PR #251, opened 2020-07-30
add support for specify exclude root module path

### LINKED ISSUE #250: Allow exclusion of paths in root module discovery
### Current Version

```
0.5.4
```

### Use-cases

As mentioned in https://github.com/hashicorp/vscode-terraform/issues/169 and https://github.com/hashicorp/terraform-ls/issues/186#issuecomment-665330015 the old version of VS Code extension had a way of excluding paths from indexing. 

While in principle similar, the purpose was different - the extension was merely indexing `*.tf` files AFAIK.

The root module auto-discovery mechanism in the language server doesn't (yet) care about `*.tf` files, but it would benefit from a similar exclusion option. It could basically be the opposite of `rootModulePaths` where the user can still use the discovery, but exclude known "problematic" paths from it. This can be specifically useful for users with more complex hierarchies and/or Terragrunt users.

### Attempted Solutions

Use [`rootModulePaths`](https://github.com/hashicorp/terraform-ls/blob/master/docs/SETTINGS.md#rootmodulepaths-string), which then bypasses the discovery entirely.

### Proposal

Introduce `excludeModulePaths []string` option which can initially support the same kind of input as `rootModulePaths` - i.e. absolute & relative, symlink eval'd, `~` replaced, but later also glob patterns.

njuCZ:
fix #250

not support glob pattern yet.

another question: if users change the settings in editor side, for now the vscode extension would remind user to reload. But I wonder implementing configurationChanged event in language server side  might be a better way?

### DIFF
--- a/internal/settings/settings.go
+++ b/internal/settings/settings.go
@@ -1,12 +1,14 @@
 package settings
 
 import (
+	"fmt"
 	"github.com/mitchellh/mapstructure"
 )
 
 type Options struct {
 	// RootModulePaths describes a list of absolute paths to root modules
-	RootModulePaths []string `mapstructure:"rootModulePaths"`
+	RootModulePaths    []string `mapstructure:"rootModulePaths"`
+	ExcludeModulePaths []string `mapstructure:"excludeModulePaths"`
 
 	// TODO: Need to check for conflict with CLI flags
 	// TerraformExecPath string
@@ -15,6 +17,9 @@ type Options struct {
 }
 
 func (o *Options) Validate() error {
+	if len(o.RootModulePaths) != 0 && len(o.ExcludeModulePaths) != 0 {
+		return fmt.Errorf("at most one of `rootModulePaths` and `excludeModulePaths` could be set")
+	}
 	return nil
 }
 
--- a/internal/terraform/rootmodule/path.go
+++ b/internal/terraform/rootmodule/path.go
@@ -10,3 +10,8 @@ func pathEquals(path1, path2 string) bool {
 	volume2 := filepath.VolumeName(path2)
 	return strings.EqualFold(volume1, volume2) && path1[len(volume1):] == path2[len(volume2):]
 }
+
+func ToLowerVolumePath(path string) string {
+	volume := filepath.VolumeName(path)
+	return strings.ToLower(volume) + path[len(volume):]
+}
--- a/internal/terraform/rootmodule/path.go
+++ b/internal/terraform/rootmodule/path.go
@@ -10,8 +10,3 @@ func pathEquals(path1, path2 string) bool {
 	volume2 := filepath.VolumeName(path2)
 	return strings.EqualFold(volume1, volume2) && path1[len(volume1):] == path2[len(volume2):]
 }
-
-func ToLowerVolumePath(path string) string {
-	volume := filepath.VolumeName(path)
-	return strings.ToLower(volume) + path[len(volume):]
-}
--- a/internal/terraform/rootmodule/walker.go
+++ b/internal/terraform/rootmodule/walker.go
@@ -32,7 +32,7 @@ type Walker struct {
 	cancelFunc context.CancelFunc
 	doneCh     <-chan struct{}
 
-	excludeModulePaths []string
+	excludeModulePaths map[string]bool
 }
 
 func NewWalker() *Walker {
@@ -48,7 +48,10 @@ func (w *Walker) SetLogger(logger *log.Logger) {
 }
 
 func (w *Walker) SetExcludeModulePaths(excludeModulePaths []string) {
-	w.excludeModulePaths = excludeModulePaths
+	w.excludeModulePaths = make(map[string]bool)
+	for _, path := range excludeModulePaths {
+		w.excludeModulePaths[path] = true
+	}
 }
 
 type WalkFunc func(ctx context.Context, rootModulePath string) error
@@ -112,11 +115,6 @@ func (w *Walker) IsWalking() bool {
 func (w *Walker) walk(ctx context.Context, rootPath string, wf WalkFunc) error {
 	defer w.Stop()
 
-	excludePathMap := make(map[string]bool)
-	for _, path := range w.excludeModulePaths {
-		excludePathMap[path] = true
-	}
-
 	err := filepath.Walk(rootPath, func(path string, info os.FileInfo, err error) error {
 		select {
 		case <-w.doneCh:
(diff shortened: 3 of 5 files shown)


langserver/handlers/initialize.go
```diff
@@ -108,6 +108,19 @@ func (lh *logHandler) Initialize(ctx context.Context, params lsp.InitializeParam
...
 	}
 
+	var excludeModulePaths []string
+	for _, rawPath := range cfgOpts.ExcludeModulePaths {
+		rmPath, err := resolvePath(rootDir, rawPath)
+		if err != nil {
+			jrpc2.ServerPush(ctx, "window/showMessage", &lsp.ShowMessageParams{
+				Type:    lsp.MTWarning,
+				Message: fmt.Sprintf("Ignoring exclude root module path %s: %s", rawPath, err),
+			})
```
radeksimko, line 118: I'm not sure about the value of this warning being surfaced to the user. I would probably just "downgrade" it to a silent log message if not ignore it entirely.

The worst case scenario is that we load some extra modules (e.g. due to typo) and while this may currently cause some CPU spikes I'd say that's something we have to address either way and warning on this level isn't a solution for that problem.

I assume that some users are likely to put a "preemptive list" of some relative paths which may or may not exist in each project and _not finding_ these paths there would be _expected_, so flooding their IDE with warnings is probably undesired in such case.

internal/settings/settings.go
```diff
@@ -6,7 +6,8 @@ import (
 
 type Options struct {
 	// RootModulePaths describes a list of absolute paths to root modules
-	RootModulePaths []string `mapstructure:"rootModulePaths"`
+	RootModulePaths    []string `mapstructure:"rootModulePaths"`
+	ExcludeModulePaths []string `mapstructure:"excludeModulePaths"`
```
radeksimko, line 10: I think we should add some validation to reflect that these two options are mutually exclusive. i.e. `RootModulePaths` conflicts with `ExcludeModulePaths`.

internal/terraform/rootmodule/walker.go
```diff
@@ -103,8 +103,14 @@ func (w *Walker) IsWalking() bool {
 	return w.walking
 }
 
-func (w *Walker) walk(ctx context.Context, rootPath string, wf WalkFunc) error {
+func (w *Walker) walk(ctx context.Context, rootPath string, excludePath []string, wf WalkFunc) error {
 	defer w.Stop()
+
+	excludePathMap := make(map[string]bool)
+	for _, path := range excludePath {
+		excludePathMap[ToLowerVolumePath(path)] = true
```
radeksimko, line 111: Thanks for adding the sanitazition but wouldn't it be better to do this in the initialize handler (`resolvePath`), where we also resolve symlinks, homedir etc.? I feel that the walker should just expect already sanitized input.

radeksimko:
> not support glob pattern yet.

That is fine, I would just keep that for another PR/issue later.

> another question: if users change the settings in editor side, for now the vscode extension would remind user to reload. But I wonder implementing configurationChanged event in language server side might be a better way?

Yes - this is the long-term plan.

Updating the configuration alone would be easy, but applying it is not as trivial in practice, because e.g. the walker currently expects just a single run per session, secondly we have to discuss the UX around this - e.g. how do we know the user has added new, yet undiscovered path - do we compare it against the list of known paths, when do we re-discover, what do we do with invalid config (conflicting options) etc. etc.

Again - it's something I would leave for another issue/PR and this one will also likely need some further discussion.

langserver/handlers/initialize.go
```diff
@@ -118,7 +131,7 @@ func (lh *logHandler) Initialize(ctx context.Context, params lsp.InitializeParam
 	// Walker runs asynchronously so we're intentionally *not*
 	// passing the request context here
 	bCtx := context.Background()
-	err = walker.StartWalking(bCtx, fh.Dir(), func(ctx context.Context, dir string) error {
+	err = walker.StartWalking(bCtx, fh.Dir(), excludeModulePaths, func(ctx context.Context, dir string) error {
```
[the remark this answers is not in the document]
njuCZ: making excludeModulePaths a property of `walker` instance will make it have state. since `startWalking` func will only be invoked once, there won't be problems. I will do it in this way, but we might take care of it because it might lead unexpected results if invoking multiple times.

langserver/handlers/initialize.go
njuCZ: yes, in fact I agree with you. I just found the same code when handling Static user-provided rootModulepaths, I wonder should we "downgrade" to a log message too ?

langserver/handlers/initialize.go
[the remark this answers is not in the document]
radeksimko: FWIW the walker isn't quite ready to be invoked more than once now either way (due to some other internal state), but I think that making it a property will make it easier later for walking and configuration resetting to be managed independently. At that point we can guard the slice with a mutex, or maybe a semaphore, so that e.g. we can detect when user attempts to change the configuration while the walker is walking and decide (or let them decide) what to do.

langserver/handlers/initialize.go
radeksimko: Good question - I'd still keep it for _absolute_ paths there, as that's much more explicit and the user probably doesn't intend for it to be ignored, but we can avoid the warning for relative paths.

You might ask why not apply the same principle here - and I wouldn't object strongly, but I still think that _not_ excluding something you might have wanted to exclude has far less impact than _not including_ - so that's how I'd draw the line.

njuCZ:
@radeksimko thanks for your suggestions. I have modified it, please have a look again if available

internal/terraform/rootmodule/path.go
```diff
@@ -10,3 +10,8 @@ func pathEquals(path1, path2 string) bool {
 	volume2 := filepath.VolumeName(path2)
 	return strings.EqualFold(volume1, volume2) && path1[len(volume1):] == path2[len(volume2):]
 }
+
+func ToLowerVolumePath(path string) string {
```
radeksimko, line 14: I think this function doesn't need to be under `rootmodule`, but that's rather a nitpick - I can move it in a separate commit. 😄

internal/terraform/rootmodule/walker.go
```diff
@@ -105,6 +111,12 @@ func (w *Walker) IsWalking() bool {
 
 func (w *Walker) walk(ctx context.Context, rootPath string, wf WalkFunc) error {
 	defer w.Stop()
+
+	excludePathMap := make(map[string]bool)
+	for _, path := range w.excludeModulePaths {
+		excludePathMap[path] = true
+	}
```
radeksimko, line 118: Totally nitpick - we could be forming the data into the map even earlier - in the setter, instead of storing the data in "inappropriate form" and then converting them when we need them. I can move it in a separate commit though.

internal/terraform/rootmodule/path.go
njuCZ: I don't know where to place it properly, please kindly take it over

radeksimko: I think the best place for the function is the package where it is called from - I will move it there, no worries. 👍

PR #252, opened 2020-08-04
auto-completion providers from registry

### LINKED ISSUE #194: Search registry providers via completion
### Use-cases

* Adding a provider to a workspace, need to search available providers that aren't present locally

### Proposal

Use a registry / algolia search endpoint to source possible providers based on a query.

njuCZ:
fix #194 

based on current logic, there seems no need to use algolia search.
we could just get all providers from registry and reuse existing logic


njuCZ:
@radeksimko thanks for your opinion, I am going to take a look at these two issues first

Issue #186, opened 2020-06-25
Set a cap for auto-loading root modules


radeksimko:
### Current Version

```
0.3.2
```

### Use-cases

Currently the language server will attempt to load all root modules (basically directories with `.terraform` in them) on `initialize`. This is probably fine for most users which have sensible amount of root modules under the directory opened in the IDE, but it's unclear what the UX may be for users with "too many" root modules.

This may lead to high CPU and/or memory usage for potentially long time, before the user even opens a single file which is certainly not a great UX.

### Proposal

We can find and set a high enough cap which won't affect most users, but still provide a decent UX for those who have many root modules.

The cap could be dynamic based on the amount of memory and/or CPU available, but we can't predict how much schema we obtain from plugins for each root module and so there should be a hard cap anyway.

We can collect some data and measure the performance with increasing amount of modules, but (_completely uneducated guess_) I would hope that we can reasonably support some hundreds of root modules at least.

Also/instead we can delay loading root modules until `didOpen`. This has a downside - obtaining schema may take some time and so completion and other features just won't work until the schema is obtained. This could be in reality low seconds per provider. Scheduling the schema loading could be configurable though.

yruss972:
If this is what's launching hundreds of calls to terraform version, then the UX is awful. My machine is useless until this finishes

radeksimko:
Just out of curiosity do you use `tfenv` @yruss972 ? There is a [known issue](https://github.com/tfutils/tfenv/issues/196) causing ~1sec delay for any `terraform` commands which may be impacting the UX for cases like these. Each `terraform version` should otherwise finish within tens of miliseconds.

That aside I think we could make the discovery a bit smarter by avoid calling `terraform version` in cases where it's unlikely that the version will differ - e.g. there is no `.terraform-version` or `.envrc` file in the directory.

phekmat:
I'm seeing the same as @yruss972 with far fewer modules, but I'm using terragrunt

radeksimko:
@phekmat Do you mind translating "fewer" into more tangible quantity for me, so we have a more accurate data point to work with? Also I'm not that familiar with Terragrunt, but I'm assuming it doesn't install/manage Terraform itself, so I'm assuming there's still a chance that you may use tfenv with terragrunt -> do you?

phekmat:
> 
> 
[...]

PR #256, opened 2020-08-10
Cap root module loading via worker pool

radeksimko:
This represents a foundation for resolving #186 

A capped worker pool is used when loading root modules, which helps spread the CPU usage and effectively flatten the spikes and therefore improve the UX.

Charts generated via https://github.com/mkevac/debugcharts

## CPU Before

plot-cpu-without-cap

## CPU After

plot-cpu-with-cap

## pprof Before

plot-pprof-without-cap

## pprof After

plot-pprof-with-cap

## Deferred ideas

We can also make parallelism configurable, if deemed necessary, but I think that can be discussed and addressed separately.


internal/terraform/rootmodule/root_module_manager.go
```diff
@@ -40,9 +43,14 @@ func NewRootModuleManager(fs tfconfig.FS) RootModuleManager {
 
 func newRootModuleManager(fs tfconfig.FS) *rootModuleManager {
 	d := &discovery.Discovery{}
+
+	defaultSize := 3 * runtime.NumCPU()
```
appilon, line 47: 3 is the magic number I'm guessing?

radeksimko: 3 is basically an educated guess. I did some simple testing as shown above with 1000 root modules in a hierarchy (which seemed to be the higher end of range reported by folks in #186 ). That effectively triggers `terraform version` 1000 times and I ran that on my i7 (Macbook) with `4` cores (= pool size `12`) and it finished within a few seconds.

I don't expect this to be the best configuration nor the best way of testing this, but I do expect it to behave better than before just because we trade time+memory for less CPU in principle.

I do expect we will tweak the number in the near future based on the feedback from end users and further (more scoped) testing. This is also why I added the logging.

radeksimko: Generally the performance will largely depend on the time it takes to finish the loading, which _after_ `terraform version` depends on the size of the schema and exec time of `terraform providers schema -json` alone may span from half a second to 10+ seconds in some worst case scenarios, so there is no easy answer to this I'm afraid.

We can however make it configurable and let users deal with edge cases (too slow CPU and/or too many modules) that way. Also it would be even better to just avoid calling `terraform version` where we don't need to as discussed in https://github.com/hashicorp/terraform-ls/issues/186#issuecomment-661761779

PR #257, opened 2020-08-11
ask for init if current folder is empty root module

### LINKED ISSUE #106: Ask to `terraform init` an uninitialized workspace

njuCZ:
fix #106.

there are some points that have not been considered. such as if current dir is in `IncludeRootModule` or `ExcludeRootModule`

### DIFF
--- a/internal/terraform/exec/exec.go
+++ b/internal/terraform/exec/exec.go
@@ -87,6 +87,17 @@ func (e *Executor) setLogPath(method string) error {
 	return e.tf.SetLogPath(logPath)
 }
 
+func (e *Executor) Init(ctx context.Context) error {
+	ctx, cancel := e.withTimeout(ctx)
+	defer cancel()
+	err := e.setLogPath("Init")
+	if err != nil {
+		return err
+	}
+
+	return e.contextfulError(ctx, "Init", e.tf.Init(ctx))
+}
+
 func (e *Executor) Format(ctx context.Context, input []byte) ([]byte, error) {
 	ctx, cancel := e.withTimeout(ctx)
 	defer cancel()
--- a/internal/terraform/exec/mock/executor.go
+++ b/internal/terraform/exec/mock/executor.go
@@ -58,6 +58,20 @@ func (_m *Executor) GetExecPath() string {
 	return r0
 }
 
+// Init provides a mock function with given fields: ctx
+func (_m *Executor) Init(ctx context.Context) error {
+	ret := _m.Called(ctx)
+
+	var r0 error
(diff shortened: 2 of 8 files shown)


radeksimko (commented):
Sorry it took me so long to review this PR.

I appreciate the PR as a form of starting a discussion.

A few high-level comments/questions which came up as I was reading the code:

1. I'm not fully convinced about the value of tracking "potential root modules" specifically in this context. I'd say that the root module is no different from the (existing initialized) tracked ones after the user confirms the initialization. Relatedly that recently initialized module also needs to be watched for changes.
2. I understand we need to track these "candidates" somewhere until they're actually initialized and added for tracking, but I'm not sure whether expanding the responsibility of `RootModuleManager` to do this is the right approach. 🤔  It would also mean we'd need to integrate the watcher there somehow. So maybe it's better to keep this logic apart in some form of `RootModuleInitializer`?
3. We're in the process of integrating `hashicorp/terraform-exec` which may change how we call Terraform a little bit internally.
4. As part of https://github.com/hashicorp/terraform-ls/issues/27 we'll need to enable clients also to run certain commands, and that would likely include `terraform init`. I'm not saying that it should be a blocker for fixing #106 or that we need to call init the exact same way as clients would, but it's something to keep in mind.

njuCZ:
@radeksimko @appilon I have updated this PR, could you have a look when free?
1:  have added watch for recently initialized module
2:  introduced ths `RootModuleInitializer`. After considering the relation of `RootModuleManager` and `RootModuleInitializer`, I think maybe it's better to make `RootModuleInitializer` as a field of `RootModuleManager`, and I have moved some Init Root module logic to the new struct
3 & 4: have used `terraform-exec` to execute `terraform init`

radeksimko (commented):
I'm sorry it took me a while to get to this PR, but a colleague of mine is working on a similar feature on VS Code side and we needed to sync up and make sure that we don't introduce something conflicting.

The PR looks good in general - Frankly I think the UX around modules will go through many changes in the near future as we get it more accurate and relevant, but since your PR just makes the existing message more actionable and still leaves user the choice I think there is no harm in merging it, once we address the in-line comments.

njuCZ:
@radeksimko thanks for your suggestions. I have updated this PR.

PR #268, opened 2020-09-17
Use terraform-exec

### LINKED ISSUE #266: Use terraform-exec
Upcoming features such as workspace switching would benefit from the internal terraform execution wrapper being swapped for the official wrapper: https://github.com/hashicorp/terraform-exec

appilon:
Will bump terraform-exec to proper release before merging, but ready for review. Previously we maintained an "executor" library internally, the implementation had a nice trick that allowed it to become mocked for unit testing. While switching to terraform-exec was relatively easy, decoupling the built in mock required some refactoring.

This represents the least obtrusive drop in replacement of terraform-exec I could think of. With further refactoring we could likely remove the wrapping `Executor` struct. This refactor tried to strictly follow placing the responsibility of interface definitions on the consuming packages, which allows the producing packages to work in concrete types, and thus not forcing 1 large monolithic interface just for the sake of a mock to build up, albeit the interface created in `rootmodule` covers most of the surface area.

Closes #266


internal/terraform/exec/exec.go
```diff
@@ -59,12 +51,36 @@ func NewExecutor(path string) *Executor {
...
+// directory after the creation of the Executor instance.
+func (e *Executor) tfExec() *tfexec.Terraform {
+	if e.tf == nil {
+		tf, err := tfexec.NewTerraform(e.workDir, e.execPath)
+		if err != nil {
+			panic(err)
+		}
+		tf.SetLogger(e.logger)
+
+		// TODO: make sense of how this works
```
appilon, line 68: Still need to reason about how this worked

internal/terraform/exec/exec.go
```diff
@@ -59,12 +51,36 @@ func NewExecutor(path string) *Executor {
...
+		// if e.execLogPath != "" {
+		// 	logPath, err := logging.ParseExecLogPath(cmd.Args, e.execLogPath)
+		// 	tf.SetLogPath(logPath)
+		// }
+
+		// TODO: figure out what env vars are already handled by tfexec
```
appilon, line 74: Still need to get env vars to whatever they were before, goal of the PR was to keep the "experience" the same as before (given my limited knowledge of the LS codebase at this time)

internal/terraform/exec/exec.go
```diff
@@ -210,65 +124,30 @@ func (e *Executor) FormatterForVersion(v string) (Formatter, error) {
...
-	outString := string(out)
-	lines := strings.Split(outString, "\n")
-	if len(lines) < 1 {
-		return "", fmt.Errorf("unexpected version output: %q", outString)
+		return "", err
 	}
-	version := strings.TrimPrefix(lines[0], "Terraform v")
-
-	return version, nil
+	// TODO: consider refactoring codebase to work directly with go-version.Version
```
appilon, line 149: I noticed a fair bit of back and forth between `go-version.Version` and the stringified representation. I think given I saw `go-version` types bled onto a few structs throughout the codebade, and given `terraform-exec` works with it directly it may be worth changing the language server codebase to work directly in that type, in a follow up PR.

internal/terraform/exec/exec_mock.go
```diff
@@ -66,33 +72,126 @@ func (mc *MockQueue) NextMockItem() *MockItem {
...
+	}
+
+	return &schemas, nil
+}
+
+func (m *mockExecutor) FormatterForVersion(v string) (Formatter, error) {
+	return formatterForVersion(v, m.Format)
+}
+
+func (m *mockExecutor) run(ctx context.Context, args ...string) ([]byte, error) {
```
appilon, line 133: the `mockExecutor` is essentially a stripped down copy of the previous "melded" implementation we had before

internal/terraform/exec/exec_mock_test.go
```diff
@@ -13,3 +17,82 @@ func TestMain(m *testing.M) {
 
 	os.Exit(m.Run())
 }
+
+func TestExec_timeout(t *testing.T) {
```
appilon, line 21: These tests were moved out of `exec_test.go` because I felt it was just testing the mock.

internal/terraform/exec/exec.go
```diff
@@ -210,65 +112,30 @@ func (e *Executor) FormatterForVersion(v string) (Formatter, error) {
 	}
 
 	if ver.GreaterThanOrEqual(fmtCapableVersion) {
```
paultyng, line 114: You could move this version compatibility check to `terraform-exec` if you wanted.

internal/terraform/exec/exec.go
```diff
@@ -290,16 +157,16 @@ func (e *Executor) VersionIsSupported(ctx context.Context, c version.Constraints
...
-	if err != nil {
-		return nil, fmt.Errorf("failed to unmarshal response: %w", err)
-	}
+	return e.tfExec().ProvidersSchema(ctx)
+}
 
-	return &schemas, nil
+func contextWithTimeout(ctx context.Context, timeout time.Duration) (context.Context, context.CancelFunc) {
+	cancel := func() {}
+	if timeout > 0 {
```
paultyng, line 168: should it just be a `WithCancel` if its zero?

internal/terraform/exec/exec.go
appilon: this is something @radeksimko upstreaming now. Depending on the outcome of the upstreaming work, this PR may be superseded.

internal/terraform/exec/exec_mock_test.go
paultyng: yeah i think you can just drop them

internal/terraform/exec/exec.go
radeksimko: https://github.com/hashicorp/terraform-exec/pull/88

radeksimko:
Just to summarize what we discussed via Zoom today:

1. templating support in logging paths - We could upstream that by introducing a method similar to `SetLogPath` (e.g. `SetLogPathTemplate`). I will look into that.
2. Formatting method with `[]byte`-compatible API - This is pending some questions about what the right name and API is. `io.Writer`+`io.Reader` for stdin/stdout may be better, but I'm unsure how well that fits next to `FormatWrite` and also whether we should sunset `FormatString`, or if we just introduce `FormatBytes`. Pinging @kmoe for opinions here.
3. Mocking library - I will look into upstreaming it.
4. Version handling in formatter - https://github.com/hashicorp/terraform-exec/pull/88

PR #269, opened 2020-09-30
Implement hcl parse diagnostics

appilon:
Diagnostic service asynchronously parses opened and changed documents, and pushes diagnostics to the client. Async model prevents `didOpen`, `didChange`, or future handlers from blocking/slowing down during more expensive validation.

The protocol specifies that all files of the project should be diagnosed at once:
>if a language has a project system (for example C#) diagnostics are not cleared when a file closes. When a project is opened all diagnostics for all files are recomputed (or read from a cache).

This PR should be followed up on to try and deliver on that, wanted to keep it simple win for now


langserver/diagnostics/diagnostics.go
```diff
@@ -0,0 +1,91 @@
...
+
+type Notifier struct {
+	sessCtx          context.Context
+	hclDocs          chan<- documentContext
+	closeHclDocsOnce sync.Once
+}
+
+func NewNotifier(sessCtx context.Context) *Notifier {
+	hclDocs := make(chan documentContext, 10)
+	go hclDiags(hclDocs)
```
radeksimko, line 28: Is there a reason we need the asynchronicity?

I mean each request is already being processed in a dedicated goroutine and parallelism is controlled in the jRPC library and that's limited to 1 for ordering reasons, so we'd never actually process more than 1 at a time anyway and even if for some reason we do in the future, then I reckon the performance/parallelism would be handled by the jRPC library on handler level?

radeksimko (commented):
LGTM functionally, but I left some questions about the implementation.

langserver/diagnostics/diagnostics.go
paultyng: Are push requests limited by the parallelism? Seems like this is just pushing each chan value.

langserver/diagnostics/diagnostics.go
```diff
@@ -0,0 +1,91 @@
...
+	return sev
+}
+
+func lspMessage(diag *hcl.Diagnostic) string {
+	m := diag.Summary
+	if diag.Detail != "" {
+		m += ": " + diag.Detail
+	}
+	return m
+}
```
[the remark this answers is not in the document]
appilon: Will move

langserver/diagnostics/diagnostics.go
appilon: The idea is we have a diagnostics worker thread that takes in document requests and pushes diags to the client as soon as possible, but it does not ever block or add milliseconds to the `didOpen` and `didChange` handlers (and upcoming `didSave` for validation).

radeksimko (approved):
Totally nitpick: I would probably pick a different name for `lsp/convert.go`, just because technically every logic in that package is "converting" in some form or shape. Perhaps it can be `diagnostics.go` for the severity stuff and `range.go` for the other two functions.

> This PR should be followed up on to try and deliver on that, wanted to keep it simple win for now

Agreed - can you please create an issue for this, describing roughly what needs to be done, just so we don't forget?

paultyng:
We were testing this yesterday and it looked like the squiggles may have an off by one display issue but unsure. Just wanted to note it here for double checking.

appilon:
> We were testing this yesterday and it looked like the squiggles may have an off by one display issue but unsure. Just wanted to note it here for double checking.

I will see if I can quickly resolve this or if it's coming from the hcl parser

PR #271, opened 2020-10-06
Replace most of terraform/exec w/ terraform-exec

### LINKED ISSUE #266: Use terraform-exec

radeksimko:
This should address all of these points https://github.com/hashicorp/terraform-ls/pull/268#issuecomment-698531081 and replace as much of the old package as possible.

Closes #266 
Closes #268


appilon (changes requested):
LGTM but need to tidy and vendor

appilon (approved):
NVM just noticed we stopped vendoring

PR #272, opened 2020-10-07
Remove vendoring

radeksimko:
This PR removes vendoring, as inspired [by Terraform](https://github.com/hashicorp/terraform/pull/26358) and motivated by the fact that `terraform-exec` currently depends on `go-getter` which in turn brings plenty of transitive dependencies, so this should make https://github.com/hashicorp/terraform-ls/pull/271 much smaller.


.github/workflows/ci.yml
```diff
@@ -8,6 +8,9 @@ on:
     branches:
       - master
 
+env:
+  GOPROXY: https://proxy.golang.org/
```
paultyng, line 12: is this not set by default?

paultyng (approved):
LGTM, one comment, also I wanted to inquire internally again to see if we have a proxy or not.

.github/workflows/ci.yml
radeksimko: The default (according to `go env`) is `https://proxy.golang.org,direct` and so the idea here is that by excluding `direct` we can effectively protect ourselves against the "leftpad situation" where the source would suddenly disappear by making sure that all dependencies are always cached before being added.

radeksimko:
> I wanted to inquire internally again to see if we have a proxy or not.

Sure, I'm happy to add it to the config once we confirm we have one, but I assume it's not a blocker.

PR #279, opened 2020-10-23
Unique command names

### LINKED ISSUE #278: Multiple languageserver instances will crash on vscode
The recently merged #274 will crash on vscode (tested), it seems to hit the exact case here

https://github.com/microsoft/vscode-languageserver-node/issues/333

screengrab

Screen Shot 2020-10-21 at 3 51 18 PM

One reply suggests doing something to ensure command uniqueness between server instances, but that is really disappointing and feels like a hack.

appilon:
This PR ensures the uniqueness of the exported commands across different language servers (with a `terraform-ls.` prefix, this is a common practice) and an additional `<commandPrefix>.`, this latter ensures multiple instances of the language server can run side by side, it will become the clients responsibility to send requests with the correct command name, the `commandPrefix` is only enabled if specified in the `initializationOptions` for the server.
```
"executeCommandProvider": {
    "commands": ["1234.terraform-ls.rootmodules"]
}
```

Closes #278

### DIFF
--- a/internal/context/context.go
+++ b/internal/context/context.go
@@ -34,6 +34,7 @@ var (
 	ctxRootModuleWalker  = &contextKey{"root module walker"}
 	ctxRootModuleLoader  = &contextKey{"root module loader"}
 	ctxRootDir           = &contextKey{"root directory"}
+	ctxCommandPrefix     = &contextKey{"command prefix"}
 	ctxDiags             = &contextKey{"diagnostics"}
 )
 
@@ -188,6 +189,28 @@ func RootDirectory(ctx context.Context) (string, bool) {
 		return "", false
 	}
 	return *rootDir, true
+}
+
+func WithCommandPrefix(ctx context.Context, prefix *string) context.Context {
+	return context.WithValue(ctx, ctxCommandPrefix, prefix)
+}
+
+func SetCommandPrefix(ctx context.Context, prefix string) error {
+	commandPrefix, ok := ctx.Value(ctxCommandPrefix).(*string)
+	if !ok {
+		return missingContextErr(ctxCommandPrefix)
+	}
+
+	*commandPrefix = prefix
+	return nil
+}
+
+func CommandPrefix(ctx context.Context) (string, bool) {
+	commandPrefix, ok := ctx.Value(ctxCommandPrefix).(*string)
+	if !ok {
+		return "", false
+	}
+	return *commandPrefix, true
 }
 
 func WithRootModuleWalker(ctx context.Context, w *rootmodule.Walker) context.Context {
--- a/internal/context/context.go
+++ b/internal/context/context.go
@@ -35,6 +35,7 @@ var (
 	ctxRootModuleLoader  = &contextKey{"root module loader"}
 	ctxRootDir           = &contextKey{"root directory"}
 	ctxDiags             = &contextKey{"diagnostics"}
+	ctxServerID          = &contextKey{"servier id"}
 )
 
 func missingContextErr(ctxKey *contextKey) *MissingContextErr {
@@ -188,6 +189,28 @@ func RootDirectory(ctx context.Context) (string, bool) {
 		return "", false
 	}
 	return *rootDir, true
+}
+
+func WithServerID(ctx context.Context, id *string) context.Context {
+	return context.WithValue(ctx, ctxServerID, id)
+}
+
+func SetServerID(ctx context.Context, id string) error {
+	serverID, ok := ctx.Value(ctxServerID).(*string)
+	if !ok {
+		return missingContextErr(ctxServerID)
+	}
+
+	*serverID = id
+	return nil
+}
+
+func ServerID(ctx context.Context) (string, bool) {
+	serverID, ok := ctx.Value(ctxServerID).(*string)
+	if !ok {
+		return "", false
+	}
+	return *serverID, true
 }
 
 func WithRootModuleWalker(ctx context.Context, w *rootmodule.Walker) context.Context {
--- a/internal/context/context.go
+++ b/internal/context/context.go
@@ -35,7 +35,7 @@ var (
 	ctxRootModuleLoader  = &contextKey{"root module loader"}
 	ctxRootDir           = &contextKey{"root directory"}
 	ctxDiags             = &contextKey{"diagnostics"}
-	ctxServerID          = &contextKey{"servier id"}
+	ctxServerID          = &contextKey{"server id"}
 )
 
 func missingContextErr(ctxKey *contextKey) *MissingContextErr {
--- a/internal/context/context.go
+++ b/internal/context/context.go
@@ -35,7 +35,7 @@ var (
 	ctxRootModuleLoader  = &contextKey{"root module loader"}
 	ctxRootDir           = &contextKey{"root directory"}
 	ctxDiags             = &contextKey{"diagnostics"}
-	ctxServerID          = &contextKey{"server id"}
+	ctxCommandPrefix     = &contextKey{"command prefix"}
 )
 
 func missingContextErr(ctxKey *contextKey) *MissingContextErr {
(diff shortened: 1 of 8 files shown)


appilon:
Paul had suggested the unique ID (maybe even just an incrementing integer) be generated by the client and passed to the server to prefix/suffix, which I'm open to as an alternative, or perhaps we use the `pid` as the unique ID, the client would have to figure out that mapping though.

aeschright:
For that second option to work, would the server then just keep track of what uuid was used to call it the first time, or is there some other way we'd register the setting?

appilon:
Correct the value would be passed to the server during initialization, it would have the same effect, just maybe feel a little cleaner without the long root folder URI, the client would likely keep the ID as the key to it's map (which as of writing is the workspace folder URI).

internal/context/context.go
```diff
@@ -35,6 +35,7 @@ var (
 	ctxRootModuleLoader  = &contextKey{"root module loader"}
 	ctxRootDir           = &contextKey{"root directory"}
 	ctxDiags             = &contextKey{"diagnostics"}
+	ctxServerID          = &contextKey{"servier id"}
```
aeschright, line 38:
```suggestion
	ctxServerID          = &contextKey{"server id"}
```

appilon: great 👁️

langserver/handlers/handlers_test.go
```diff
@@ -16,29 +17,38 @@ import (
...
-			"textDocumentSync": {
-				"openClose": true,
-				"change": 2
-			},
-			"completionProvider": {},
-			"documentSymbolProvider":true,
-			"documentFormattingProvider":true,
-			"executeCommandProvider": {
-				"commands": ["rootmodules"]
+func initializeResponse(t *testing.T, commandSuffix string) string {
```
radeksimko, line 20: Assuming that the ID is an optional parameter, should we maybe have a dedicated test which is the only one that defines it, instead of changing the default test data which used throughout other tests?

Relatedly I think we should have tests which do _not_ set that parameter, because that will at this moment be any client other than VSCode and we intend to continue supporting these clients.

langserver/handlers/initialize.go
```diff
@@ -77,6 +74,16 @@ func (lh *logHandler) Initialize(ctx context.Context, params lsp.InitializeParam
...
+
+	// set server ID
+	err = lsctx.SetServerID(ctx, out.Options.ID)
+	if err != nil {
+		return serverCaps, err
+	}
+	// apply suffix to executeCommand handler names
+	serverCaps.Capabilities.ExecuteCommandProvider = &lsp.ExecuteCommandOptions{
+		Commands: handlers.Names(out.Options.ID),
+	}
```
radeksimko, line 86: Nitpick: Is there a reason for setting it down here instead of moving the setter to the top and keeping it all capabilities in one place?

langserver/handlers/execute_command.go
```diff
@@ -7,26 +7,35 @@ import (
...
-	return names
+	return fullnames
 }
 
 func (lh *logHandler) WorkspaceExecuteCommand(ctx context.Context, params lsp.ExecuteCommandParams) (interface{}, error) {
-	handler, ok := handlers[params.Command]
+	serverID, _ := lsctx.ServerID(ctx)
+	name := strings.TrimPrefix(params.Command, commandPrefix)
+	name = strings.TrimSuffix(name, "."+serverID)
+	handler, ok := handlers[name]
```
radeksimko, line 38: Could we perhaps simplify this logic here by decoupling the lookup into a dedicated `executeCommandHandlers` method - say `FindCommand(name, suffix string) (executeCommandHandler, bool)` - then we can also more easily utilize `Names()` there and perform exact match on the key, instead of string key trimming?

internal/settings/settings.go
```diff
@@ -2,13 +2,15 @@ package settings
...
 	"fmt"
+
 	"github.com/mitchellh/mapstructure"
 )
 
 type Options struct {
 	// RootModulePaths describes a list of absolute paths to root modules
 	RootModulePaths    []string `mapstructure:"rootModulePaths"`
 	ExcludeModulePaths []string `mapstructure:"excludeModulePaths"`
+	ID                 string   `mapstructure:"id"`
```
radeksimko, line 13: Because this is exposed to the client and essentially becoming part of a public API, could we give it a slightly more descriptive name - e.g. `CommandPrefix` (`mapstructure:commandPrefix`)? If we ever need to use a unique ID for something else, then we can just add that as a separate option, but I would hope this is the only one we need to support for the foreseeable future. 😄

internal/context/context.go
```diff
@@ -35,6 +35,7 @@ var (
 	ctxRootModuleLoader  = &contextKey{"root module loader"}
 	ctxRootDir           = &contextKey{"root directory"}
 	ctxDiags             = &contextKey{"diagnostics"}
+	ctxServerID          = &contextKey{"server id"}
```
radeksimko, line 38: Related to my comment below, given what we use this option for, would "command prefix" be a perhaps a better name for this?

radeksimko (commented):
Thank you for working on this and for discovering this issue early before the release! 😅 

Besides my comments inline I would have one more question about the format - was there a particular reason behind `<static-prefix>.<command>.<instance-specific-prefix>`?
I don't have a strong opinion on the order, just genuinely curious if this was motivated by anything on the client side or in the LSP?

langserver/handlers/handlers_test.go
appilon: yes I agree I will change the default

langserver/handlers/initialize.go
appilon: I need the `initializationOptions` to have been parsed, and this is where that has happened

langserver/handlers/execute_command.go
appilon: I can isolate this code a bit more certainly. I found it easier to actually keep the prefix and suffix out of the map, so I likely will still need the trimming but I'll see what I can do

internal/settings/settings.go
appilon: I sort of feel strongly to the opposite approach. I wanted to be very careful about the things we bake into essentially our own protocol we have to live with. I thought the concept of assigning the server an `ID` would be very flexible and canonical, I wanted to keep the details of command suffixing out of it if possible.

But I do agree 🤔 having something called `ID` then be implicitly suffixed is sorta of a weird piece of knowledge client developers need to know about is also weird... I could maybe be convinced of the otherside, I need to sleep on it/discuss further with you/team.

appilon:
> Thank you for working on this and for discovering this issue early before the release! 😅
> 
> Besides my comments inline I would have one more question about the format - was there a particular reason behind `<static-prefix>.<command>.<instance-specific-prefix>`?
> I don't have a strong opinion on the order, just genuinely curious if this was motivated by anything on the client side or in the LSP?

@radeksimko the static prefix is a common pattern. I should maybe see if I can find some precedent with another popular langserver (I'll try python) and go with what they do, but I felt that since the instance specific ID is optional, I felt making it a suffix felt more intuitive. 
`terraform-ls.rootmodules` contrasted with `terraform-ls.1234.rootmodules`
VS
`terraform-ls.rootmodules` contrasted with `terraform-ls.rootmodules.1234` I liked that the first two pieces `<static-prefix>.<command>.` "lined up" and the last part can be dropped.

Similar to your other point though... I also now wonder if your suggestion is more intuitive in the world of namespacing like "please route request to `terraform-ls` `instance 1234`. Need to chat about it more..

langserver/handlers/execute_command.go
```diff
@@ -13,12 +13,26 @@ import (
 type executeCommandHandler func(context.Context, commandArgs) (interface{}, error)
 type executeCommandHandlers map[string]executeCommandHandler
 
-var handlers = executeCommandHandlers{
-	"rootmodules": executeCommandRootModulesHandler,
+const langServerPrefix = "terraform-ls."
+
+var commandPrefix string
+var handlers = make(executeCommandHandlers)
```
radeksimko, line 19: Technically the same server instance/process can serve multiple sessions, each of which may be initialized with a different prefix, wouldn't such shared state lead to race conditions?

radeksimko (approved):

langserver/handlers/execute_command.go
appilon: Can it? Do we support that or are you saying the spec in general does? Once the language server supports multiple workspaces i was going to deprecate this setting

appilon: After discussion I see the foolishness of my assumption that our languageserver currently only support 1 session. This is not true and this PR uses global memory which will not work out well. I will correct the implementation, fortunately there are no changes to the public API proposed, just internals.

langserver/handlers/execute_command.go
```diff
@@ -7,26 +7,45 @@ import (
...
 type executeCommandHandlers map[string]executeCommandHandler
 
+const langServerPrefix = "terraform-ls."
+
 var handlers = executeCommandHandlers{
-	"rootmodules": executeCommandRootModulesHandler,
+	prefix("rootmodules"): executeCommandRootModulesHandler,
+}
+
+func prefix(name string) string {
```
radeksimko, line 23: Nitpick: can we call this something like `prefixedCommandName`?

radeksimko (approved):
:shipit:

langserver/handlers/execute_command.go
appilon: I shall but I'm keeping it present tense `prefixCommandName()`

Issue #280, opened 2020-10-23
unknown block type: "terraform" (terraform 0.13 unsupported?)


kevincantu:
So I'm upgrading to 0.13 just as I've changed editors to VS Code and installed this extension...  And I see somewhat limited functionality.  While working out other adjustments to settings I've discovered this in the Output tab, that's somewhat startling:
```
2020/10/23 13:18:19 parser.go:178: parsing config block failed: unknown block type: "terraform"
```

Part of my `~/Library/Application Support/Code/User/settings.json`:
```json
...
    "terraform.languageServer": {
    
        "external": true,
        "pathToBinary": "",
        "args": [
            "serve"
        ],
        "maxNumberOfProblems": 100,
        "trace.server": "off"
    },
    "terraform-ls.rootModules": [
        "~/infra/env-labinfra",
    ],
    "[terraform]": {
        "editor.formatOnSave": true
    }
}
```

The output of the tab in the console corresponding to terraform-ls: that repo:
```
2020/10/23 13:17:23 server.go:568: Received 1 new requests
2020/10/23 13:17:23 server.go:174: Processing 1 requests
2020/10/23 13:17:23 server.go:265: Checking request for "textDocument/didChange": {"textDocument":{"uri":"file:///Users/kevincantu/infra/env-labinfra/backend.tf","version":32},"contentChanges":[{"range":{"start":{"line":24,"character":3},"end":{"line":24,"character":3}},"rangeLength":0,"text":"\n  "}]}
2020/10/23 13:17:23 rpc_logger.go:29: Incoming notification for "textDocument/didChange": {"textDocument":{"uri":"file:///Users/kevincantu/infra/env-labinfra/backend.tf","version":32},"contentChanges":[{"range":{"start":{"line":24,"character":3},"end":{"line":24,"character":3}},"rangeLength":0,"text":"\n  "}]}
2020/10/23 13:17:23 root_module_manager.go:149: direct root module lookup succeeded: /Users/kevincantu/infra/env-labinfra
2020/10/23 13:17:23 server.go:438: Posting server notification "textDocument/publishDiagnostics" {"uri":"file:///Users/kevincantu/infra/env-labinfra/backend.tf","diagnostics":[]}
2020/10/23 13:17:23 root_module.go:354: 1 provider references found for /Users/kevincantu/infra/env-labinfra
2020/10/23 13:17:23 server.go:568: Received 1 new requests
2020/10/23 13:17:23 server.go:174: Processing 1 requests
[...]

Issue #121, opened 2020-05-27
Add completion and hover support for built-in / reserved blocks/attributes


paultyng:
This TF version specific potentially (ie. `for_each`).

Providers:

- [ ] `alias`
- [ ] `version` (is this 0.12 only? not sure if it still exists in 0.13)

Resources:

- [ ]  `depends_on`
- [ ]  `count`
- [ ] `for_each`
- [ ] `provider`
- [ ] `lifecycle`
  - [ ] `create_before_destroy`
  - [ ] `prevent_destroy`
  - [ ] `ignore_changes`
- [ ] `provisioner`
- [ ] `connection`

Data Sources:

- [ ]  `depends_on`
- [ ]  `count`
- [ ] `for_each`
- [ ] `provider`

Additional top level blocks:

- [ ] `terraform`
- [ ] `variable` #122 
- [ ] `output` #123 
- [ ] `backend`

radeksimko:
I'm happy to track this as a tickable list, just FYI that all of the above depends on https://github.com/hashicorp/terraform-ls/issues/36 - i.e. having a proper parser for blocks which don't have an explicitly "parseable" schema.

PR #281, opened 2020-10-30
Replace internal decoder with hcl-lang

radeksimko:
## End-User Impact

The implementation generally enables a "progressively enhanced" experience. In practice this means:

 - Completion is available as soon as the server starts - so the user is never left with zero completion, but the data available is gradually enriched between these stages:
   - (1) [universal schema](https://github.com/hashicorp/terraform-schema/tree/main/internal/schema/universal) which covers any 0.12+ version (just `resource`, `data`, and `provider` for now, but we can expand later)
   - (2) version-specific core schema (once `terraform version` ran)
   - (3) provider schemas (once `terraform providers schema -json` ran)
 - Symbol names no longer use any kind of "address" (e.g. `data.random_string.refname`) but a block is represented the same way it would be in config (e.g. `data "random_string" "refname"`). This prevents any confusion when displaying "unreferable" blocks or attributes such as `terraform` and allows us to load symbols early, before any schema (which would define the addressing scheme) is available.

Screenshot 2020-11-02 at 12 39 26

## Implementation Notes

 - The upstream implementation actually already supports nested symbols (any blocks or attributes arbitrarily nested), but the `sourcegraph/go-lsp` library doesn't have the right type to expose this, so I left enabling this for another PR
 - We could now probably leverage the diagnostics generated from the initial parsing and avoid double parsing it, but I also left that for another PR
 - While the diff may seem big, most of it is actually deletions 💥  🎉   with the logic being upstreamed into `hcl-lang` and `terraform-schema`. The main "plumbing point" is basically `RootModule` methods.

## Future

This makes it easier to solve the following issues:

 - https://github.com/hashicorp/terraform-ls/issues/36
 - https://github.com/hashicorp/terraform-ls/issues/121
 - https://github.com/hashicorp/terraform-ls/issues/122
 - https://github.com/hashicorp/terraform-ls/issues/123
 - https://github.com/hashicorp/terraform-ls/issues/280

It should be mostly (ignoring RHS completion which isn't supported at all yet) a matter of adding the appropriate schema to `hashicorp/terraform-schema` and bumping the dependency here.

## Additional Links

https://github.com/hashicorp/hcl-lang
https://github.com/hashicorp/terraform-schema


internal/terraform/rootmodule/types.go
```diff
@@ -63,26 +65,30 @@ func (rms RootModules) Paths() []string {
 
 type RootModule interface {
```
appilon, line 66: Not a real criticism, but the interface is feeling "big"

internal/terraform/rootmodule/types.go
```diff
@@ -63,26 +65,30 @@ func (rms RootModules) Paths() []string {
...
-	IsSchemaLoaded() bool
+	UpdateProviderSchemaCache(ctx context.Context, lockFile File) error
+	IsProviderSchemaLoaded() bool
 	UpdateModuleManifest(manifestFile File) error
-	Parser() (lang.Parser, error)
-	IsParserLoaded() bool
+	Decoder() (*decoder.Decoder, error)
+	IsParsed() bool
+	ParseAndLoadFiles() error
+	ParsedDiagnostics() hcl.Diagnostics
```
appilon, line 82: Curious to see this, this is just a comment

langserver/handlers/did_change.go
```diff
@@ -64,13 +64,17 @@ func TextDocumentDidChange(ctx context.Context, params DidChangeTextDocumentPara
...
 	if len(rms) > 0 {
+		// Here we assume the first match is exact match
+		// TODO: Verify *here* this is in fact the case
 		rm := rms[0]
-		err := rm.ParseProviderReferences()
+		err := rm.ParseAndLoadFiles()
 		if err != nil {
 			return err
 		}
+		// TODO: publish/update diags from rm.ParsedDiagnostics()
```
appilon, line 77: hmm I guess this changes the diagnostics implementation, fortunately it's not many lines.

langserver/handlers/did_open.go
```diff
@@ -62,6 +62,21 @@ func (lh *logHandler) TextDocumentDidOpen(ctx context.Context, params lsp.DidOpe
 			Message: msg,
 		})
 	}
+
+	for _, c := range candidates {
```
appilon, line 66: Will need a crash course on this

appilon (approved):
Overall approve, but a pairing session will go a long way for me I think

radeksimko:
One potential bug was discovered, or at least an untested scenario during our call with @appilon where user requests completion in a file (or performs any other operation requiring schema) within a module that was not initialized. We need to make sure that the relevant schema is matched. AFAIK the current behaviour is just to provide reduced completion candidates (without any provider schema), which is incorrect.

I will need to look into this.

internal/filesystem/filesystem.go
```diff
@@ -191,7 +190,7 @@ func (fs *fsystem) ReadFile(name string) ([]byte, error) {
 
 func (fs *fsystem) ReadDir(name string) ([]os.FileInfo, error) {
 	memList, err := afero.ReadDir(fs.memFs, name)
-	if err != nil {
+	if err != nil && !os.IsNotExist(err) {
```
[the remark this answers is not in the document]
radeksimko: So this was actually a bug which wasn't visible before because the `ReadDir` codepath was not really exercised anywhere properly.

`Filesystem.ReadDir(name)` should always return a combination of filenames/directories in both virtual FS and on-disk FS. Not having this condition would cause this to error out when a new file was opened in editor and its parent directory doesn't exist (yet) on the disk.

radeksimko:
That bug was addressed in the last commit I believe, it wasn't actually _that_ trivial, but I tried my best to avoid changing too much code, given how much is already being changed.

That said I think some of the `rootmodule` package is due for some refactoring.

Basically the key change this PR is introducing is that what we previously understood as "root module" was a module that is always `terraform init`'d (with `.terraform` folder in it), but now because we need to be able to work with uninitialized modules too - to provide symbols and basic early completion - we now see "root module" as any module - initialized or not.

So at the very least the naming should probably change from "root modules" to "modules", but I'd prefer to leave that for another PR.

appilon (approved):

PR #288, opened 2020-11-10
Rootmodule diagnostics

### LINKED ISSUE #270: Generate all diagnostics
#269 Will deliver preliminary support for hcl parsing and diagnostics one file at a time. The spec however, details calculating all diagnostics for the project

>if a language has a project system (for example C#) diagnostics are not cleared when a file closes. When a project is opened all diagnostics for all files are recomputed (or read from a cache).

appilon:
This refactor based on the new decoder architecture brings the language server as close as possible to having project wide diagnostics #270 . Diagnostics are calculated for the entire rootmodule on file open and change. It does mean that a user that has several rootmodules open won't see diagnostics far a particular rootmodule (effectively folder) until a file is opened within a that rootmodule. 

Closes #270


internal/terraform/rootmodule/types.go
```diff
@@ -76,7 +76,7 @@ type RootModule interface {
 	MergedSchema() (*schema.BodySchema, error)
 	IsParsed() bool
 	ParseFiles() error
-	ParsedDiagnostics() hcl.Diagnostics
+	ParsedDiagnostics() map[string]hcl.Diagnostics
```
appilon, line 79: @radeksimko I had to track the diagnostics per file, as it played nicer with the LSP.

internal/lsp/diagnostics.go
```diff
@@ -17,3 +17,24 @@ func HCLSeverityToLSP(severity hcl.DiagnosticSeverity) lsp.DiagnosticSeverity {
...
+		// only process diagnostics with an attributable spot in the code
+		if hclDiag.Subject != nil {
+			msg := hclDiag.Summary
+			if hclDiag.Detail != "" {
+				msg += ": " + hclDiag.Detail
+			}
+			diags = append(diags, lsp.Diagnostic{
+				Range:    HCLRangeToLSP(*hclDiag.Subject),
+				Severity: HCLSeverityToLSP(hclDiag.Severity),
+				Source:   "Terraform",
```
appilon, line 34: Previously I had this set to `HCL`, however now I think since the diagnostics are coming from the decoder, they could be interpreted as not just syntax, but also provider schema specific? Therefore I thought saying the source was `Terraform` was most appropriate, let me know if that is an incorrect interpretation

langserver/diagnostics/diagnostics.go
```diff
@@ -3,81 +3,59 @@ package diagnostics
 import (
 	"context"
 	"log"
+	"path/filepath"
 	"sync"
 
 	"github.com/creachadair/jrpc2"
-	"github.com/hashicorp/hcl/v2/hclparse"
+	"github.com/hashicorp/hcl/v2"
+	"github.com/hashicorp/terraform-ls/internal/filesystem"
```
radeksimko, line 11: I guess it's a nitpick, but we should probably extract the `uri*` stuff into its own dedicated `internal/uri` package, it feels awkward to be importing the whole `filesystem` just because of that function.

internal/lsp/diagnostics.go
```diff
@@ -17,3 +17,24 @@ func HCLSeverityToLSP(severity hcl.DiagnosticSeverity) lsp.DiagnosticSeverity {
 	}
 	return sev
 }
+
+func HCLDiagsToLSP(hclDiags hcl.Diagnostics) []lsp.Diagnostic {
+	diags := []lsp.Diagnostic{}
+
+	for _, hclDiag := range hclDiags {
+		// only process diagnostics with an attributable spot in the code
```
radeksimko, line 25: Why do we omit subject-less diagnostics here? 🤔 

I'm not sure what kind of diagnostics these may be in practice, so it's hard for me to judge the impact, I just remember that in case of Terraform's own diags (e.g. exposed via `terraform validate` or by any command which parses code really) this is more common than we'd like.

For example providers are common source of such diags, because provider blocks are technically optional in the configuration: https://github.com/hashicorp/terraform/issues/20121

radeksimko (approved):
LGTM, just one question about the subject-less diags.

internal/lsp/diagnostics.go
appilon: I don't have any good reason really, I'll see what the UX is like for a non subject diag

langserver/diagnostics/diagnostics.go
appilon: I can do that in this PR, I think URIs (within the context of this project) is somewhat LSP specific, so could it move to `ilsp`?

appilon:
@radeksimko I moved `uri` to it's own pkg, I stopped suppressing subjectless diagnostics, and simply ensured a zero'd range is provided (So in theory a diagnostic will be attributed to the file, but will default to first line/column, diagnostics must have a range in LSP). PTAL

internal/lsp/diagnostics.go
radeksimko: The [spec](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#diagnostic) doesn't seem to go into much detail in describing this field, but it seems to suggest that each linter should be reported with its own `Source` - in our case I reckon that would apply to `tflint` and I would think potentially to `terraform validate` too?

The key question is how does the field affect the UX or anything in practice - I have no idea 🤷‍♂️ 
So let's keep `Terraform` there for now and we can revisit that decision later as we expand diagnostics support.

internal/lsp/range.go
```diff
@@ -16,6 +16,21 @@ func HCLRangeToLSP(hclRng hcl.Range) lsp.Range {
...
+	}
+	if r.Start.Line < 0 {
+		r.Start.Line = 0
+	}
+	if r.End.Character < 0 {
+		r.End.Character = 0
+	}
+	if r.End.Line < 0 {
+		r.End.Line = 0
+	}
```
radeksimko, line 31: Was this added as a reaction to any particular edge case you ran into?

radeksimko (approved):

internal/lsp/diagnostics.go
appilon: I'll switch it back then.

internal/lsp/range.go
appilon: HCL range start with index 1, in the event there is no subject, I passed a zero'd HCL range to the converter, this would result in negative values in the LSP conversion (since LSP range is index 0), this just ensures we don't send an invalid range

radeksimko (approved):

PR #293, opened 2020-11-12
Update contribution docs

radeksimko:
We don't use make for building anymore, as mentioned in https://discuss.hashicorp.com/t/how-to-build-terraform-ls/17302 and we no longer vendor dependencies.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -57,7 +57,7 @@ the language server
 
 ```
 cd terraform-ls
-make build
+go install
 ```
 
 Once the compilation process succeeds, you can find a `terraform-ls` executable in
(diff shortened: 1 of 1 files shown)


.github/CONTRIBUTING.md
```diff
@@ -57,7 +57,7 @@ the language server
 
 ```
 cd terraform-ls
-make build
+go install
```
rchl, line 60: I believe that `go install` just installs dependencies? You also need to run `go build` to build.

radeksimko: I believe that `go install` will also compile the whole project (including its dependencies), as per the help output:
```
~ go help install
usage: go install [-i] [build flags] [packages]

Install compiles and installs the packages named by the import paths.

Executables are installed in the directory named by the GOBIN environment
variable, which defaults to $GOPATH/bin or $HOME/go/bin if the GOPATH
environment variable is not set. Executables in $GOROOT
are installed in $GOROOT/bin or $GOTOOLDIR instead of $GOBIN.
```

There may be many more differences between the two, but in simple terms I'd describe `go install` as a "shortcut" for (more verbose) `go build -o $GOBIN/terraform-ls`.

"Installation" in Go from what I can tell is basically just a matter of moving the executable binary in `$GOBIN`.

FWIW This is also aligned with the [Terraform core instructions](https://github.com/hashicorp/terraform/blob/master/.github/CONTRIBUTING.md#terraform-clicore-development-environment), so if I'm wrong here then those instructions should also be updated.

rchl: It's my first time with Go so you are probably right .:)

appilon (approved):

Issue #117, opened 2020-05-26
Create/Use an alternative library for LSP structs


radeksimko:
### Current Version

```
0.2.1
```

### Use-cases

LS needs a library to represent LSP requests and responses and reflect the reality where clients and servers move at a different pace and hence may not use the exact same version of LSP which could be just _strictly_ unmarshaled.

### Attempted Solutions

Currently the LS uses a fork of [sourcegraph/go-lsp](https://github.com/sourcegraph/go-lsp). The main reason it uses the fork is because it allows us to tolerate missing fields in client requests - which could be missing because the client uses later version of LSP, or just because `go-lsp` has not codified that field yet - since all structs are maintained manually and often times fields are missing.

As pointed out by @njuCZ this solution however has a flaw. Because go-lsp uses embedded structs, custom unmarshalers [may not work as expected](https://github.com/golang/go/issues/39175) and fields will be ignored due to the fact that only unmarshalers of the embedded structs are called, which are unaware of "parent" struct fields - which are in turn ignored.

### Proposal

The go-lsp maintainer agreed that a better and more maintainable solution would be generating structs from the spec in the way gopls does it, instead of trying to keep up with the LSP spec.
https://github.com/sourcegraph/go-lsp/pull/8#issuecomment-633467535

gopls maintainers are hesitant to externalizing https://github.com/golang/tools/tree/master/internal/lsp/protocol just because they can't guarantee a stable API due to the code being generated from TypeScript "spec". However govim maintains their own copy of internal packages at https://github.com/govim/govim/tree/main/cmd/govim/internal/golang_org_x_tools, including the protocol library. I was told that most bugs related to breaking changes are dealt with during the generation phase and most of the remaining ones can be caught via compilation check https://github.com/govim/govim/blob/1b85ff518040c369babbb5a090eefc3fe37bd2f1/cmd/govim/gopls_server.go#L15

With that in mind I believe we could use the same approach as govim.

PR #311, opened 2020-11-19
Replace sourcegraph/go-lsp with gopls' internal/lsp/protocol

radeksimko:
Closes https://github.com/hashicorp/terraform-ls/issues/117

The most visible difference between the two libraries is probably the change from `int` to `float` which may look awkward especially in the context of line or column numbers, but this is mostly a side effect of the code being generated from TypeScript, where `number` may or may not have a fractional part. TypeScript does not differentiate so strictly between `float` and `int` unlike Go.

### DIFF
--- a/internal/lsp/hover.go
+++ b/internal/lsp/hover.go
@@ -3,7 +3,7 @@ package lsp
 import (
 	"github.com/hashicorp/hcl-lang/lang"
 	"github.com/hashicorp/terraform-ls/internal/mdplain"
-	"github.com/sourcegraph/go-lsp"
+	lsp "github.com/hashicorp/terraform-ls/internal/protocol"
 )
 
 func HoverData(data *lang.HoverData, cc lsp.TextDocumentClientCapabilities) lsp.Hover {
--- a/internal/lsp/hover.go
+++ b/internal/lsp/hover.go
@@ -2,25 +2,21 @@ package lsp
 
 import (
 	"github.com/hashicorp/hcl-lang/lang"
-	"github.com/hashicorp/terraform-ls/internal/mdplain"
 	lsp "github.com/hashicorp/terraform-ls/internal/protocol"
 )
 
 func HoverData(data *lang.HoverData, cc lsp.TextDocumentClientCapabilities) lsp.Hover {
-	mdSupported := cc.Hover != nil &&
-		len(cc.Hover.ContentFormat) > 0 &&
+	mdSupported := len(cc.Hover.ContentFormat) > 0 &&
 		cc.Hover.ContentFormat[0] == "markdown"
 
-	value := data.Content.Value
-	if data.Content.Kind == lang.MarkdownKind && !mdSupported {
-		value = mdplain.Clean(value)
-	}
-
-	content := lsp.RawMarkedString(value)
-	rng := HCLRangeToLSP(data.Range)
+	// In theory we should be sending lsp.MarkedString (for old clients)
+	// when len(cc.Hover.ContentFormat) == 0, but that's not possible
+	// without changing lsp.Hover.Content field type to interface{}
+	//
+	// We choose to follow gopls' approach (i.e. cut off old clients).
 
 	return lsp.Hover{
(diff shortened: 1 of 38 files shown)


internal/lsp/hover.go
```diff
@@ -2,25 +2,21 @@ package lsp
...
-		value = mdplain.Clean(value)
-	}
-
-	content := lsp.RawMarkedString(value)
-	rng := HCLRangeToLSP(data.Range)
+	// In theory we should be sending lsp.MarkedString (for old clients)
+	// when len(cc.Hover.ContentFormat) == 0, but that's not possible
+	// without changing lsp.Hover.Content field type to interface{}
+	//
+	// We choose to follow gopls' approach (i.e. cut off old clients).
```
radeksimko, line 16: This was discussed briefly in https://github.com/microsoft/language-server-protocol/issues/1144#issuecomment-730321055

langserver/handlers/cancel_request.go
```diff
@@ -2,13 +2,29 @@ package handlers
...
+		return err
+	}
+
 	jrpc2.CancelRequest(ctx, id)
 	return nil
 }
+
+func decodeRequestID(v interface{}) (string, error) {
+	if val, ok := v.(string); ok {
+		return val, nil
```
radeksimko, line 23: `sourcegraph/go-lsp` actually quoted string IDs

https://github.com/sourcegraph/go-lsp/blob/219e11d77f5d414b4fe5ba7e532b277fca34c1b4/jsonrpc2.go#L25

but I'm considering that a bug which we never encountered because of how rarely strings are used by language clients (most use numbers, which in case of `go-lsp` were interpreted as `uint64` and in gopls' as `float64`). Either way, if the client expects the server to quote the ID I would consider that a bug on the client side too, because from what I can tell the spec doesn't mention anything about quoting IDs:
https://microsoft.github.io/language-server-protocol/specifications/specification-current/#requestMessage

internal/protocol/gen/gen.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	filename, err := filepath.Abs(args[0])
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	url := fmt.Sprintf(urlFmt, goplsRef)
+
+	resp, err := http.Get(url)
```
radeksimko, line 37: This is far simpler way of extracting it compared to [govim's approach](https://github.com/govim/govim/tree/main/cmd/govim/internal/golang_org_x_tools), which we can apply just because we're only interested in the protocol structs and nothing else and thankfully everything is contained within that one file and that file does not have any non-stdlib imports.

It may need to be revisited if/when we get closer to extracting the JSON-RPC package before it's exported.

radeksimko:
CI tests will get triggered once https://github.com/hashicorp/terraform-ls/pull/313 is merged and this PR is rebased.

langserver/handlers/handlers_test.go
```diff
@@ -18,30 +17,38 @@ import (
...
+				"documentSymbolProvider": true,
+				"codeLensProvider": {},
+				"documentLinkProvider": {},
+				"documentFormattingProvider": true,
+				"documentOnTypeFormattingProvider": {
+					"firstTriggerCharacter": ""
+				},
 				"executeCommandProvider": {
-					"commands": %s
+					"commands": %q
```
appilon, line 41: This won't result in a proper array:
https://play.golang.org/p/Xjr6_HycaF9

It will be like `["one" "two"]` (missing comma). The reason this currently works is because we only have 1 command, so it is a proper array, but once it gets to 2 command it will not work

appilon (changes requested):

langserver/handlers/handlers_test.go
radeksimko: oh, I see, good point - I will fix that! 😅

radeksimko: Fixed

appilon (approved):

PR #312, opened 2020-11-20
Fix provider name/alias mixup

appilon:
Very minor mixup of the registry API response, should have used the name field. I believe all alias' on the registry is the same as the name for the official providers, however I discovered playing around, that partner (verified) providers don't have an alias, which led to a config error.

This PR still only packages official providers, but corrects the mistake. I did regenerate the provider schemas and I guess there are some recent releases that created a diff.


aeschright (approved):

radeksimko:
Am I right that this doesn't affect the UX, as it doesn't affect the exported JSON? In other words, does it need mentioning in the Changelog?

appilon:
@radeksimko correct it's not worth calling out

PR #318, opened 2020-11-23
Expose server version via LSP

### LINKED ISSUE #6: Add ServerInfo identification of the LS
https://microsoft.github.io/language-server-protocol/specifications/specification-current/#initialize

```ts
interface InitializeResult {
	/**
	 * The capabilities the language server provides.
	 */
	capabilities: ServerCapabilities;

	/**
	 * Information about the server.
	 *
	 * @since 3.15.0
	 */
	serverInfo?: {
		/**
		 * The name of the server as defined by the server.
		 */
		name: string;

		/**
		 * The server's version as defined by the server.
		 */
		version?: string;
	};
}
```

radeksimko:
Closes #6

### DIFF
--- a/langserver/handlers/service.go
+++ b/langserver/handlers/service.go
@@ -166,6 +166,11 @@ func (svc *service) Assigner() (jrpc2.Assigner, error) {
 			ctx = lsctx.WithRootModuleManager(ctx, svc.modMgr)
 			ctx = lsctx.WithRootModuleLoader(ctx, rmLoader)
 
+			version, ok := lsctx.LanguageServerVersion(svc.srvCtx)
+			if ok {
+				ctx = lsctx.WithLanguageServerVersion(ctx, version)
+			}
+
 			return handle(ctx, req, lh.Initialize)
 		},
 		"initialized": func(ctx context.Context, req *jrpc2.Request) (interface{}, error) {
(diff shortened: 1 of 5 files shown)


langserver/handlers/service.go
```diff
@@ -166,6 +166,11 @@ func (svc *service) Assigner() (jrpc2.Assigner, error) {
 			ctx = lsctx.WithRootModuleManager(ctx, svc.modMgr)
 			ctx = lsctx.WithRootModuleLoader(ctx, rmLoader)
 
+			version, ok := lsctx.LanguageServerVersion(svc.srvCtx)
```
appilon, line 169: I don't get the cycle here?

appilon (changes requested):

langserver/handlers/service.go
radeksimko: This is because the server starts with its own empty context, for better or worse:
https://github.com/creachadair/jrpc2/blob/3fb8081ec1bde4adba1b3a57dff458328a8ea769/server.go#L304

You could argue both sides - it's good for isolation - you don't get to inherit things by accident, but also it can make it more verbose 🤷‍♂️

appilon: ah okay!

appilon (approved):

PR #320, opened 2020-11-25
fix issue when hoverData is nil

### LINKED ISSUE #317: terraform-ls is crashing on hover over variable inside locals block
### Server Version

```
0.10.0
```

### Terraform Version

```
0.13.5
```

### Client Version

```
GNU Emacs 27.1 (build 1, x86_64-apple-darwin19.4.0, Carbon Version 162 AppKit 1894.4) of 2020-11-22 (using emacs-lsp client version 7.1.0)
```

```hcl
locals {
  test_domain = "test.employerawsdev.com"
}
```

### Log Output

https://gist.github.com/flurie/491859d2577409b22bc4a865e72970f0

### Expected Behavior

The server should handle the hover gracefully.
### Actual Behavior

The server handled the hover ungracefully.

### Steps to Reproduce

1. Create a single terraform file with a `locals` block and add a variable of your choosing.
2. Start the langserver and client in the specified editor, ensuring that `lsp-ui-doc-mode` is enabled
3. Create a hover event on the variable inside the `locals` block. I triggered a hover event with the cursor in emacs.

njuCZ:
fix #317

tried to return empty `lsp.Hover`, however, the vscode client will complain: `unsupported Markup content received. Kind is: `
Capture

The official recommendation is just return `null` from here: https://github.com/Microsoft/language-server-protocol/issues/261

generally two ways：
* make lsp protocol model fields be pointer type.
* in handle function, just return pointer type result

It seems we generate the lsp protocol model, so I just choose the second way.
everything seems fine:
image

### DIFF
--- a/internal/lsp/hover.go
+++ b/internal/lsp/hover.go
@@ -5,7 +5,10 @@ import (
 	lsp "github.com/hashicorp/terraform-ls/internal/protocol"
 )
 
-func HoverData(data *lang.HoverData, cc lsp.TextDocumentClientCapabilities) lsp.Hover {
+func HoverData(data *lang.HoverData, cc lsp.TextDocumentClientCapabilities) *lsp.Hover {
+	if data == nil {
+		return nil
+	}
 	mdSupported := len(cc.Hover.ContentFormat) > 0 &&
 		cc.Hover.ContentFormat[0] == "markdown"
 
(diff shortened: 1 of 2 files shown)


radeksimko (approved):
Good find and 👍  on the decision to use pointers here.

Thank you for the fix.

`nil` data is certainly a valid response and should be expected, in fact we should prefer that over the errors which are being returned for a number of edge cases, such as hovering over braces, but I haven't yet decided whether to just ignore them already in `hcl-lang` and never return such error in the first place or whether to filter out these errors and ignore them here.

PR #323, opened 2020-12-02
Support Terraform Validate Command

### LINKED ISSUE #27: Validate configuration and publish diagnostics

appilon:
This PR adds support to call terraform validate via executeCommand. It expects a `uri` to the dir to run it from to be passed. It will error if the specified dir is an uninitialized module.

Closes #27


langserver/handlers/did_save.go
```diff
@@ -0,0 +1,59 @@
...
+	"context"
+
+	lsctx "github.com/hashicorp/terraform-ls/internal/context"
+	ilsp "github.com/hashicorp/terraform-ls/internal/lsp"
+	lsp "github.com/hashicorp/terraform-ls/internal/protocol"
+	"github.com/hashicorp/terraform-ls/langserver/diagnostics"
+)
+
+func (h *logHandler) TextDocumentDidSave(ctx context.Context, params lsp.DidSaveTextDocumentParams) error {
+	fs, err := lsctx.DocumentStorage(ctx)
```
radeksimko, line 13: To comply with the protocol I believe we should be gating the execution of this handler on the [relevant client capability](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_didSave). Technically this would just uncover buggy clients which attempt to call `didSave` without actually announcing support for it, but we should still do it IMO.

Returning a custom error with a message suggesting a bug on the client side is probably best.

go.mod
```diff
@@ -7,12 +7,12 @@ require (
...
 	github.com/fsnotify/fsnotify v1.4.9
 	github.com/gammazero/workerpool v1.0.0
-	github.com/google/go-cmp v0.5.1
+	github.com/google/go-cmp v0.5.2
 	github.com/hashicorp/go-multierror v1.1.0
 	github.com/hashicorp/go-version v1.2.1
 	github.com/hashicorp/hcl-lang v0.0.0-20201116081236-948e43712a65
 	github.com/hashicorp/hcl/v2 v2.6.0
-	github.com/hashicorp/terraform-exec v0.11.1-0.20201007122305-ea2094d52cb5
+	github.com/hashicorp/terraform-exec v0.11.1-0.20201201215927-6e0fd8a457b8
```
radeksimko, line 15: I'm assuming this points to a branch https://github.com/hashicorp/terraform-exec/pull/68 which means we should probably merge that PR and point to a `main` revision or a version. Not important for a WIP PR, but I'd rather avoid pointing to a revision on a branch which may disappear on a `main` of terraform-ls.

i.e. I think this should be resolved before merging this PR.

appilon:
Just need a release of terraform-exec, as well as a PR for terraform-schema, then will bump the dependency references

internal/langserver/diagnostics/diagnostics.go
```diff
@@ -16,26 +16,30 @@ import (
...
+	diags  []lsp.Diagnostic
 }
 
 // Notifier is a type responsible for queueing hcl diagnostics to be converted
 // and sent to the client
 type Notifier struct {
+	logger         *log.Logger
 	sessCtx        context.Context
 	diags          chan diagContext
+	diagsCache     map[string]map[string][]lsp.Diagnostic
```
radeksimko, line 29: Just for readability, would you mind creating some custom types here and compose the final `diagsCache` type from these?

For example:
```go
type fileURI string
type diagnosticSource string
type fileDiagnostics map[diagnosticSource][]lsp.Diagnostic

diagsCache map[fileURI]fileDiagnostics

File: internal/langserver/diagnostics/diagnostics.go [hunk 4]
```diff
@@ -16,26 +16,30 @@ import (
...
 	closeDiagsOnce sync.Once
 }
 
 func NewNotifier(sessCtx context.Context, logger *log.Logger) *Notifier {
-	diags := make(chan diagContext, 50)
-	go notify(diags, logger)
-	return &Notifier{sessCtx: sessCtx, diags: diags}
+	n := &Notifier{logger: logger, sessCtx: sessCtx, diags: make(chan diagContext, 50), diagsCache: make(map[string]map[string][]lsp.Diagnostic)}
```
radeksimko, line 34: Nitpick

```suggestion
	n := &Notifier{
		logger: logger,
		sessCtx: sessCtx,
		diags: make(chan diagContext, 50),
		diagsCache: make(map[string]map[string][]lsp.Diagnostic),
	}
```

internal/langserver/diagnostics/diagnostics.go
```diff
@@ -16,26 +16,30 @@ import (
...
+	go n.notify()
+	return n
 }
 
-// Publish accepts a map of diagnostics per file and queues them for publishing
-func (n *Notifier) Publish(ctx context.Context, rmDir string, diags map[string]hcl.Diagnostics, source string) {
+// PublishHCLDiags accepts a map of hcl diagnostics per file and queues them for publishing.
+// A dir path is passed which is joined with the filename keys of the map, to form a file URI.
+// A source string is passed and set for each diagnostic, this is typically displayed in the client UI.
+func (n *Notifier) PublishHCLDiags(ctx context.Context, dir string, diags map[string]hcl.Diagnostics, source string) {
```
radeksimko, line 42:
```suggestion
func (n *Notifier) PublishHCLDiags(ctx context.Context, dirPath string, diags map[string]hcl.Diagnostics, source string) {
```
I know the comment above clarifies that it's a path and not URI, but I think it could be even more obvious from the name. 😉

internal/langserver/diagnostics/diagnostics.go
```diff
@@ -45,22 +49,37 @@ func (n *Notifier) Publish(ctx context.Context, rmDir string, diags map[string]h
...
 	}
 
-	if source == "" {
-		source = "Terraform"
-	}
-
-	for path, ds := range diags {
-		n.diags <- diagContext{ctx: ctx, diags: ds, source: source, uri: lsp.DocumentURI(uri.FromPath(filepath.Join(rmDir, path)))}
+	for filename, ds := range diags {
+		n.diags <- diagContext{ctx: ctx, source: source, diags: ilsp.HCLDiagsToLSP(ds, source), uri: lsp.DocumentURI(uri.FromPath(filepath.Join(dir, filename)))}
```
radeksimko, line 53: Nitpick
```suggestion
		n.diags <- diagContext{
			ctx: ctx, source: source,
			diags: ilsp.HCLDiagsToLSP(ds, source),
			uri: lsp.DocumentURI(uri.FromPath(filepath.Join(dir, filename))),
		}
```

internal/terraform/rootmodule/root_module.go
```diff
@@ -378,6 +378,67 @@ func (rm *rootModule) ExecuteTerraformInit(ctx context.Context) error {
 	return rm.tfExec.Init(ctx)
 }
 
+func (rm *rootModule) ExecuteTerraformValidate(ctx context.Context) (map[string]hcl.Diagnostics, error) {
```
radeksimko, line 381: I reckon we'll need to find a more flexible way of integrating tfexec and the `rootmodule` functionality as adding proxy method for every command probably won't scale. Which is certainly not a criticism of your change here - This is fine in this PR as it maintains consistency, I'm just thinking out loud about the future. 😄

internal/terraform/rootmodule/root_module.go
```diff
@@ -378,6 +378,67 @@ func (rm *rootModule) ExecuteTerraformInit(ctx context.Context) error {
...
+	validationDiags, err := rm.tfExec.Validate(ctx)
+	if err != nil {
+		return diagsMap, err
+	}
+
+	// tfjson.Diagnostic is a conversion of an internal diag to terraform core,
+	// tfdiags, which is effectively based on hcl.Diagnostic.
+	// This process is really just converting it back to hcl.Diagnotic
+	// since it is the defacto diagnostic type for our codebase currently
+	// https://github.com/hashicorp/terraform/blob/ae025248cc0712bf53c675dc2fe77af4276dd5cc/command/validate.go#L138
```
radeksimko, line 410: It may be convertible now, but even Terraform itself has its own type for interpreting diags, so I wouldn't be surprised if at some point we find out that `hcl` types just aren't sufficient to interpret `validate` diags.

I think that returning `[]tfjson.Diagnostic` and then merging it into `[]lsp.Diagnostic` outside of this package only when we need to, may be safer approach.

internal/langserver/handlers/command/validate.go
```diff
@@ -0,0 +1,49 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	rm, err := cf.RootModuleByPath(dh.Dir())
+	if err != nil {
+		return nil, err
+	}
+
+	wasInit, _ := rm.WasInitialized()
```
radeksimko, line 32: I remember we had a conversation about this, but what was the reasoning behind ignoring the error here? Would we not want to differentiate e.g. between a state where the module doesn't exist at all and when it exists, but isn't initialized?

radeksimko (approved):
I left you some mostly minor comments inline. Aside from these it LGTM.

internal/langserver/handlers/command/validate.go
appilon: Is there really a difference? As long as the URI is a folder, it is either initialized and a place where you can call terraform validate, or it isn't?

internal/terraform/rootmodule/root_module.go
appilon: Could we punt on that for a future PR? It helped to only work in `hcl` until it gets to the diagnostics notifier package. But if you felt strongly I could create a second `Publish` method.

radeksimko: Yeah, I suppose it's fine - I reckon we'll need to revisit this logic soon anyway to deliver on the promise of integrating tflint and somehow more "generalizing" the interface, or at least making it extendable.

So I am fairly confident this will need to change in the future, but I'm ok with leaving it for another PR. 👌

internal/langserver/handlers/command/validate.go
radeksimko: Well, in most cases there probably won't be a difference, but that's just because we rely on the client always calling `didOpen` on _a_ file from any folder that we run validate on.

In theory a client might integrate the command in such a way that it allows user to run it in folders which were never opened in the editor (but may in fact be initialized).

We should probably somehow account for this edge case and automatically load the module and make it known, but I think that raising a correct error is a solid start?

appilon: Ahh okay makes sense now, I will raise the error

PR #332, opened 2020-12-07
Add human readable name to rootmodules command API

aeschright:


radeksimko (approved):

internal/langserver/handlers/execute_command_rootmodules_test.go
```diff
@@ -169,15 +170,18 @@ func TestLangServer_workspaceExecuteCommand_rootmodules_multiple(t *testing.T) {
...
 				},
 				{
-					"uri": %q
+					"uri": %q,
+					"name": %q
 				}
 			]
 		}
-	}`, dev.URI(), prod.URI(), staging.URI()))
+	}`, dev.URI(), "env/dev", prod.URI(), "env/prod", staging.URI(), "env/staging"))
```
radeksimko, line 186: I think you might need to use `filepath.Join` here for cross-platform compatibility, as shown by the failing Windows test.

radeksimko (approved):
Except for the small path change LGTM.

PR #333, opened 2020-12-07
Use directory ilsp helper in init command

appilon:
This may not have any practical difference, but I did notice the handler was using a helper for creating a file handler from a file uri. This change uses the directory helper. I was concerned that the calling code was passing a file URI as well, which I thought was not the ideal interface we wanted to expose.

As it turns out, the calling code does in fact pass a directory URI, so I think the intention was for the `uri` to be a folder all along


radeksimko (approved):
Good catch! 👍

PR #335, opened 2020-12-09
fix: pass dir handler to init command

radeksimko:
This is just fixing an unforeseen side effect of https://github.com/hashicorp/terraform-ls/pull/333
Screenshot 2020-12-09 at 11 09 58

where we'd pass full file path to `askInitForEmptyRootModule` treating it like a directory path, when it isn't.

I guess the cleanest way of prevent similar bugs would be to have clearer separation between file handlers and dir handlers, instead of interfaces hiding parts of the "universal" handler.


appilon (approved):
Huh i swear i looked at this... my eyes lied to me

Issue #127, opened 2020-06-01
Add build environment details to version output


radeksimko:
### Current Version

```
v0.2.1
```

### Use-cases

As mentioned in https://github.com/hashicorp/terraform-ls/issues/88 there are cases when user may decide to compile the language server in their own environment.

While custom-compiled binaries and bugs stemming from differences in build environment should be fairly uncommon, they do exist and can be challenging to debug. As mentioned in the linked thread Go itself does not provide ways of prescribing the build environment yet - e.g. there is no way of guaranteeing what Go version, compiler (cgo/gc), nor compilation flags will be used.

Maintainers should at least have an easy way of discovering custom builds to aid with initial issue triaging and debugging.

### Attempted Solutions

Enforce building via makefile, which breaks cross-platform compatibility and can appear unnatural to seasoned Go developers who are used to run just native `go build/install`. It also enforces testing via makefile, which is suboptimal.

### Proposal

Add details about build environment to `version` output, e.g.

```
v0.2.1-dev
go1.14.3 darwin/amd64 (compiler: gc)
```

radeksimko:
@aeschright I think this would likely break the logic here https://github.com/hashicorp/vscode-terraform/blob/1c882b35bb52e834214fad186c167f8ecaf04804/src/languageServerInstaller.ts#L33 - it may be worth just cutting out the first line of output, alternatively you could wait for https://github.com/hashicorp/terraform-ls/issues/184

aeschright:
I think getting the JSON output in first would be my preference.

radeksimko:
As mentioned in https://github.com/hashicorp/terraform-ls/pull/341#discussion_r551386550 we should also consider reporting whether, or perhaps when particular `go generate` ran.

PR #341, opened 2020-12-18
Preload official+verified providers on release

### LINKED ISSUE #321: Include "Verified" provider schemas in the LS builds.
We should expand preloaded schemas from #302 to include both "Official" and "Verified" assuming the resulting binary is not prohibitively large and performance is not adversely affected.

appilon:
Refactor codebase to remove binary data from git, and have provider schemas generate during release process. Preloaded schemas will be nil during development. Verified partner providers now included on release (as well as official).

Closes #321


.github/workflows/release.yml
```diff
@@ -31,6 +31,9 @@ jobs:
           # TODO: Replace with go-version-from-file when it is supported
           # https://github.com/actions/setup-go/pull/62
           go-version: ${{ steps.go-version.outputs.content }}
+      -
+        name: Generate provider schemas
+        run: go generate ./internal/schemas
```
radeksimko, line 36: This makes me realize that users installing LS via `go install/get` will be missing these schemas.
This would likely also affect the community-maintained Homebrew formula https://formulae.brew.sh/formula/terraform-ls

Even if we (re)introduce `make build` target, which was removed as part of https://github.com/hashicorp/terraform-ls/pull/98 the remaining problem is that we have no way of communicating there is any makefile and that it may need to run. This was one of the reasons I initially actively prevented `go get` installations as per https://github.com/hashicorp/terraform-ls/issues/88#issuecomment-625780914 Also I would rather avoid adding a makefile target which isn't being actively used/tested in our CI.

I think the best way forward would be to keep encouraging people to just use our builds and make sure we can more easily identify unofficial builds as per https://github.com/hashicorp/terraform-ls/issues/127 if people experience bugs in those. Then builders of unofficial builds will either keep following changes in our CI configs and add `go generate` on their side, or produce builds with missing schema, which may be just fine? 🤷🏻‍♂️ 

We started baking the schemas in based on the feedback from VS Code extension users, all/most of which have _our_ builds of LS installed automatically - so it shouldn't affect that critical majority.

--- 

Longer term this may not be an issue though once we stop compiling schemas into the binary and start pulling them from the Registry.

--- 

All in all I would just add a note about this into the Changelog once merging it, to give folks a chance to spot it and adapt their build configs.

radeksimko (commented):
The build is currently failing, so I reckon that needs fixing, but otherwise the change looks good. I left you some more comments inline.

internal/schemas/gen/gen.go
```diff
@@ -199,5 +207,22 @@ func listProviders(tier string) ([]provider, error) {
...
+	return filter(response.Data), err
+}
+
+var ignore = map[string]bool{
+	"icinga2":  true,
+	"vthunder": true,
+	"sematext": true,
+}
+
+// certain providers are problematic/fail to download on terraform init
```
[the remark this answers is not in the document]
appilon: will do, they just aren't compatible with TF 0.12, or have a hash mismatch.

internal/schemas/version_output.go
```diff
@@ -1,6 +1,13 @@
 package schemas
 
+import "github.com/hashicorp/go-version"
+
 type VersionOutput struct {
 	CoreVersion string            `json:"core"`
 	Providers   map[string]string `json:"providers"`
 }
+
+type Version struct {
```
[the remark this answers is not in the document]
appilon: will do

.github/workflows/ci.yml
```diff
@@ -36,9 +36,12 @@ jobs:
           # TODO: Replace with go-version-from-file when it is supported
           # https://github.com/actions/setup-go/pull/62
           go-version: ${{ steps.go-version.outputs.content }}
+      - # since release builds specify the preload tag, we must run the generator
+        name: Generate provider schemas
+        run: go generate ./internal/schemas
       -
         name: Snapshot build (cross-platform)
-        uses: goreleaser/goreleaser-action@v1
+        uses: goreleaser/goreleaser-action@v2
```
radeksimko, line 44: Good catch! I completely forgot about this other action reference when updating the release job. 😅

.goreleaser.yml
```diff
@@ -36,6 +37,7 @@ builds:
     mod_timestamp: '{{ .CommitTimestamp }}'
     flags:
       - -trimpath
+      - -tags=preload
```
radeksimko, line 40: Nitpick: I would call it `preloadschema`

.github/workflows/ci.yml
```diff
@@ -36,9 +36,12 @@ jobs:
           # TODO: Replace with go-version-from-file when it is supported
           # https://github.com/actions/setup-go/pull/62
           go-version: ${{ steps.go-version.outputs.content }}
+      - # since release builds specify the preload tag, we must run the generator
```
radeksimko, line 39: I would say it's good to run this for each PR anyway as it's a way of testing the generator logic. 🤷🏻‍♂️

radeksimko (approved):

.github/workflows/release.yml
radeksimko: cc @chenrui333 ^ (I noticed you contributed some recent updates to [the Homebrew formula](https://github.com/Homebrew/homebrew-core/commits/b3a83e2fdec689b58a337a1b29eb0eb02b2aa1db/Formula/terraform-ls.rb))

.github/workflows/ci.yml
appilon: Okay, it just adds a few minutes to the build but if it's useful then great!

PR #372, opened 2021-01-08
Paginate registry results

appilon:
The verified partner providers was getting close to the max page size (100)

### DIFF
--- a/internal/schemas/gen/gen.go
+++ b/internal/schemas/gen/gen.go
@@ -50,14 +50,14 @@ func gen() error {
 	if err != nil {
 		return err
 	}
-	log.Printf("len=%d", len(providers))
+	log.Printf("fetched official providers: %d", len(providers))
 
 	log.Println("fetching verified partner providers from registry")
 	partnerProviders, err := listProviders("partner")
 	if err != nil {
 		return err
 	}
-	log.Printf("len=%d", len(partnerProviders))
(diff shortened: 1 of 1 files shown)


appilon:
I don't see how the changes I made would affect the tests, I suspect a race in some of our unit tests, specifically around shutdown, I can't recreate locally on darwin

aeschright (approved):
👍

internal/schemas/gen/gen.go
```diff
@@ -50,12 +50,14 @@ func gen() error {
 	if err != nil {
 		return err
 	}
+	log.Printf("len=%d", len(providers))
```
radeksimko, line 53:
```suggestion
	log.Printf("fetched official providers: %d", len(providers))
```

internal/schemas/gen/gen.go
```diff
@@ -50,12 +50,14 @@ func gen() error {
...
 
 	log.Println("fetching verified partner providers from registry")
 	partnerProviders, err := listProviders("partner")
 	if err != nil {
 		return err
 	}
+	log.Printf("len=%d", len(partnerProviders))
```
radeksimko, line 60:
```suggestion
	log.Printf("fetched partner providers: %d", len(partnerProviders))
```

radeksimko (approved):

Issue #354, opened 2021-01-05
Maintain global provider schema cache


radeksimko:
### Past/Current Situation

Currently (v0.11.0) we obtain provider schemas via the CLI (`terraform providers schema -json`) and track it for each module individually. This is suboptimal especially in cases of more root modules with duplicate provider versions as we store the same data more than once. More importantly we have no way of cross-referencing and reusing the data.

LS assumes that there's a single source of truth for the schema, when sources could in reality be combined.

Global tracking was initially avoided, because of the limited ways of identifying a provider (there was no concept of provider "address" until Terraform 0.13) and because the first architecture of LS didn't account for the complexity of module nesting nor it parsed provider requirements nor had capability to detect installed provider versions.

We have gathered feedback from GitHub and VS Code Marketplace in the past months where users were confused by various warnings we issue which can be attributed in one way or another to lack or inaccuracy of provider schema and practically very limited functionality of the language server. 

 - https://github.com/hashicorp/vscode-terraform/issues/524
 - https://github.com/hashicorp/terraform-ls/issues/221
 - https://github.com/hashicorp/vscode-terraform/issues/447
 - https://github.com/hashicorp/vscode-terraform/issues/329

We started pre-baking some schemas to address this, which provided at least some data to these users, but the solution is generally very naive (which is not a criticism, but a chosen trade-off) and it's becoming clear that we'll need a more flexible solution, especially once we start pulling schemas from more sources per https://github.com/hashicorp/terraform-ls/issues/193 and https://github.com/hashicorp/terraform-ls/issues/355

### Proposal

We now (since 0.13 was released) have _a mechanism_ for identifying providers, whether they are installed locally or come from a Registry (via an address).

It seems natural to therefore allow combining provider schemas, which we do to a limited degree already by falling back to preloaded schemas.

Here is what will likely need to be done:

1. create a global cache which can hold unique schema in a queryable cache
   - provider identity (e.g. `registry.terraform.io/hashicorp/azurerm`)
   - version (e.g. `2.40.0`)
   - source ? (e.g. path to the module where `terraform providers schema -json` ran, or preloaded JSON export)
2. parse provider requirements for any loaded module, so we know which versions to pick
3. create a mechanism for picking versions based on:
   - parsed requirements
   - source, e.g.
     - locally sourced schema is preferred
     - when two locally sourced schemas of different version are available, hierarchy proximity is considered, as also explained in https://github.com/hashicorp/terraform-ls/issues/180
[...]

PR #374, opened 2021-01-11
refactoring: root module -> module

radeksimko:
This PR pays back some technical debt. We used to care only about "root modules" (those modules which are `init`ed), but recently logic was introduced into the `rootmodule` package which deals with any kind of module, initalized or not. Therefore this renaming should provide more clarity for future readers.

There are still places where "root module" is being used in the context of "any module":

 - `rootmodules` executable LSP command
 - `rootModulePaths` settings option
 - any related docs for the above

I intentionally kept these intact as changing it will require a little more effort than just "find & replace" and I'd like to do it gracefully without breaking existing workflows.

The only user-visible change here is wording of the two messages as summarized below. ⬇️  Perhaps this still isn't the best wording that everyone will understand, but I thought it may be better than "root module"? 🤷🏻  Either way this is a wider problem to be solved as part of https://github.com/hashicorp/terraform-ls/issues/354 and some other related issues.

**Before**
- `Alternative root modules found for %s (%s), picked: %s. You can try setting paths to root modules explicitly in settings.`
- `No root module found for %q.`

**After**
- `Alternative schema source found for %s (%s), picked: %s. You can try setting path to a module explicitly in settings.`
- `No schema found for %q.`


internal/langserver/handlers/did_open.go
```diff
@@ -94,8 +94,8 @@ func (lh *logHandler) TextDocumentDidOpen(ctx context.Context, params lsp.DidOpe
 	if len(candidates) > 1 {
 		candidateDir := humanReadablePath(rootDir, candidates[0].Path())
 
-		msg := fmt.Sprintf("Alternative root modules found for %s (%s), picked: %s."+
-			" You can try setting paths to root modules explicitly in settings.",
+		msg := fmt.Sprintf("Alternative schema source found for %s (%s), picked: %s."+
+			" You can try setting path to a module explicitly in settings.",
```
aeschright, line 98:
```suggestion
			" You can create an explicit module path in your settings.",
```

internal/terraform/module/module_manager.go
```diff
@@ -175,15 +175,15 @@ func (rmm *rootModuleManager) rootModuleByPath(dir string) (*rootModule, bool) {
...
+func (rmm *moduleManager) ModuleCandidatesByPath(path string) Modules {
 	path = filepath.Clean(path)
 
-	candidates := make([]RootModule, 0)
+	candidates := make([]Module, 0)
 
 	// TODO: Follow symlinks (requires proper test data)
 
-	rm, foundPath := rmm.rootModuleByPath(path)
+	rm, foundPath := rmm.moduleByPath(path)
```
aeschright, line 186: Do you want the abbreviated variable name `rm` and `rmm` to match?

internal/terraform/module/module_manager_mock.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+type ModuleManagerMockInput struct {
+	Modules           map[string]*ModuleMock
+	TfExecutorFactory exec.ExecutorFactory
+}
+
+func NewModuleManagerMock(input *ModuleManagerMockInput) ModuleManagerFactory {
+	return func(fs filesystem.Filesystem) ModuleManager {
+		rmm := newModuleManager(fs)
```
aeschright, line 37: Same question about `rmm` here

aeschright (approved):
Looks straightforward, just a couple of comments.

PR #381, opened 2021-01-15
Add terraform plan command

aeschright:
TODO: finish tests


internal/langserver/handlers/command/plan.go
```diff
@@ -0,0 +1,53 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	wasInit, err := rm.WasInitialized()
+	if err != nil {
+		return nil, fmt.Errorf("error checking if %s was initialized: %s", dirUri, err)
+	}
+	if !wasInit {
+		return nil, fmt.Errorf("%s is not an initialized module, terraform validate cannot be called", dirUri)
```
appilon, line 38:
```suggestion
		return nil, fmt.Errorf("%s is not an initialized module, terraform plan cannot be called", dirUri)
```

internal/terraform/rootmodule/root_module.go
```diff
@@ -0,0 +1,792 @@
+package rootmodule
```
appilon, line 1: PR needs rebase

appilon (commented):
Overall looking good, however it's unclear how the plan UX will work, would we want to return the output in the response message?

radeksimko (commented):
As mentioned on Slack, I believe the process of regenerating the executor mock requires an extra step of `rm`ing the existing file.

```sh
rm internal/terraform/exec/exec_mock.go
go generate ./internal/terraform/exec
```
I'm unsure whether there is any better/different workflow which doesn't involve `rm`. I would perhaps expect mockery to be able to somehow handle this. Maybe it does and we just use in the wrong way. I just haven't had time to investigate.

Aside from that - I agree with Alex that UX is going to be hard here. We can't get away with just surfacing errors via popup anymore, like we did for `init`. Did you have any ideas on how we can/should surface the output?

Also you may want to do the following:

```sh
git checkout main
git rebase main command-plan
# resolve conflicts, if any arise
git push origin command-plan --force
```
Then you should end up with a cleaner git log.

aeschright:
I'll have to go through and squash some things, the problem was that rebasing from main can't clean up anything new I added, and I overlooked a couple of items.

aeschright:
I'm closing this because we've decided to run the command on the client side.

PR #385, opened 2021-01-25
watcher: Detect 'terraform init' from scratch

### LINKED ISSUE #222: Watch prospective root modules (to detect "terraform init" from scratch)
### Current Version

```
0.4.1
```

### Use-cases

The user may not start with initialized root modules and therefore they won't be discovered by the server.

Instinctively though users will follow the error message and try to `terraform init` while keeping the editor open. The server however won't detect these new files and won't load the schema from the newly initialized root module automatically.

The server only detects _changes_ for now.

### Attempted Solutions

Restart editor/server.

### Proposal

Identifying prospective root modules is difficult - in theory any folder with `*.tf` files, or even without may become a root module and be `init`-ed in the future.

We could however track prospective root modules by tracking any directory of a file which is received via `didOpen`, so we'd only actively track the folders of files which are currently open.

radeksimko:
This is a bigger change than perhaps expected but I was not able to come up with any decent way of splitting it up.

As mentioned in one in-line comment, there is currently no reliable and efficient way of watching a directory *recursively* in Go. See https://github.com/fsnotify/fsnotify/issues/18 There are some alternative solutions, but they either involve cgo or don't actually use any "event API" of the underlying OS but repeatedly do filewalk.

Previously we watched individual manifest and lock file of a module and we can't watch for *creation* of these files in a directory which doesn't exist yet. Therefore it was decided to instead watch the relevant parent directories and add relevant nested directories on creation and/or walk the hierarchy of the newly created dir. This creates even more places from where parts of module can be (re)loaded (walker, watcher, handlers), which would be hard to accommodate in the old design.

Therefore a mechanism for queueing all module operations was created.

This allows for a naive deduplication of operations (we don't run the same operation on the same module if it's already queued) and also allows us to report progress more easily in the future.

Finally this design also allows any caller to decide whether run an operation synchronously or asynchronously. e.g. We can wait for a module to be parsed in textDocument/didChange handler.

--- 

Closes #222


aeschright (approved):
👍

PR #386, opened 2021-01-27
cmd: Introduce version JSON output

### LINKED ISSUE #184: cmd/version: Introduce JSON output
### Current Version

```
0.3.2
```

### Use-cases

The VSCode extension and potentially other IDEs/extensions might want to manage the installation/upgrade of the language server. This would be little easier if they could read JSON output instead of line-based one.

### Attempted Solutions

https://github.com/hashicorp/vscode-terraform/blob/1c882b35bb52e834214fad186c167f8ecaf04804/src/languageServerInstaller.ts#L32-L33

### Proposal

Introduce `terraform-ls version -json` which has similar output to the recently introduced `terraform version -json`: https://github.com/hashicorp/terraform/pull/25252

radeksimko:
Closes #184

This unblocks https://github.com/hashicorp/terraform-ls/issues/127 (once JSON output is being used by the VS Code extension).

This introduces a new subcommand `terraform-ls version` with (optional) `-json` flag.
The existing `terraform-ls -version` still works and will continue working, but won't have the JSON flag - at least for now.

### DIFF
--- a/internal/cmd/version_command.go
+++ b/internal/cmd/version_command.go
@@ -1,0 +1,68 @@
+package cmd
+
+import (
+	"encoding/json"
+	"flag"
+	"fmt"
+	"strings"
+
+	"github.com/mitchellh/cli"
+)
+
+type VersionOutput struct {
+	Version string `json:"version"`
+}
+
+type VersionCommand struct {
+	Ui      cli.Ui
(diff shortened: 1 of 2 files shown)


aeschright (approved):
👍

PR #390, opened 2021-02-03
ilsp: Fix miscalculated semantic tokens

### LINKED ISSUE #387: Incorrect semantic highlighting for block braces
When `editor.semanticHighlighting.enabled` is set to `true` in vscode, it causes an issue with syntax highlighting for resource/data block declarations.

Specifically, the curly brace bracket(s) after `resource "type" "name"` is wrapped *in the same* syntax highlighting token that is used for `"type" "name"`, which causes the bracket to take on the wrong color in the editor. Hopefully that makes sense.

My guess is that the recent additions of this extension dipping its toes into using semantic tokens (#523) is causing this to occur.

### `editor.semanticHighlighting.enabled` ENABLED
Screen Shot 2021-02-02 at 11 23 33 AM

This also seems to happen for data resource declarations:
Screen Shot 2021-02-02 at 11 22 07 AM

### `editor.semanticHighlighting.enabled` DISABLED
Screen Shot 2021-02-02 at 11 24 36 AM

radeksimko:
Fixes #387 

This affects most cases where there is more than 2 tokens on the same line (typically blocks with 2 labels, like resources and data sources), in which case the `deltaStartChar` was being miscalculated.

### Before

Screenshot 2021-02-03 at 11 49 33

### After

Screenshot 2021-02-03 at 11 48 56

### DIFF
--- a/internal/lsp/token_encoder.go
+++ b/internal/lsp/token_encoder.go
@@ -78,13 +78,16 @@ func (te *TokenEncoder) encodeTokenOfIndex(i int) []float64 {
 	previousStartChar := 0
 	if i > 0 {
 		previousLine = te.Tokens[i-1].Range.End.Line - 1
-		previousStartChar = te.Tokens[i-1].Range.Start.Column - 1
+		currentLine := te.Tokens[i].Range.End.Line - 1
+		if currentLine == previousLine {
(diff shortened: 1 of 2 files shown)


aeschright (approved):
👍

PR #391, opened 2021-02-03
Utilize CPU better when loading modules

### LINKED ISSUE #186: Set a cap for auto-loading root modules

radeksimko:
Closes #186

While the number of operations in parallel is capped, we can still _overutilize_ the available CPU as per 
https://github.com/hashicorp/terraform-ls/blob/d31f6b4bdd99a3e6b889ac0b0a9aac5fc5a6bb84/internal/terraform/module/module_manager.go#L46

(prior to #385)

and now

https://github.com/hashicorp/terraform-ls/blob/f5ddf9ac740469f762b3aa9175b40187e0e44a7e/internal/terraform/module/module_loader.go#L25-L26

This patch therefore ensures  that except for single-core we never run more threads than there is CPUs available, which should provide some relief to users with many modules in the hierarchy.

### DIFF
--- a/internal/terraform/module/module_loader.go
+++ b/internal/terraform/module/module_loader.go
@@ -22,21 +22,44 @@ type moduleLoader struct {
 }
 
 func newModuleLoader() *moduleLoader {
-	nonPrioParallelism := 2 * runtime.NumCPU()
-	prioParallelism := 1 * runtime.NumCPU()
-
+	p := loaderParallelism(runtime.NumCPU())
 	plc, lc := int64(0), int64(0)
 	ml := &moduleLoader{
 		queue:              newModuleOpsQueue(),
 		logger:             defaultLogger,
-		nonPrioParallelism: int64(nonPrioParallelism),
-		prioParallelism:    int64(prioParallelism),
+		nonPrioParallelism: p.NonPriority,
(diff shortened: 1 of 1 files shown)


aeschright (approved):
LGTM!

PR #394, opened 2021-02-03
go mod tidy

radeksimko:
This removes now unused workerpool library. I just forgot to run `go mod tidy` as part of #385


radeksimko:
I'm going to merge without review, but it is still welcomed later

PR #396, opened 2021-02-04
module loader: Avoid deadlock

radeksimko:
This fixes a number of bugs I found while testing this both manually in VS Code and by running tests with high enough `-count`.

1. Deadlock could occur when retry of dispatch is attempted, because while `run()` is executing, another goroutine (e.g. one executing a different operation) could have finished and triggered another `tryDispatchingModuleOp` which then filled the channel and the next `tryDispatchingModuleOp` triggered from `run()` as a result of retry would block, therefore blocking the main loop.
   - This was resolved by running retry `tryDispatchingModuleOp` in an additional goroutine
2. `tryDispatchingModuleOp` can get triggered by many threads at the same time and its original implementation left an opportunity for race condition due to `queue.Peek()` relying on `queue.Len()` but the reliance was not guaranteed in any way. Additionally `Peek` could have "peeked" at the exact same operation more than once, which then resulted in that duplicate operation being dispatched.
   - This was resolved by getting rid of `Peek` method entirely and ensuring we never pop from an empty queue by utilizing the internal mutex. Retries should be rare enough that (especially with the 100ms delay) shouldn't cause concern wrt resource consumption by repeated push/pop.
3. `tryDispatchingModuleOp` was triggered from inside of `executeModuleOp`, but loading counters were being decremented via `defer` _after_, which could've caused the dispatch attempt to fail when the queue was full, just because the two operations were done in the wrong order.
   - This was resolved by getting rid of `defer` and calling `tryDispatchingModuleOp` from the main loop in `run()` where we can better control the order.

### DIFF
--- a/internal/terraform/module/module_loader.go
+++ b/internal/terraform/module/module_loader.go
@@ -85,17 +85,17 @@ func (ml *moduleLoader) run(ctx context.Context) {
 
 			if nextOp.Module.HasOpenFiles() && ml.prioCapacity() > 0 {
 				atomic.AddInt64(ml.prioLoadingCount, 1)
-				mod := ml.queue.PopOp()
 				go func(ml *moduleLoader) {
-					defer atomic.AddInt64(ml.prioLoadingCount, -1)
-					ml.executeModuleOp(ctx, mod)
+					ml.executeModuleOp(ctx, nextOp)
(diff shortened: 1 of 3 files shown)


aeschright (approved):
👍

PR #400, opened 2021-02-08
schema preload: Ensure provider names are unique

radeksimko:
This is to address a situation where two providers of the same `Name` (but different namespace) appear in the list.

Currently this would result in
```tf
    buildkite = {
      source = "buildkite/buildkite"
    }
  
    buildkite = {
      source = "jradtilbrook/buildkite"
    }
```
which is invalid due to the duplication.

Hence this patch just appends a number to make it unique, i.e. the above becomes

```tf
    buildkite-47 = {
      source = "buildkite/buildkite"
    }
  
    buildkite-48 = {
      source = "jradtilbrook/buildkite"
    }
```

### DIFF
--- a/internal/schemas/gen/gen.go
+++ b/internal/schemas/gen/gen.go
@@ -23,8 +23,8 @@ import (
 const terraformBlock = `terraform {
 	required_version = "~> 0.13"
   required_providers {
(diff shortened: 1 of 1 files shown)


aeschright (approved):
👍

Issue #190, opened 2020-06-26
Link to Terraform documentation and module repos


richeney:
# Link to Terraform documentation and module repos

## Enhancement summary

1. **Resource type and data type become follow links to the matching documentation pages.** 
1. **The values for a module's source argument also become follow links for known URI formats.**

This would be a major productivity benefit!

## Resource Type Example

For example, if I have 

```terraform
resource "azurerm_resource_group" "example" {
  name     = "example"
  location = "West Europe"
}
```

then make `azurerm_resource_group` a follow link (CTRL+click) to <https://www.terraform.io/docs/providers/azurerm/r/resource_group.html>.

If a data source then `azurerm_resource_group` would target <https://www.terraform.io/docs/providers/azurerm/d/resource_group.html>.

## Module Source Example

Example module:

```terraform
module "web_pool1_vms" {
  source            = "github.com/terraform-azurerm-modules/terraform-azurerm-linux-vm?ref=v0.1"
  defaults          = local.web_vm_defaults
  module_depends_on = [module.web_pool1_set]

  attachType = "ApplicationGateway"
  attach     = module.web_pool1_set.set_ids

  names           = ["web-pool1-a", "web-pool1-b", "web-pool1-c"]
  source_image_id = data.azurerm_image.ubuntu_18_04.id
}
```

Make `github.com/terraform-azurerm-modules/terraform-azurerm-linux-vm?ref=v0.1` a follow link to <https://github.com/terraform-azurerm-modules/terraform-azurerm-linux-vm?ref=v0.1>.

radeksimko:
Hi @richeney 
This sounds like a feature that could likely be implemented in the language server, rather than on the client side (such as VSCode extension) - that's why I decided to transfer it here.

We do have plans to build [hover support](https://github.com/hashicorp/terraform-ls/issues/87) - there is already [a draft PR](https://github.com/hashicorp/terraform-ls/pull/135) - that isn't adding a link to docs yet, but I would imagine that would be a place where we can add it, similar to how gopls does it:

![Screenshot 2020-06-26 at 16 14 56](https://user-images.githubusercontent.com/287584/85872822-317f7400-b7c8-11ea-9491-d50751687818.png)

[...]

PR #402, opened 2021-02-09
Add docs links to document and hover

radeksimko:
Depends on https://github.com/hashicorp/hcl-lang/pull/8
Depends on https://github.com/hashicorp/terraform-schema/pull/10

--- 

This adds links to relevant documentation for providers via hover data and `documentLink` method, which then results in the following UX in VS Code:

Screenshot 2021-02-09 at 16 19 47

Maybe this is not the best UX, but we don't have any control over how clients (such as VS Code) use the `documentLink` and if at all. Sublime Text for example doesn't implement it at all, so we are left with just hover:

Screenshot 2021-02-09 at 16 54 20

I don't have strong opinion on how we link to docs UX-wise, but I think we could just roll it out and see how people (dis)like it.

Relatedly the UTM parameters will help us identify any click-throughs to the Registry using any of these two features. All URLs have these parameters appended:
 - `utm_source=terraform-ls`
 - `utm_content` - e.g.`documentLink` or `documentHover`
 - `utm_media` - e.g. `vscode` or `Sublime Text LSP`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,11 +10,11 @@ require (
 	github.com/google/uuid v1.2.0
 	github.com/hashicorp/go-multierror v1.1.0
 	github.com/hashicorp/go-version v1.2.1
-	github.com/hashicorp/hcl-lang v0.0.0-20201209145723-0c4061e492db
+	github.com/hashicorp/hcl-lang v0.0.0-20210213170001-bd00c3f68680
 	github.com/hashicorp/hcl/v2 v2.8.2
 	github.com/hashicorp/terraform-exec v0.13.0
 	github.com/hashicorp/terraform-json v0.8.0
-	github.com/hashicorp/terraform-schema v0.0.0-20201208163444-44d0347ab290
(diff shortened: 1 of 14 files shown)


aeschright (approved):
👍

nfx:
this PR does not allow going to resource documentation, only provider.

mercuriete:
this only works on provider block
but not in required_providers block:

example:

versions.tf
```
terraform {
  required_version = ">= 0.14.5, <= 0.14.7"
  required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = "3.29.1"
    }
  }
}
```

radeksimko:
Correct, this is the current limitation we plan to address. I'm sorry for not being clear about it in the PR description.

Feel free to follow https://github.com/hashicorp/terraform-ls/issues/190 for resources, data sources and modules, or https://github.com/hashicorp/terraform-ls/issues/440 for `required_providers` block.

PR #403, opened 2021-02-09
watcher: Refresh versions when plugin lockfile changes

radeksimko:
When lock file changes we always reload the schema but it is likely that provider versions may have also changed and therefore we need to reload that too.


aeschright (approved):
👍

PR #407, opened 2021-02-16
Add goversion, goos, and goarch to version command

cappyzawa:
fix: https://github.com/hashicorp/terraform-ls/issues/127

```bash
# dev version
$ terraform-ls version
0.0.0-dev
$ terraform-ls version -json
{
  "version": "0.0.0-dev"
}

# release version
$ ./dist/terraform-ls_darwin_amd64/terraform-ls version
0.13.0-snapshot.6b81b78
go1.15.2 darwin/amd64

$ ./dist/terraform-ls_darwin_amd64/terraform-ls version -json
{
  "version": "0.13.0-snapshot.6b81b78",
  "go_version": "1.15.2",
  "go_os": "darwin",
  "go_arch": "amd64"
}
```

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -14,7 +14,7 @@ builds:
       - -trimpath
       - -tags=preloadschema
     ldflags:
-      - '-s -w -X "main.version={{ .RawVersion }}" -X "main.prerelease={{ if .IsSnapshot }}snapshot.{{ .ShortCommit }}{{ else }}{{ .Prerelease }}{{ end }}" -X "main.buildGoVersion={{ .Env.GOVERSION }}" -X "main.buildGoOS={{ .Os }}" -X "main.buildGoArch={{ .Arch }}"'
+      - '-s -w -X "main.version={{ .RawVersion }}" -X "main.prerelease={{ if .IsSnapshot }}snapshot.{{ .ShortCommit }}{{ else }}{{ .Prerelease }}{{ end }}"'
     goarch:
       - '386'
       - amd64
@@ -39,7 +39,7 @@ builds:
       - -trimpath
       - -tags=preloadschema
     ldflags:
-      - '-s -w -X "main.version={{ .RawVersion }}" -X "main.prerelease={{ if .IsSnapshot }}snapshot.{{ .ShortCommit }}{{ else }}{{ .Prerelease }}{{ end }}" -X "main.buildGoVersion={{ .Env.GOVERSION }}" -X "main.buildGoOS={{ .Os }}" -X "main.buildGoArch={{ .Arch }}"'
+      - '-s -w -X "main.version={{ .RawVersion }}" -X "main.prerelease={{ if .IsSnapshot }}snapshot.{{ .ShortCommit }}{{ else }}{{ .Prerelease }}{{ end }}"'
     goarch:
       - '386'
       - amd64
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -14,7 +14,7 @@ builds:
       - -trimpath
       - -tags=preloadschema
     ldflags:
-      - '-s -w -X "main.version={{ .RawVersion }}" -X "main.prerelease={{ if .IsSnapshot }}snapshot.{{ .ShortCommit }}{{ else }}{{ .Prerelease }}{{ end }}"'
+      - '-s -w -X "main.version={{ .RawVersion }}" -X "main.prerelease={{ if .IsSnapshot }}snapshot.{{ .ShortCommit }}{{ else }}{{ .Prerelease }}{{ end }}" -X "main.buildGoVersion={{ .Env.GOVERSION }}" -X "main.buildGoOS={{ .Os }}" -X "main.buildGoArch={{ .Arch }}"'
     goarch:
       - '386'
       - amd64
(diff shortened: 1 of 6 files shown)


cappyzawa:
CI will pass if https://github.com/hashicorp/terraform-ls/pull/406  is merged 😄

radeksimko:
Thank you for the PR @cappyzawa 

Before we can merge this however we will need to take at least two more steps:
 - Add these new fields to the [`version -json` output](https://github.com/hashicorp/terraform-ls/blob/f9eb59f7d5eeb23c3f0af94176961d6189bdcf1a/internal/cmd/version_command.go#L12-L14)
 - Ensure that VS Code extension, [which uses `-version` output](https://github.com/hashicorp/vscode-terraform/blob/cd7dfed427d31b6982bd7f7d426fa14927f07305/src/languageServerInstaller.ts#L12) is switched to `version -json` and therefore this change doesn't break it.

cappyzawa:
@radeksimko Thank you for checking.

Should I add fields to the [version -json output](https://github.com/hashicorp/terraform-ls/blob/f9eb59f7d5eeb23c3f0af94176961d6189bdcf1a/internal/cmd/version_command.go#L12-L14)  in this PR?

cappyzawa:
The previous code did not support the json flag, so I fixed it (https://github.com/hashicorp/terraform-ls/pull/407/commits/c09af642bae9ec1b7804696caffda2c5930bfbf8).

Currently, those commands are output as follows.
```
# dev version
$ terraform-ls version
0.0.0-dev
$  terraform-ls version -json
{
  "version": "0.0.0-dev"
}

# release version
$ ./dist/terraform-ls_darwin_amd64/terraform-ls version
0.13.0-snapshot.6b81b78
go1.15.2 darwin/amd64

$ ./dist/terraform-ls_darwin_amd64/terraform-ls version -json
{
  "version": "0.13.0-snapshot.6b81b78",
  "go_version": "1.15.2",
  "go_os": "darwin",
  "go_arch": "amd64"
}
```

The field name in the json is tentative, so I will fix it if you point it out.

internal/cmd/version_command.go
```diff
@@ -11,15 +11,26 @@ import (
 
 type VersionOutput struct {
 	Version string `json:"version"`
+
+	BuildGoVersion string `json:"go_version,omitempty"`
+	BuildGoOS      string `json:"go_os,omitempty"`
+	BuildGoArch    string `json:"go_arch,omitempty"`
```
radeksimko, line 17: Nitpick: I know it doesn't affect the output, but would you mind omitting the `Build` prefix from the fields here? I think the purpose of each variable is clear without it.

radeksimko: Also do you mind adding one more field

```go
Compiler string `json:"compiler,omitempty"`
```
which would use `runtime.Compiler` constant https://golang.org/pkg/runtime/#pkg-constants

internal/cmd/version_command.go
```diff
@@ -50,7 +65,12 @@ func (c *VersionCommand) Run(args []string) int {
 		return 0
 	}
 
-	c.Ui.Output(string(c.Version))
+	ver := string(c.Version)
+	if output.BuildGoVersion != "" && output.BuildGoOS != "" && output.BuildGoArch != "" {
+		ver = fmt.Sprintf("%s\ngo%s %s/%s", c.Version, output.BuildGoVersion, output.BuildGoOS, output.BuildGoArch)
```
radeksimko, line 70: As mentioned, the existing output format may be used by other tools which parse it (because they haven't yet switched to parsing the JSON output) and may expect the first line to only have the version, so perhaps we can avoid at least _some_ breaking changes by moving the additional fields below? e.g.

```
0.11.0
go: go1.15.2
platform: darwin/amd64
compiler: gc
```

main.go
```diff
@@ -47,6 +47,11 @@ func main() {
 			return &cmd.VersionCommand{
 				Ui:      ui,
 				Version: VersionString(),
+				BuildInfo: &cmd.BuildInfo{
```
radeksimko, line 50: As mentioned we may not need to pass any extra arguments here, if we just source it all from the `runtime` package.

radeksimko (commented):
Thank you for the update as well, I left you some comments in-line with regards to the format and some other things.

Once these are addressed, I reckon it will be in a mergeable state, but we would still need to update the VS Code extension logic prior to shipping this PR - I can send a PR with a patch there.

.goreleaser.yml
```diff
@@ -39,7 +39,7 @@ builds:
       - -trimpath
       - -tags=preloadschema
     ldflags:
-      - '-s -w -X "main.version={{ .RawVersion }}" -X "main.prerelease={{ if .IsSnapshot }}snapshot.{{ .ShortCommit }}{{ else }}{{ .Prerelease }}{{ end }}"'
+      - '-s -w -X "main.version={{ .RawVersion }}" -X "main.prerelease={{ if .IsSnapshot }}snapshot.{{ .ShortCommit }}{{ else }}{{ .Prerelease }}{{ end }}" -X "main.buildGoVersion={{ .Env.GOVERSION }}" -X "main.buildGoOS={{ .Os }}" -X "main.buildGoArch={{ .Arch }}"'
```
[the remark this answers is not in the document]
cappyzawa: I see.
I will fix it.

radeksimko:
@cappyzawa FYI: I just opened https://github.com/hashicorp/vscode-terraform/pull/560 to address the extension compatibility problem.

cappyzawa:
@radeksimko Thanks for your kind suggestion 👍 

I fixed.

Outputs of the `version`  command are described in this PR summary.

radeksimko:
We're going to cut and distribute a release of the extension (with the mentioned patch) possibly within the next week, so I'm going to put this PR on hold until then, but you can rebase it in the meantime if you like. All tests should now pass as I merged #406

radeksimko:
Just FYI https://github.com/hashicorp/vscode-terraform/pull/560 was released yesterday (on Monday), so I would let that release be distributed to users over the week and consider merging this some time next week, to reduce the risk of breaking someone's installation.

PR #420, opened 2021-02-24
textDocument/documentSymbol: Support nested symbols

### LINKED ISSUE #290: Provide navigation for nested symbols (blocks/attributes)
### Current Version
```
0.9.0
```

### Use-cases

Navigation could be easier through more complex and nested configs if the user could also see and navigate through all the nested structures - blocks and attributes.

### Proposal

Leverage `NestedSymbols()` of the existing struct as implemented in `hcl-lang`:
https://pkg.go.dev/github.com/hashicorp/hcl-lang@v0.0.0-20201110071249-4e412924f52b/decoder#Symbol

### Related LSP methods

https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_documentSymbol

radeksimko:
Closes #290

In terms of supported symbol kinds this is still limited to only primitive types for now. It would be probably weird not to display an attribute at all just because we don't know its type, so we fall back to a generic `Variable` kind in such cases.

https://github.com/hashicorp/hcl-lang/pull/16 brings some initial expression support which we can later use to improve the symbol navigation accuracy.

### Before

Screenshot 2021-02-24 at 11 28 42

### After

Screenshot 2021-02-24 at 11 29 24


aeschright (approved):
LGTM

PR #430, opened 2021-03-05
Exclude brightbox preloaded schemas

radeksimko:
As discovered here brightbox provider releases seem to be broken somehow:

https://github.com/hashicorp/terraform-ls/runs/2038251539?check_suite_focus=true#step:6:51


radeksimko:
This appears to be no longer needed.

PR #454, opened 2021-04-12
Introduce global provider schema cache

### LINKED ISSUE #354: Maintain global provider schema cache

radeksimko:
Fixes #354

Depends on https://github.com/hashicorp/hcl-lang/pull/32

Depends on https://github.com/hashicorp/terraform-schema/pull/35


aeschright (approved):
🚀

PR #472, opened 2021-04-27
state: Copy data on write - not on read

### LINKED ISSUE #471: performance: Avoid copying-on-read for memdb structures
### Server Version

```
0.16.0
```

```hcl
terraform {
  required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = "~> 3.1"
    }
  }
}

data "aws_ami" "ubuntu" {
  most_recent = true

  filter {
    name   = "name"
    values = ["ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-*"]
  }

  filter {
    name   = "virtualization-type"
    values = ["hvm"]
  }

  owners = ["099720109477"] # Canonical
}

resource "aws_instance" "web" {
  # <-- request completion HERE
  instance_type = "t3.micro"
}
```

### Expected Behavior

Pulling data from memdb takes _reasonable_ time (sub ms).

### Actual Behavior

Most actions (completion, hover etc.) are delayed depending on the size of affected provider schema(s) or module data size.

### Context

Most of the time in the hot codepath for e.g. `ModuleByPath` or `ProviderSchema` is spent on making a copy of the data before returning it, e.g.

https://github.com/hashicorp/terraform-ls/blob/77c26dc53decf7bba31f38b218efe007eeedc052/internal/state/module.go#L120-L127

https://github.com/hashicorp/terraform-ls/blob/77c26dc53decf7bba31f38b218efe007eeedc052/internal/state/provider_schema.go#L259-L263

### Steps to Reproduce

0. Use config above
1. Request any schema-dependent data, e.g. completion or hover
2. Watch UX and log
3. See delay ~500-1000ms

radeksimko:
Depends on https://github.com/hashicorp/terraform-schema/pull/39

Closes #471

--- 

### Before (current `main`)

```
go test -run=XXX -bench=. ./internal/state
```
```
goos: darwin
goarch: amd64
pkg: github.com/hashicorp/terraform-ls/internal/state
cpu: Intel(R) Core(TM) i7-7567U CPU @ 3.50GHz
BenchmarkModuleByPath-4     	    3937	    268528 ns/op
BenchmarkProviderSchema-4   	   58588	     20559 ns/op
PASS
ok  	github.com/hashicorp/terraform-ls/internal/state	3.495s
```

### After (branch linked to this PR)

```
go test -run=XXX -bench=. ./internal/state
```
```
goos: darwin
goarch: amd64
pkg: github.com/hashicorp/terraform-ls/internal/state
cpu: Intel(R) Core(TM) i7-7567U CPU @ 3.50GHz
BenchmarkModuleByPath-4     	    9272	    116078 ns/op
BenchmarkProviderSchema-4   	  104325	     11234 ns/op
PASS
ok  	github.com/hashicorp/terraform-ls/internal/state	2.537s
```

### DIFF
--- a/internal/state/provider_schema_test.go
+++ b/internal/state/provider_schema_test.go
@@ -758,6 +758,63 @@ func TestStateStore_ListSchemas(t *testing.T) {
 	}
 }
 
+// BenchmarkProviderSchema exercises the hot path for most handlers which require schema
+func BenchmarkProviderSchema(b *testing.B) {
+	s, err := NewStateStore()
+	if err != nil {
+		b.Fatal(err)
+	}
+
+	schemas := []*ProviderSchema{
+		{
+			tfaddr.Provider{
+				Hostname:  tfaddr.DefaultRegistryHost,
+				Namespace: "hashicorp",
+				Type:      "blah",
+			},
+			testVersion(b, "0.9.0"),
+			PreloadedSchemaSource{},
+			&tfschema.ProviderSchema{
+				Provider: &schema.BodySchema{
+					Description: lang.PlainText("preload: hashicorp/blah 0.9.0"),
+				},
+			},
+		},
+		{
+			tfaddr.Provider{
+				Hostname:  tfaddr.DefaultRegistryHost,
+				Namespace: "hashicorp",
+				Type:      "aws",
+			},
+			testVersion(b, "0.9.0"),
+			PreloadedSchemaSource{},
+			&tfschema.ProviderSchema{
+				Provider: &schema.BodySchema{
+					Description: lang.PlainText("preload: hashicorp/aws 0.9.0"),
+				},
+			},
+		},
+	}
+	for _, ps := range schemas {
+		addAnySchema(b, s.ProviderSchemas, s.Modules, ps)
+	}
+
+	expectedPs := &tfschema.ProviderSchema{
+		Provider: &schema.BodySchema{
+			Description: lang.PlainText("preload: hashicorp/aws 0.9.0"),
+		},
+	}
+	for n := 0; n < b.N; n++ {
+		foundPs, err := s.ProviderSchemas.ProviderSchema("/test", tfaddr.NewDefaultProvider("aws"), testConstraint(b, "0.9.0"))
+		if err != nil {
+			b.Fatal(err)
+		}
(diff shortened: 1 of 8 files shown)


internal/state/provider_schema_test.go
```diff
@@ -439,12 +439,16 @@ func TestStateStore_ProviderSchema_legacyAddress_exactMatch(t *testing.T) {
 			expectedDescription, ps.Provider.Description.Value)
 	}
 
-	// Check that detail has "fixed" legacy namespace
```
aeschright, line 442: What's the reason behind this change?

internal/state/provider_schema_test.go
```diff
@@ -780,6 +846,7 @@ func addAnySchema(t *testing.T, ss *ProviderSchemaStore, ms *ModuleStore, ps *Pr
 		err := ss.AddLocalSchema(s.ModulePath, ps.Address, ps.Schema)
 		if err != nil {
 			t.Fatal(err)
+
```
aeschright, line 849: Extra newline?

internal/state/provider_schema_test.go
aeschright: Never mind, found it on the other PR.

aeschright (approved):

internal/state/provider_schema_test.go
radeksimko: Good question.

The idea _was_ that if user is still on 0.12 we'd pull schemas attributed to legacy addresses, which can still be used for completion, hover etc., but linking to docs for example is much trickier as the Registry doesn't have a stable URL for legacy addresses. This is why we'd have previously attempted to "fix" the namespace as linking to `hashicorp/aws` is possible, when linking to `-/aws` is not.

However this required modification of the schema in-place essentially, as we place the generated links inside the schema. This is no longer a viable option as we'd have to keep copying the schema on read.

https://github.com/hashicorp/terraform-schema/pull/39 has more radical solution to that problem

Basically we can keep displaying `-/aws 3.0.0` in the detail as that's still better than not displaying anything, but we omit the links entirely.

internal/state/provider_schema_test.go
radeksimko: Good catch!

PR #474, opened 2021-04-29
ci: Fix broken key importing

radeksimko:
gpg apparently uses tty for printing warnings (in this case likely about trust level for the new key)

Because there's no valid tty in GitHub Actions (or any other CI) we reflect that by passing extra flags to gpg.

This should address the following error during release:

```
gpg: key 34365D9472D7468F/B0B441097685B676: error sending to agent: Inappropriate ioctl for device
gpg: error reading '[stdin]': Inappropriate ioctl for device
gpg: import from '[stdin]' failed: Inappropriate ioctl for device
```

If this doesn't work then we can consider swapping that `echo | gpg` command for https://github.com/hashicorp/ghaction-import-gpg which is used by most Terraform providers AFAIK.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -62,7 +62,7 @@ jobs:
           hc-releases version
       -
         name: Import key for archive signing
-        run: echo -e "${{ secrets.GPG_PRIVATE_KEY }}" | gpg --import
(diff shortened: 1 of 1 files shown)


radeksimko:
I'm going to merge this now (without review) to try unblocking a release which is due, but post-merge reviews are welcomed.

Issue #93, opened 2018-12-21
Completion of module variables in module block


pecigonzalo:
I could not find any issues or references to it, but does autocompletion work from modules? We use a lot of them, and it does not seem to work for us.

radeksimko:
Hi @pecigonzalo 
I'm aware that this may be a bit too late to ask given that you filed the issue in December 2018, but do you remember what the context was?

i.e. what do you mean by "completion from modules"?

There is a few aspects of modules where completion comes to mind:

 - opening module's root in IDE and getting completion inside configuration that belongs to the module
   - in theory this should already work out of the box, but it's probably less common for folks to `terraform init` inside the module directory and more importantly you'd have to open the folder explicitly, _not_ it's parent folder where the module is referenced, because the server needs the relevant providers to be downloaded, so it can obtain the schema (which in turn gives it the completion data) - see https://github.com/hashicorp/terraform-ls/issues/83 to read more about how we plan to address that
 - completing arguments within `module` block (module reference)
   - this doesn't work yet, and I think it generally falls under https://github.com/hashicorp/terraform-ls/issues/36
 - completing `source` of `module` block (module reference)
   - this doesn't work and I'm unsure what the ideal UX be here. We could probably implement this as Terraform Registry search for modules there, but we might need to also account for private module Registries which are firewalled-off and I'm not sure if we have a mechanism for discovering the existence of such registry, so we can search in it

pecigonzalo:
Unfortunately no and I was not really good at describing it initially, ill give it a look tomorrow and try to repro.

pecigonzalo:
@radeksimko I got around to remember what this was about. As we use tons of modules, we wanted to be able to autocomplete module outputs when typing them.

Lets say I have module for a vpc like
```hcl
resource "aws_vpc" "default" {
  [...]
}

output "vpc_id" {
  value       = "${aws_vpc.default.id}"
  description = "The VPC ID"
}
```

and some code like
```hcl
module "vpc" {
  [...]
}

output "id" {
  value = module.vpc.vpc_id
}
```


When typing `module.vpc.` I would like to get suggestions for the possible attributes of that module, in this example `vpc_id`.
Ideally when consuming the module and writing inside the module block, it would be great to also get suggestions for the attributes .

radeksimko:
@pecigonzalo Understood - that's another valid case I forgot to mention above. 👍 

[...]

PR #485, opened 2021-05-05
Add support for traversals/references

radeksimko:
Builds on https://github.com/hashicorp/hcl-lang/pull/36
Builds on https://github.com/hashicorp/terraform-schema/pull/41

--- 

Some known (most likely desirable) features were intentionally left out from the MVP, but are tracked individually and will be eventually implemented:

- variable `type` declarations: https://github.com/hashicorp/terraform-ls/issues/490
- references to variables & locals (via `var.x` or `local.x`): https://github.com/hashicorp/terraform-ls/issues/491
- step-based completion: https://github.com/hashicorp/terraform-ls/issues/492
- showing interpolated expression value on hover: https://github.com/hashicorp/terraform-ls/issues/495
- recognition of individual expressions when nested: https://github.com/hashicorp/terraform-ls/issues/496
- references to `module` outputs & completion of inputs: https://github.com/hashicorp/terraform-ls/issues/93

Generally if something related to reference completion/hover etc. seems missing from that list, or the implementation I would suggest opening a new issue (rather than commenting here on this PR).

--- 

### Screenshots

Screenshot 2021-05-17 at 21 00 34
Screenshot 2021-05-17 at 21 00 46
Screenshot 2021-05-17 at 21 01 37
Screenshot 2021-05-17 at 21 01 56

### DIFF
--- a/internal/lsp/completion.go
+++ b/internal/lsp/completion.go
@@ -48,12 +48,14 @@ func toCompletionItem(candidate lang.Candidate, caps lsp.CompletionClientCapabil
 		kind = lsp.EnumCompletion
 	case lang.MapCandidateKind, lang.ObjectCandidateKind:
 		kind = lsp.StructCompletion
+	case lang.TraversalCandidateKind:
+		kind = lsp.VariableCompletion
 	}
 
 	// TODO: Omit item which uses kind unsupported by the client
 
 	var cmd *lsp.Command
-	if candidate.TriggerSuggest {
+	if candidate.TriggerSuggest && snippetSupport {
 		cmd = &lsp.Command{
 			Command: "editor.action.triggerSuggest",
 			Title:   "Suggest",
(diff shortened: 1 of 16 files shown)


internal/lsp/completion.go
```diff
@@ -48,12 +48,14 @@ func toCompletionItem(candidate lang.Candidate, caps lsp.CompletionClientCapabil
...
 		kind = lsp.StructCompletion
+	case lang.TraversalCandidateKind:
+		kind = lsp.VariableCompletion
 	}
 
 	// TODO: Omit item which uses kind unsupported by the client
 
 	var cmd *lsp.Command
-	if candidate.TriggerSuggest {
+	if candidate.TriggerSuggest && snippetSupport {
```
radeksimko, line 58: This is probably something that should've been done some time ago. We do not include the `=` in the simplified TextEdits and so triggering suggest in that case is kind of pointless.

It is not a big deal for most clients today which probably do support snippets anyway, but it also helps reducing the amount of test data where in test we opt into snippet support only where we need it.

aeschright (approved):
👍

PR #489, opened 2021-05-06
Increase request concurrency & make it configurable

### LINKED ISSUE #426: Raise request concurrency
Request concurrency is required in certain scenarios:
 - request cancellation https://github.com/hashicorp/terraform-ls/issues/314
 - publishing diagnostics directly from the handler (we're currently working around the limitation by publishing in a separate routine https://github.com/hashicorp/terraform-ls/blob/87c4e22c2d1b7f9ca66d2e7bca46019db7027e49/internal/langserver/diagnostics/diagnostics.go#L44-L79 )
 - reporting progress https://github.com/hashicorp/terraform-ls/issues/5 https://github.com/hashicorp/terraform-ls/issues/150 directly from handlers

We can attempt to reverse https://github.com/hashicorp/terraform-ls/pull/120 - i.e. increase concurrency.

Before doing that though it would be very useful to add some telemetry to be able to tell whether requests are still getting received and responded to in the right order. We can leverage the native [`telemetry/event`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#telemetry_event) method of LSP, and/or just log this on the VS Code extension side only.

Related: https://github.com/creachadair/jrpc2/issues/43

### LINKED ISSUE #504: Expose RPC request concurrency as a flag
### Current Version

```
0.16.1
```

### Use-cases

Generally the language server should work well within the resource constraints of the machine that it is running on. i.e. it shouldn't use excessive amount of CPU.

The most common way of achieving this in Go is to cap number of concurrent operations to the number of available CPUs, i.e. [`runtime.NumCPU()`](https://golang.org/pkg/runtime/#NumCPU).

This should provide balance (between speed and CPU usage) for most users, but in some cases the user may wish to trade speed for CPU or vice versa - for example when they're actually more constrained and multiple other processes are competing for the same CPUs.

### Proposal

Introduce `-req-concurrency` flag to `terraform-ls serve` which will control the concurrency of the underlying jrpc2 library. That way users have an escape hatch they can utilize in described (hopefully rare) cases.

radeksimko:
Fixes #426
Fixes #504

### DIFF
--- a/internal/cmd/serve_command.go
+++ b/internal/cmd/serve_command.go
@@ -26,13 +26,14 @@ type ServeCommand struct {
 	Version string
 
 	// flags
-	port          int
-	logFilePath   string
-	tfExecPath    string
-	tfExecLogPath string
-	tfExecTimeout string
-	cpuProfile    string
-	memProfile    string
+	port           int
+	logFilePath    string
+	tfExecPath     string
+	tfExecLogPath  string
+	tfExecTimeout  string
+	cpuProfile     string
+	memProfile     string
+	reqConcurrency int
 }
 
 func (c *ServeCommand) flags() *flag.FlagSet {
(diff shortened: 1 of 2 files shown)


aeschright (approved):
👍

PR #501, opened 2021-05-10
support templated paths for cpuprofile & memprofile

### LINKED ISSUE #224: Support templated paths for cpuprofile & memprofile
### Current Version

```
0.4.1
```

### Use-cases

https://github.com/hashicorp/terraform-ls/pull/223 introduces new troubleshooting flags, which - unlike the other two (`log-file` & `tf-log-file`) don't support templating. As the profiling should be something users need to enable only rarely the value of templating is much lower, but it would still be useful to be able to see things in context and correlate log files with profiles e.g. by PIDs.

### Attempted Solutions

Run only one instance of the server, or try to match log manually (error prone).

### Proposal

Decouple template processing into its own package and support the same variables/functions as we support for the logging flags.

radeksimko:
Closes #224

### DIFF
--- a/docs/TROUBLESHOOTING.md
+++ b/docs/TROUBLESHOOTING.md
@@ -40,7 +40,7 @@ templated paths (as described below) may produce many log files over time.
 
 ### Log Path Templating
 
-Log paths support template syntax. This allows sane separation of logs while accounting for:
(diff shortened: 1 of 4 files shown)


aeschright (approved):
👍

PR #508, opened 2021-05-12
Add new 'module.callers' command

### LINKED ISSUE #507: Introduce "module.callers" custom command
### Current Version

```
0.16.1
```

### Background

Currently the VS Code extension uses custom command called `rootmodules` which has historically been used to answer a question of "Where could a module in question be `init`-ed `validate`-d etc. from?", i.e. What is its root module?

https://github.com/hashicorp/terraform-ls/blob/c4ea46f33c11bd47b0dbadfc413bb2f61c98b673/internal/langserver/handlers/command/modules.go#L20-L29

Internally to answer that question the server would basically inspect module manifests of all _initialized_ modules it found in the hierarchy of the whole workspace.

To maintain compatibility, the `rootmodules` command currently still returns only _initialized_ modules, or more precisely modules for which it was able to obtain schema via `terraform providers schema -json` successfully.

A module can have just initialized submodules (e.g. via `terraform get`) and still be a valid candidate despite our inability to obtain schema (e.g. where plugins were not downloaded or state backend requires auth). Therefore we can be missing some useful suggestions.

### Use-cases

Provide clients with a list of paths to module callers, which they can use as a hint in the UI for where to run commands, such as `init`, `validate` etc.

#### New Command: `module-callers`

Introduce new command `module-callers` accepting a single _folder_ URI, e.g. `file:///path/to/module` with the following response structure:

```go
type moduleCallersResponse struct {
	FormatVersion int          `json:"v"`
	Callers         []moduleCaller `json:"callers"`
}

type moduleCaller struct {
	URI  string `json:"uri"`
	Name string `json:"name"`
}
```

#### Deprecation of `rootmodules`

It is also proposed to deprecate the existing `rootmodules` command in favour of `module-callers`, although I'm not sure what the best way of deprecating it is. Ideally clients (including the VS Code extension) should ask the server whether it supports certain commands before attempting to execute them, but I'm unsure whether this LSP convention is being followed.

Given that this command was historically used to "just" improve the UI, I'm thinking that returning error on execution should only mean that users are forced to type out module paths more often, but it shouldn't _block_ anyone.

radeksimko:
Closes #507

Deprecation is tracked under https://github.com/hashicorp/terraform-ls/issues/515


aeschright:
I created an issue for the extension here: https://github.com/hashicorp/vscode-terraform/issues/632
Feel free to add or modify anything you need.

aeschright (approved):

Issue #509, opened 2021-05-13
Reduce CPU usage by implementing custom Copy funcs for schema


radeksimko:
### Current Version

```
0.16.2
```

### Background

A schema cache was introduced as part of `0.16.0` release. memdb was also introduced to make operations above the state easier to handle. One side-effect of these changes was that we would create a full copy of every provider schema before storing that into memdb. This should have negligible impact on memory usage (as GC should cleanup the original copy), however because we make the copy using `copystructure` which in turn makes heavy use of the `reflect` library, copying these schemas consumes significant amount of CPU.

This can be seen from CPU profiles collected by @mattduguid as part of a debugging session per https://github.com/hashicorp/vscode-terraform/issues/591 , where about 70% of the CPU is spent in [`schemas.PreloadSchemasToStore()`](https://github.com/hashicorp/terraform-ls/blob/65fb2cf4b3a6019880aafeb9db44a51d9a037da3/internal/schemas/preload_schemas.go#L9) and a big part of those 70% is spent in `reflectwalk`.

![Screenshot 2021-05-13 at 12 00 57](https://user-images.githubusercontent.com/287584/118116986-e58aed00-b3e2-11eb-9c52-0168de39cfd9.png)

### Use-cases

Reduce CPU spikes upon initial language server launch.

### Proposal

Investigate whether replacing `copystructure` with explicit `Copy()` implementations on schema structures would help reduce the CPU usage as Go won't have to use the (presumably) costly reflection at runtime.

radeksimko:
FWIW I was able to obtain a very similar profile on my Mac with fresh 0.16.2 - at least percentage-wise. I'm unsure why the total CPU time spent is about half, but the profile clearly points to the same bottleneck.

![Screenshot 2021-05-13 at 12 07 52](https://user-images.githubusercontent.com/287584/118117963-27686300-b3e4-11eb-80f4-d40a12345afa.png)

Issue #506, opened 2021-05-11
Provide opt out from preloaded schemas


radeksimko:
### Current Version

```
0.16.1
```

### Background

Users generally expect completion, hover and most of other functionality to work "out of the box". This is why from X.Y.Z we started pre-baking schemas of official and partner providers (~150). While this makes it convenient for users and makes most of the functionality work almost straight away after initial load, it comes at a cost.

That cost is increased memory usage.

At this point we build in 34 (official) + 115 (partner) providers. This has the following impact on memory usage:

 - total memory (represented as [`MemStats.Sys`](https://pkg.go.dev/runtime#MemStats.Sys)) **before** allocation: 69 MiB
 - total memory (represented as [`MemStats.Sys`](https://pkg.go.dev/runtime#MemStats.Sys)) **after** allocation: 206 MiB

Preloading all mentioned schemas increases total memory usage by approximately 137 MiB. This is measured before and after this block of code: https://github.com/hashicorp/terraform-ls/blob/c4ea46f33c11bd47b0dbadfc413bb2f61c98b673/internal/schemas/preload_schemas.go#L10-L13

So it seems to be just a result of unmarshaling the (currently ~26MB on disk) JSON file.

For many users this may be ok cost to pay for the convenience, for some it may not.

### Use Case

User should have a way of opting out from builtin provider schema and reduce the memory footprint of the language server.

### Proposal

Introduce a new flag `-preload-schemas=false` which will prevent builtin schema from ever being unmarshaled or otherwise loaded into memory.

radeksimko:
We have significantly reduced the memory usage as part of https://github.com/hashicorp/terraform-ls/pull/1071 which is included in the latest release (0.29.3). We now only load the schemas into the memory dynamically when needed. Therefore there should be no need to opt out anymore.

PR #513, opened 2021-05-17
Reduce CPU usage via custom Copy methods instead reflection

radeksimko:
Depends on https://github.com/hashicorp/hcl-lang/pull/40
Depends on https://github.com/hashicorp/terraform-schema/pull/44

Closes https://github.com/hashicorp/terraform-ls/issues/509

--- 

The rationale and more context is described in #509 - I can estimate that this will reduce the CPU usage in the hot path (state manipulation) by ~66% (from ~3sec to ~1sec in my configuration) based on this snapshot of a CPU profile collected from a build based on this PR:

Screenshot 2021-05-17 at 20 43 29

The "new" bottleneck is JSON decoding, which is something we could address via https://github.com/hashicorp/terraform-ls/issues/506


aeschright (approved):
👍

PR #514, opened 2021-05-18
Add test for '$/cancelRequest' method

### LINKED ISSUE #314: Fix request cancellation
Request cancellation via LSP's [`$/cancelRequest`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#cancelRequest) don't work since #120 which disabled concurrency.

Cancellation requests therefore now end up queued behind the request which may need to be cancelled.

https://github.com/hashicorp/terraform-ls/commit/0aa7c0ffaa6ecd737373003279762e6b72429940 contains a failing test to reproduce this.

### Server Version

```
0.10.0
```

radeksimko:
Cancellation was fixed via https://github.com/hashicorp/terraform-ls/pull/489

Here I'm just adding a test (which fails when concurrency is set to `1`).

Closes #314

### DIFF
--- a/internal/langserver/handlers/cancel_request_test.go
+++ b/internal/langserver/handlers/cancel_request_test.go
@@ -1,0 +1,77 @@
+package handlers
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"sync"
+	"testing"
+	"time"
+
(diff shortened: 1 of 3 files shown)


aeschright (approved):
👍

PR #525, opened 2021-05-20
cmd/module.callers: Remove rel_path field

radeksimko:
As discussed this field is actually no longer needed by the VS Code extension and so it's better to keep the API minimal.

https://github.com/hashicorp/vscode-terraform/pull/633#issuecomment-843592045


aeschright (approved):
👍

Issue #519, opened 2021-05-19
Support JSON variable files (`*.tfvars.json`)


radeksimko:
### Use-cases

When declaring variables in a file, users would usually be editing the "native" HCL version and https://github.com/hashicorp/terraform-ls/issues/50 describes how we can help them edit these files.

Variable files generally can also be useful for the language server to provide a pre-computed value of an expression on hover, as described in https://github.com/hashicorp/terraform-ls/issues/495.

Users would benefit from #495 if _all_ variable values were reflected there, i.e. not just `*.tfvars`, but also `*.tfvars.json` which would often be generated.

Users would also benefit from an early detection of duplicate variable names across **all auto-loaded variable files**, i.e.

 - `terraform.tfvars`
 - `*.auto.tfvars`
 - `terraform.tfvars.json`
 - `*.auto.tfvars.json`

### Proposal

- Build on the recent PR which added support for `*.tf.json` https://github.com/hashicorp/terraform-ls/pull/672
- Collect variables from `terraform.tfvars.json` and `*.auto.tfvars.json` files in addition to `*.tfvars`

Issue #50, opened 2020-04-15
Support variable files (`*.tfvars`)


radeksimko:
Depends on:
 - https://github.com/hashicorp/terraform-ls/issues/490
 - https://github.com/hashicorp/terraform-ls/issues/491

--- 

### Current Version

```
0.16.1
```

### Use-cases

Many users declare variables in a dedicated file, i.e. `*.tfvars` as per https://www.terraform.io/docs/language/values/variables.html#variable-definitions-tfvars-files

Users would benefit from the same level of support they get for Terraform configurations (`*.tf`) for `*.tfvars` files too.

#### Completion

_Assuming there is a `*.tf` file with the following declarations_

```hcl
variable "image_id" {
  type = string
  description = "Amazon AMI ID"
}

variable "availability_zone_names" {
  type    = list(string)
  default = ["us-west-1a"]
  description = "Names of availability zones"
}
```
then in an empty configuration, completion would bring up names of these declared variables with their type & description, i.e.

 - **image_id** _string_ - Amazon AMI ID
 - **availability_zone_names** - _list(string)_ - Names of availability zones

Then further completion would be available for each variable value, depending on its type, e.g. for a more complex variable type, such as object

```hcl
variable "docker_ports" {
  type = list(object({
    internal = number
    external = number
    protocol = string
  }))
}
```
the internal attributes of the object would be completed
```hcl
docker_ports = [{
  # HERE -> internal, external, protocol
}]
```

#### Hover

```hcl
variable "availability_zone_names" {
  type    = list(string)
  description = "Names of availability zones"
}
variable "docker_ports" {
  type = list(object({
    internal = number
    external = number
    protocol = string
  }))
}
```

If user was to hover over a variable name, e.g. `availability_zone_names` in the above config they would be presented with a popup with details about the field - i.e. its type and description

```
_list(string)_
Names of availability zones
```

#### Semantic Tokens

All valid (known) variable names which are declared via `variable` blocks would be reported as attribute tokens and compatible expressions would be highlighted if present, for example:

```hcl
docker_ports = [{
  internal = 80
}]
```

i.e. `docker_ports` and `internal` would be reported as attribute, and `80` as number.

#### Diagnostics Reporting

Any duplicates or mismatch variable types would be reported as diagnostics, similar to how HCL syntax errors are reported today. For example the following config

```hcl
test = "something"
test = "blah"
```

would result in the following diagnostic being published:

```
[...]

PR #539, opened 2021-05-27
Add LanguageID to file

Nepomuceno:
Adding the language id so we could distinguish between tfvars and terraform files

This will assist the implementation for #50 and #519.

### DIFF
--- a/internal/lsp/file_handler.go
+++ b/internal/lsp/file_handler.go
@@ -67,6 +67,10 @@ func (fh *fileHandler) URI() string {
 	return string(fh.uri)
 }
 
+func (fh *fileHandler) LanguageID() string {
+	return ""
+}
+
 type versionedFileHandler struct {
 	fileHandler
 	v int
--- a/internal/lsp/file_handler.go
+++ b/internal/lsp/file_handler.go
@@ -67,10 +67,6 @@ func (fh *fileHandler) URI() string {
 	return string(fh.uri)
 }
 
-func (fh *fileHandler) LanguageID() string {
-	return ""
-}
-
 type versionedFileHandler struct {
 	fileHandler
 	v int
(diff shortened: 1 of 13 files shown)


radeksimko (commented):
This looks good at first sight - can you also try wiring it up to `didOpen` handler now and maybe write a small integration test which checks that the `languageID` is stored?

See the existing test in https://github.com/hashicorp/terraform-ls/blob/main/internal/langserver/handlers/did_open_test.go for inspiration. You can create your own instance of the `filesystem.Filesystem` via `fileystem.NewFilesystem()`, pass it to the mock server and then check the languageID stored via `GetDocument()` & your new `LanguageID()` method.

Then I'd be happy to merge it - as _storing_ language ID (although we wouldn't do anything with it yet) looks like a good-sized deliverable.

Nepomuceno:
Added some test and hopefully fix the build

internal/lsp/file_handler.go
```diff
@@ -67,6 +67,10 @@ func (fh *fileHandler) URI() string {
 	return string(fh.uri)
 }
 
+func (fh *fileHandler) LanguageID() string {
+	return ""
```
radeksimko, line 71: This makes me realize that metadata such as language ID should be tracked under file/document, not the handler. Handlers' responsibility is typically to just represent the location or pointer to the file/document - which is why e.g. version is also not exposed via handler.

It will take a bit more refactoring through the filesystem package to do this though - I'll do it in a separate commit - I hope you don't mind! 😉

radeksimko: This is what I mean https://github.com/hashicorp/terraform-ls/commit/8fa3718b2ab1a00f98e1d2ce77b2b852d29f8dd6

For some reason I can't push directly to your branch, but feel free to cherry-pick that commit.

Nepomuceno: Not a problem at all

radeksimko (approved):

PR #540, opened 2021-05-28
Add support for `tfvars` (variable files)

### LINKED ISSUE #50: Support variable files (`*.tfvars`)

beandrad:
So that we can use the language server for "tfvars" files.

Closes #50


go.mod
```diff
@@ -31,3 +31,5 @@ require (
 	github.com/zclconf/go-cty-debug v0.0.0-20191215020915-b22d67c1ba0b
 	golang.org/x/tools v0.0.0-20200323144430-8dcfad9e016e
 )
+
+replace github.com/hashicorp/terraform-schema v0.0.0-20210522075401-0f5e258f5e97 => github.com/Nepomuceno/terraform-schema v0.0.0-20210527083206-b43d158a9361
```
beandrad, line 35: Once @Nepomuceno changes are merged in the terraform-schema, I'll update go.mod.

internal/terraform/module/module_manager.go
```diff
@@ -97,16 +99,38 @@ func (mm *moduleManager) EnqueueModuleOp(modPath string, opType op.OpType) error
...
 
-	return schemaForModule(mod, mm.schemaStore)
+	if languageID == "tfvars" {
+		return schemaForTfVars(mod)
+	} else {
+		return schemaForTerraform(mod, mm.schemaStore)
+	}
+}
+
+func schemaForTfVars(mod *state.Module) (*schema.BodySchema, error) {
```
radeksimko, line 115: The implementation is AOK, but I think this could live somewhere here https://github.com/hashicorp/terraform-schema/tree/main/schema (e.g. `variables_schema.go`); exported as `SchemaForVariables(vars map[string]tfmod.Variable) (*schema.BodySchema, error)`

Then the implementation here would be practically just a single line, leaving the important bits (about how the schema is assembled) in `terraform-schema`.

--- 

Side note: I reckon we would benefit from adding `IsSensitive` to the [`AttributeSchema`](https://pkg.go.dev/github.com/hashicorp/hcl-lang@v0.0.0-20210525083843-b62f3a996576/schema#AttributeSchema), which could then be also reflected in completion and hover somehow - e.g. displaying it next to the type and other flags (optional/required/computed)
Screenshot 2021-06-01 at 13 07 46
Screenshot 2021-06-01 at 13 08 06

That would also help us flag up the provider-flagged sensitive attributes here https://github.com/hashicorp/terraform-schema/blob/18f435878cf26e12bff5dcec121b030518997422/schema/convert_json.go#L104

But feel free to leave this particular change in `hcl-lang` for later - I'm just thinking out loud here! 😄

radeksimko (commented):
Thank you for the PR, this looks like a good start! I left you some in-line comments.

I have yet to decide about the language ID to use for these files but keep it `tfvars` for now - I assume it's just a minor implementation detail anyway that can be changed easily.

internal/terraform/module/module_manager.go
beandrad: Thanks for the review! :) PR to add IsSensitive to attribute schema: https://github.com/hashicorp/hcl-lang/pull/42

beandrad: PR to implement `SchemaForVariables()` in terraform-schema: https://github.com/hashicorp/terraform-schema/pull/49 (will add `isSensitive` when the one above is merged).

internal/cmd/completion_command.go
```diff
@@ -118,7 +119,12 @@ func (c *CompletionCommand) Run(args []string) int {
 		return 1
 	}
 
-	schema, err := modMgr.SchemaForModule(fh.Dir())
+	var sc *schema.BodySchema
+	if file.LanguageID() == ilsp.Tfvars.String() {
+		sc, err = modMgr.SchemaForVariables(file.Dir())
+	} else {
+		sc, err = modMgr.SchemaForModule(file.Dir())
+	}
```
beandrad, line 127: I feel this logic should probably be extracted out to avoid repetition in the handlers; not sure where to put it, that's one of the reasons it was initially implemented in the module manager.

internal/langserver/handlers/did_change.go
```diff
@@ -82,7 +84,16 @@ func TextDocumentDidChange(ctx context.Context, params lsp.DidChangeTextDocument
...
 	}
-	diags.PublishHCLDiags(ctx, mod.Path, mod.Diagnostics, "HCL")
+
+	d := map[string]hcl.Diagnostics{}
+	for _, md := range []interface{}{mod.ConfigDiagnostics, mod.VarsDiagnostics} {
+		err = mergo.Merge(&d, md)
+		if err != nil {
+			return err
+		}
+	}
```
beandrad, line 94: this should probably be a function of the `state.Module`; however, since `module.Module` is an alias to the previous one, I don't seem to be able to call it:

```
func (m *Module) Diagnostics() (map[]hcl.Diagnostics, error) {
          d := map[string]hcl.Diagnostics{}	
          for _, md := range []interface{}{mod.ConfigDiagnostics, mod.VarsDiagnostics} {		
                  err = mergo.Merge(&amp;d, md)		
                  if err != nil {			
                         return d, err		
                  }	
              }
         return d, nil
}
```

internal/state/module.go
```diff
@@ -397,7 +427,14 @@ func (s *ModuleStore) UpdateParsedFiles(path string, pFiles map[string]*hcl.File
 		return err
 	}
 
-	mod.ParsedFiles = pFiles
+	langID, _ := lsp.LanguageIDFromPath(path)
```
radeksimko, line 430: I initially thought we would create a new dedicated operation such as `ParseVariables` in `internal/terraform/module/module_ops.go` and relatedly create a new dedicated method here, e.g.

```go
func (s *ModuleStore) UpdateVariableFiles(path string, pFiles map[string]*hcl.File, pErr error) error
```
at which point the separation would be clearer and only _locally_ based on file extensions inside the operation code - no need to care about language IDs there.

If this is _not_ what you plan to do then I'd probably reconsider the whole idea of storing the files separately, because that makes it more difficult here and in a few other places.

internal/state/module.go
```diff
@@ -465,7 +503,13 @@ func (s *ModuleStore) UpdateDiagnostics(path string, diags map[string]hcl.Diagno
 		return err
 	}
 
-	mod.Diagnostics = diags
+	langID, _ := lsp.LanguageIDFromPath(path)
```
radeksimko, line 506: Similarly here I assumed we would have a dedicated method, e.g.

```go
func (s *ModuleStore) UpdateVariableDiagnostics(path string, diags map[string]hcl.Diagnostics) error
```

which would then negate the need for the conditional.

internal/state/module_test.go
```diff
@@ -349,10 +349,10 @@ provider "blah" {
 		t.Fatal(err)
 	}
 
-	expectedParsedFiles := map[string]*hcl.File{
+	expectedParsedConfigFiles := map[string]*hcl.File{
```
radeksimko, line 352: Nitpick: as mentioned I'd suggest `expectedParsedModuleFiles`, "config" could technically be even tfvars config

internal/lsp/language_id.go
```diff
@@ -0,0 +1,30 @@
...
+	Terraform LanguageID = "terraform"
+	Tfvars    LanguageID = "tfvars"
+	Unknown   LanguageID = "Unknown"
+)
+
+func (l LanguageID) String() string {
+	return string(l)
+}
+
+func LanguageIDFromPath(path string) (LanguageID, error) {
```
radeksimko, line 22: AFAIK per LSP spec, it is client's responsibility to make assumptions about language IDs based on extensions. Server should accept any file/bytes it is given and languageID assigned to it and not make any further assumptions.

I understand that internally the LS does need to parse files other than the ones provided by the client (so that it has full module context), but that logic should be IMO contained locally and not signal any relationship with language ID as that's LSP concept. We can just compare the file extensions in the parser, like we do now for `*.tf`.

internal/terraform/module/module_manager_test.go
```diff
@@ -408,3 +413,135 @@ func testLogger() *log.Logger {
 
 	return log.New(ioutil.Discard, "", 0)
 }
+
+func TestSchemaForVariables(t *testing.T) {
```
radeksimko, line 417: Thanks for the test 👍🏻 

I assume most of it is already tested in https://github.com/hashicorp/terraform-schema/blob/main/schema/variable_schema_test.go though, isn't it?

internal/terraform/module/module_ops.go
```diff
@@ -163,10 +164,10 @@ func ParseConfiguration(fs filesystem.Filesystem, modStore *state.ModuleStore, m
 		}
 
 		name := info.Name()
-		if !strings.HasSuffix(name, ".tf") || IsIgnoredFile(name) {
+		langID, err := lsp.LanguageIDFromPath(name)
```
radeksimko, line 167: I understand the intentions, but as mentioned above - Language ID as a concept should ideally stay only on the top layer of LSP (handlers). I think there is no harm in still using `HasSuffix` and keep filtering by file extensions that way?

internal/terraform/module/module_manager_test.go
beandrad: you're right, I removed all of the tests cases but one so that only the actual `SchemaForVariables()` is tested

internal/cmd/completion_command.go
beandrad: re: completion command, I left it as it originally was since it doesn't have access to `schemaForDocument()` in `service.go` and since you're thinking of deprecating it...

radeksimko (commented):
This is starting to look almost ready for merge 👍🏻 

Aside from my in-line comments - can you also add some E2E tests here?

 - https://github.com/hashicorp/terraform-ls/blob/main/internal/langserver/handlers/complete_test.go
 - https://github.com/hashicorp/terraform-ls/blob/main/internal/langserver/handlers/hover_test.go
 - https://github.com/hashicorp/terraform-ls/blob/main/internal/langserver/handlers/semantic_tokens_test.go
 - https://github.com/hashicorp/terraform-ls/blob/main/internal/langserver/handlers/formatting_test.go

internal/terraform/module/module_ops.go
```diff
@@ -166,14 +166,76 @@ func ParseConfiguration(fs filesystem.Filesystem, modStore *state.ModuleStore, m
...
+		if info.IsDir() {
+			// We only care about files
+			continue
+		}
+
+		name := info.Name()
+		if !(strings.HasSuffix(name, ".auto.tfvars") ||
+			strings.HasSuffix(name, ".auto.tfvars.json") ||
+			name == "terraform.tfvars" ||
+			name == "terraform.tfvars.json") || IsIgnoredFile(name) {
```
radeksimko, line 230:
```suggestion
		if !(strings.HasSuffix(name, ".auto.tfvars") ||
			name == "terraform.tfvars") || IsIgnoredFile(name) {
```

I think the JSON is better to be addressed in a separate PR - I'm not even sure what exact implications this should have from user perspective yet.

Also we'll need to double check that all upstream decoders can deal with JSON files - although the parser always returns `*hcl.File` for both "native" HCL and JSON, they have different structure and AFAICT we cannot use `hclsyntax` package at all for parsed JSON files.

I assume that eventually we'll want to show values + depending on Terraform's own behaviour we may want to report duplicate variable names across auto-loaded files. Either way I'd leave this as explicitly unsupported for now.

internal/decoder/decoder.go
```diff
@@ -25,12 +26,25 @@ func DecoderForModule(ctx context.Context, mod module.Module) (*decoder.Decoder,
...
-	for name, f := range mod.ParsedFiles {
+	err := loadFiles(d, mod.ParsedModuleFiles)
+	if err != nil {
+		return nil, err
+	}
+
+	err = loadFiles(d, mod.ParsedVarsFiles)
+	if err != nil {
+		return nil, err
+	}
```
radeksimko, line 37: I know this may looks like a minor thing, but I'd rather see a separate function `DecoderForVariables` here, in a similar vein as `schemaForDocument`. 

I'm not sure what exact implications does it have (probably nothing visible to the user), but I am fairly sure that the decoder does not need both, so it's better to just not load it IMO. Relatedly the decoder for tfvars doesn't need any of the setup above (reference reader, and UTM stuff).

radeksimko (approved):
Aside from the last two comments this LGTM! Thank you for all your effort to get it to this point @beandrad

I will have a discussion internally about what the right language ID is for tfvars - if not `tfvars` and then we can merge it.

internal/decoder/decoder.go
beandrad: done!

radeksimko (commented):
I did some more thorough testing in VS Code as well and I just noticed two more things that we should address here:

 - symbols are not reported for `tfvars`, I reckon there are some minor changes needed in https://github.com/hashicorp/terraform-ls/blob/adf5e7d22fe483c72c7209dee8ee90868738f123/internal/langserver/handlers/symbols.go#L40-L43
 - document links handler should probably just completely ignore tfvars: https://github.com/hashicorp/terraform-ls/blob/adf5e7d22fe483c72c7209dee8ee90868738f123/internal/langserver/handlers/document_link.go - I guess it's no-op anyway, but better to be explicit about it

Screenshot 2021-06-03 at 13 42 08

PR #543, opened 2021-06-04
Support variable `default`

### LINKED ISSUE #537: Completion/hover/highlighting for `variable` `default`
Will likely depend on https://github.com/hashicorp/terraform-ls/issues/50 or at least on some parts of it.

#### Completion

Given a config

```hcl
variable "test" {
  type = object({
    port = number
    protocol = string
  })
  default = {
    # HERE
  }
}
```
when user triggers completion in the highlighted position, they are presented with two candidates:

 - **port** _number_
 - **protocol** _string_

#### Hover

When user hovers over the expression after `default = `, they will be provided with a relevant popup describing type of the expression. For example

```hcl
variable "test" {
  type = map(string)
  default = {
    alpha = "A"
    beta = "B"
  }
}
```

when hovering over `{ alpha ... }` they would be presented with `map(string)`.

In most cases this would basically be just whatever is in `type`, but if `type = any`, `type = list(any)` and similar, the hover popup should reflect the _real_ type inferred from the value (as long as it is compatible).

#### Semantic Tokens

Given a config

```hcl
variable "test" {
  type = string
  default = "blah"
}
variable "foo" {
  type = list(number)
  default = [ 1, 3, 5 ]
}
variable "bar" {
  type = object({
    foo = string
  })
  default = {
    foo = "test"
  }
}
```
 - `"blah"` will be reported as _string_
 - `1`, `3` and `5` will be reported as _number_
 - `foo` will be reported as _attribute name_
 - `"test"` will be reported as _string_

### Proposal

 - (assuming https://github.com/hashicorp/terraform-ls/issues/50 is implemented as per `Proposal` section)
 - Introduce `SetVariables(map[string]module.Variable)` to `terraform-schema`/[`schema.SchemaMerger`](https://pkg.go.dev/github.com/hashicorp/terraform-schema@v0.0.0-20210522075401-0f5e258f5e97/schema#NewSchemaMerger)
 - Set relevant `schema.LiteralTypeExpr` to `default` attribute of a `variable` block via label-driven [`DependentBody`](https://pkg.go.dev/github.com/hashicorp/hcl-lang@v0.0.0-20210522074354-f7480edf31b5/schema#BlockSchema). Set `IsDepKey: true` for the `variable` block label as part of [`SchemaMerger.SchemaForModule`](https://pkg.go.dev/github.com/hashicorp/terraform-schema@v0.0.0-20210522075401-0f5e258f5e97/schema#SchemaMerger.SchemaForModule)
 - Bump `terraform-schema`

beandrad:
So that completion/hover/highlighting is available for variable
`default`.

Fixes #537

### DIFF
--- a/internal/terraform/module/module_manager.go
+++ b/internal/terraform/module/module_manager.go
@@ -143,6 +143,7 @@ func schemaForModule(mod *state.Module, schemaReader state.SchemaReader) (*schem
 		CoreRequirements:     coreRequirements,
 		ProviderRequirements: mod.Meta.ProviderRequirements,
 		ProviderReferences:   mod.Meta.ProviderReferences,
+		Variables:            mod.Meta.Variables,
 	}
 
 	return sm.SchemaForModule(meta)


go.mod
```diff
@@ -33,3 +33,5 @@ require (
 	github.com/zclconf/go-cty-debug v0.0.0-20191215020915-b22d67c1ba0b
 	golang.org/x/tools v0.0.0-20200323144430-8dcfad9e016e
 )
+
+replace github.com/hashicorp/terraform-schema v0.0.0-20210601142728-f0b01c296a30 => github.com/aztfmod/terraform-schema v0.0.0-20210604102211-dbbfe1505190
```
beandrad, line 37: I'll update once https://github.com/hashicorp/terraform-schema/pull/53 is merged

radeksimko (approved):
LGTM!

PR #546, opened 2021-06-10
Bump `terraform-schema`

### LINKED ISSUE #490: Support for type declarations (for variable `type`)
#### Completion

Given a config

```hcl
variable "test" {
  type = # HERE
}
```
when user triggers completion in the highlighted position, they are presented with all available types, e.g. `string`, `number`, `bool`, `object()` etc.

#### Hover

There doesn't seem to be any reasonable use case for hover - we would duplicate (and reformat) the type declaration, or just display "type declaration".

#### Semantic Tokens

Given a config

```hcl
variable "test" {
  type = string
}
variable "foo" {
  type = list(number)
}
variable "bar" {
  type = object({
    foo = string
  })
}
```
all of `string`, `list(number)`, `object(...)` will be highlighted accordingly by clients which support semantic tokens.

### Proposal

 - Introduce `TypeDeclarationExpr` (implementation of [`ExprConstraint`](https://pkg.go.dev/github.com/hashicorp/hcl-lang@v0.0.0-20210522074354-f7480edf31b5/schema#ExprConstraint)) into `hcl-lang.schema` to help express what `variable`'s `type` attribute is
 - Reflect that new expression type in all relevant `Decoder` functions:
    - `CandidatesAtPos` - so that relevant types are returned
       - [`typeexpr.TypeString`](https://pkg.go.dev/github.com/hashicorp/hcl/v2/ext/typeexpr#TypeString) implementation might be helpful for implementing the completion as it will have all the different combinations of types listed there
    - `HoverAtPos` - not sure what useful hover data to return on hover over the type declarations - perhaps just say that it's "type declaration"?
    - `SemanticTokensInFile` - so that relevant types are reflected as tokens
       - "capsule" types such as `map`, `list`, or `set` can be reported as `lang.TokenTypeCapsule` (new)
       - primitive types such as `string`, `number`, `bool` can be reported as `lang.TokenTypePrimitive` (new)
       - attribute names within `object()` can be reported as `lang.TokenAttrName`
 - Use `TypeDeclaration` in the schema of the variable block within `terraform-schema`, e.g.
    - https://github.com/hashicorp/terraform-schema/blob/6d07fd682f6596ef668e918fa29b605e72fbffb8/internal/schema/0.12/variable_block.go#L38
    - https://github.com/hashicorp/terraform-schema/blob/6d07fd682f6596ef668e918fa29b605e72fbffb8/internal/schema/0.14/variable_block.go#L37
 - Bump `hcl-lang` in `terraform-schema`
 - Bump `terraform-schema` & `hcl-lang` in `terraform-schema`

### Notes

Terraform is currently running an experiment with "optional attributes" for object types. The relevant logic is currently only present in Terraform itself, and not `typeexpr` package, and so for simplicity it's better to just ignore that experiment, i.e. avoid completing `optional()`.

https://www.terraform.io/docs/language/expressions/type-constraints.html#experimental-optional-object-type-attributes

beandrad:
This brings hashicorp/terraform-schema#56 to add support for
`isRequired` in variable schema.

### DIFF
--- a/internal/langserver/handlers/hover_test.go
+++ b/internal/langserver/handlers/hover_test.go
@@ -197,7 +197,7 @@ func TestVarsHover_withValidData(t *testing.T) {
 		"textDocument": {
 			"version": 0,
 			"languageId": "terraform",
-			"text": "variable \"test\" {\n type=string\n}\n",
+			"text": "variable \"test\" {\n type=string\n sensitive=true}\n",
 			"uri": "%s/variables.tf"
 		}
 	}`, tmpDir.URI())})
@@ -228,7 +228,7 @@ func TestVarsHover_withValidData(t *testing.T) {
 			"result": {
 				"contents": {
 					"kind": "plaintext",
-					"value": "test string"
+					"value": "test required, sensitive, string"
 				},
 				"range": {
 					"start": { "line":0, "character":0 },
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,7 @@ require (
 	github.com/hashicorp/terraform-exec v0.13.3
 	github.com/hashicorp/terraform-json v0.11.0
 	github.com/hashicorp/terraform-registry-address v0.0.0-20210412075316-9b2996cce896
-	github.com/hashicorp/terraform-schema v0.0.0-20210610083714-03085295bd36
+	github.com/hashicorp/terraform-schema v0.0.0-20210610085118-e42d6aa8c63d
 	github.com/kylelemons/godebug v1.1.0 // indirect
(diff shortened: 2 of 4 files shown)


internal/langserver/handlers/hover_test.go
```diff
@@ -197,7 +197,7 @@ func TestVarsHover_withValidData(t *testing.T) {
 		"textDocument": {
 			"version": 0,
 			"languageId": "terraform",
-			"text": "variable \"test\" {\n type=string\n}\n",
+			"text": "variable \"test\" {\n type=string\n sensitive=true}\n",
```
radeksimko, line 200: 🙏🏻  Thank you for adding the test

radeksimko (approved):

Issue #495, opened 2021-05-07
Show interpolated expression values on hover


radeksimko:
### Current Version

```
0.17 (future version with traversal expression support)
```

### Use-cases

Terraform supports a number of [expressions](https://www.terraform.io/docs/language/expressions/index.html) (such as function calls, arithmetic operators, references etc.).

While expressions help users achieve and increase readability of their configurations they can make it harder to tell what exact value a complex expression will interpolate into, until `plan` phase.

For example

```hcl
resource "aws_vpc" "example" {
  count      = 3
  cidr_block = "10.0.${count.index}.0/24"
}

locals {
  number_of_subnets = length(aws_vpc.example) * 3
}
```
when hovering over `length(aws_vpc.example) * 3` we could display 

```hcl
length(aws_vpc.example) * 3 = 9
```
in the hover popup and when hovering over `"10.0.${count.index}.0/24"` we could try to enumerate over the first few indexes and display e.g.

```
10.0.0.0/24
10.0.1.0/24
10.0.2.0/24
```

### Attempted Solutions

`terraform plan`

### Proposal

TODO

PR #554, opened 2021-06-16
internal: Parse variable default values

radeksimko:
With this patch we'll be persisting decoded default values as `cty.Value`. This will be useful for https://github.com/hashicorp/terraform-ls/issues/495

This will be also used to help decide a type of a variable in cases where the explicit type is not declared (i.e. variable `type` is missing) or is vague (i.e. `type = any`).

Depends on https://github.com/hashicorp/terraform-schema/pull/59


radeksimko:
^ cc @beandrad

PR #588, opened 2021-07-21
internal: expose tf related options via init opts

### LINKED ISSUE #240: Expose Terraform-related options via native LSP "InitializationOptions"
### Current Version

```
0.5.3
```

### Use-cases

Currently we only expose `rootModulePaths` in "LSP native" configuration:

https://github.com/hashicorp/terraform-ls/blob/a1ff20f9f5378eaf2b224f84c8bc05b2028da18f/internal/settings/settings.go#L11-L13

Some runtime settings which we don't strictly need to know before the server is actually initialized is still exposed via CLI flags only, e.g. all Terraform-related options:

https://github.com/hashicorp/terraform-ls/blob/a1ff20f9f5378eaf2b224f84c8bc05b2028da18f/commands/serve_command.go#L43-L47 

We execute Terraform _after_ the initialization, so these settings would be better suited for the LSP native configuration, which in the long run could also be modified at runtime via [`workspace/didChangeConfiguration`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workspace_didChangeConfiguration).

### Attempted Solutions

Keep using flags

### Proposal

We can expose the configuration as `terraformExecPath`, `terraformExecTimeout` and `terraformLogPath`.

https://github.com/hashicorp/terraform-ls/blob/a1ff20f9f5378eaf2b224f84c8bc05b2028da18f/internal/settings/settings.go#L15-L18

I wouldn't go as far as removing the existing flags in favour of LSP settings _yet_, but I _would_ document that as the preferred way of configuring these options and I would only allow one of these ways of configuring any option at a time. i.e. I would error out the initialization if the same settings is provided via flag _and_ LSP.

danishprakash:
Closes #240


radeksimko (changes requested):
Looks like the tests are failing - can you take a look at the results and make these pass again?

You can also run tests locally via `go test ./...`.

Thanks.

danishprakash:
Sure. It looks like new contributors can't rerun tests on subsequent pushes, can you help out here? Thanks

internal/settings/settings.go
```diff
@@ -29,6 +31,23 @@ func (o *Options) Validate() error {
 	if len(o.ModulePaths) != 0 && len(o.ExcludeModulePaths) != 0 {
 		return fmt.Errorf("at most one of `rootModulePaths` and `excludeModulePaths` could be set")
 	}
+
+	// TODO: abstract out this validation logic
+	// since it's being used in both the places
```
danishprakash, line 36: @radeksimko does it make sense to abstract out this validation for paths to a separate package? something like `fileutil` or maybe rename the existing `pathcmp` package to path and have such utilities moved there?

danishprakash (commented):
@radeksimko does it make sense to abstract out this validation for paths to a separate package? something like `fileutil` or maybe rename the existing `pathcmp` package to `path` and have such utilities moved there?

internal/settings/settings.go
```diff
@@ -20,7 +22,7 @@ type Options struct {
 	ExperimentalFeatures ExperimentalFeatures `mapstructure:"experimentalFeatures"`
 
 	// TODO: Need to check for conflict with CLI flags
-	// TerraformExecPath string
+	TerraformExecPath string `mapstructure:"tfExecPath"`
```
radeksimko, line 25:
```suggestion
	TerraformExecPath string `mapstructure:"terraformExecPath"`
```

internal/settings/settings.go
```diff
@@ -29,6 +31,23 @@ func (o *Options) Validate() error {
...
+	// TODO: abstract out this validation logic
+	// since it's being used in both the places
+	if o.TerraformExecPath != "" {
+		path := o.TerraformExecPath
+		if !filepath.IsAbs(path) {
+			return fmt.Errorf("Expected absolute path for Terraform binary, got %q", path)
+		}
+		stat, err := os.Stat(path)
+		if err != nil {
+			return fmt.Errorf("** Unable to find Terraform binary: %s", err)
```
radeksimko, line 44:
```suggestion
			return fmt.Errorf("Unable to find Terraform binary: %s", err)
```

internal/settings/settings.go
```diff
@@ -29,6 +31,23 @@ func (o *Options) Validate() error {
...
+		}
+		if stat.IsDir() {
+			return fmt.Errorf("** Expected a Terraform binary, got a directory: %q", path)
```
radeksimko, line 47:
```suggestion
			return fmt.Errorf("Expected a Terraform binary, got a directory: %q", path)
```

internal/settings/settings.go
radeksimko, line 36: I appreciate the comment but I expect that it won't be a big deal as we would eventually deprecate the CLI options so the validation logic won't be needed there anymore and the duplication in short term is fine.

radeksimko (commented):
This looks decent, thank you for the PR. Aside from my inline comments:

Do you mind also adding a note to the CLI option, so that users are aware that we are deprecating it in favour of `terraformExecPath` LSP config option?

I would add a brief note to the flag help itself here https://github.com/hashicorp/terraform-ls/blob/9acdc678d2d494f8f8a4538a64ad0f0c74780c51/internal/cmd/serve_command.go#L45
and then also log a warning (just a message prepended with `[WARN]`) after here https://github.com/hashicorp/terraform-ls/blob/9acdc678d2d494f8f8a4538a64ad0f0c74780c51/internal/cmd/serve_command.go#L142

--- 

Down the line, after this is released and clients had some chance to reflect this we can also start sending a warning message but I wouldn't do that _yet_. I'd just create a separate issue to revisit later.

internal/settings/settings.go
radeksimko: You can add a comment above the other (duplicate) validation function to say that it's deprecated, so that if someone is ever tempted to make changes there they are more careful.

internal/langserver/handlers/initialize.go
```diff
@@ -132,6 +133,21 @@ func (lh *logHandler) Initialize(ctx context.Context, params lsp.InitializeParam
...
+		return serverCaps, fmt.Errorf("TerraformExecPath can either be set via cli or LSP options")
+	}
+
+	var opts = &exec.ExecutorOpts{}
+	if len(cfgOpts.TerraformExecPath) > 0 {
+		opts.ExecPath = cfgOpts.TerraformExecPath
+		ctx = exec.WithExecutorOpts(ctx, opts)
+		if err != nil {
+			return serverCaps, err
+		}
```
radeksimko, line 148: It doesn't seem like this error check does anything useful here?

internal/langserver/handlers/service.go
```diff
@@ -94,14 +94,17 @@ func (svc *service) Assigner() (jrpc2.Assigner, error) {
...
 	// The following is set via CLI flags, hence available in the server context
 	execOpts := &exec.ExecutorOpts{}
-	if path, ok := lsctx.TerraformExecPath(svc.srvCtx); ok {
-		execOpts.ExecPath = path
+	tfExecPath, ok := lsctx.TerraformExecPath(svc.srvCtx)
+	if ok {
+		execOpts.ExecPath = tfExecPath
 	} else {
-		tfExecPath, err := svc.tfDiscoFunc()
+		path, err := svc.tfDiscoFunc()
```
radeksimko, line 101: I think that long term we could probably move this fallback discovery logic into `initialize` handler but we can address that in a separate PR.

radeksimko (approved):
LGTM. I will remove that redundant error check as mentioned in-line and make some minor adjustments of the comment phrasing before merging - I hope you don't mind.

radeksimko (changes requested):
Actually on a second look I'm afraid this won't work because all the methods (`formatting`, `didSave`, `executeCommand`) which are supposed to consume the LSP-configured exec path receive `execOpts` in their own context, but the `initialize` handler never writes into it. Instead they will just all always fall back to auto-discovered Terraform exec path - that is not desirable.

radeksimko (commented):
I'm afraid this is going to be much more involved than expected - we need the exec path pretty early in the lifecycle for walker and module manager, which are both currently created _before_ the `initialize` handler is ever executed.

I will take this over.

radeksimko (approved):

radeksimko:
Thanks for the PR 👍🏻  I wish I realized the complexity earlier, I did some refactoring to address the mentioned problem and also added some docs, so this is now ready to be merged.

danishprakash:
Thanks for reviewing this PR, will go through the refactoring that you've done before making other changes.

PR #590, opened 2021-07-22
Introduce Defer to module loader

radeksimko:
@Nepomuceno Do you think this would help in https://github.com/hashicorp/terraform-ls/pull/551 ?

I guess the `Enqueue` method could return the operation and make the caller do the actual queueing and potentially setting `Defer` function instead of passing it as argument, but that's mostly a stylistic concern that can be refactored later.

If this helps solving the problem I'd ship it. 🤷🏻‍♂️

### DIFF
--- a/internal/langserver/handlers/did_open.go
+++ b/internal/langserver/handlers/did_open.go
@@ -53,7 +53,7 @@ func (lh *logHandler) TextDocumentDidOpen(ctx context.Context, params lsp.DidOpe
 	modMgr.EnqueueModuleOpWait(mod.Path, op.OpTypeDecodeReferenceOrigins)
 
 	if mod.TerraformVersionState == op.OpStateUnknown {
-		modMgr.EnqueueModuleOp(mod.Path, op.OpTypeGetTerraformVersion)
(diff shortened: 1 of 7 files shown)


radeksimko:
The idea is that you'd modify the following line(s) in the watcher and walker:

```go
w.modMgr.EnqueueModuleOp(mod.Path, op.OpTypeParseModuleManifest, nil)
```
to async enqueue all the other relevant/dependent operations, e.g.

```go
w.modMgr.EnqueueModuleOp(mod.Path, op.OpTypeParseModuleManifest, func(opErr error) {
    w.modMgr.EnqueueModuleOp(mod.Path, op.OpTypeParseModuleConfiguration, nil)
    w.modMgr.EnqueueModuleOp(mod.Path, op.OpTypeParseVariables, nil)
    w.modMgr.EnqueueModuleOp(mod.Path, op.OpTypeLoadModuleMetadata, nil)
    w.modMgr.EnqueueModuleOp(mod.Path, op.OpTypeDecodeReferenceTargets, nil)
    w.modMgr.EnqueueModuleOp(mod.Path, op.OpTypeDecodeReferenceOrigins, nil)
})
```

PR #619, opened 2021-08-15
internal: expose pending tf related options via init opts

### LINKED ISSUE #240: Expose Terraform-related options via native LSP "InitializationOptions"

danishprakash:
Closes #240

### DIFF
--- a/docs/SETTINGS.md
+++ b/docs/SETTINGS.md
@@ -6,15 +6,14 @@ The language server supports the following configuration options:
 
 ## `terraformLogFilePath` (`string`)
 
-Path to a file for Terraform executions to be logged into (TF_LOG_PATH)
+Path to a file for Terraform executions to be logged into (`TF_LOG_PATH`)
 with support for variables (e.g. Timestamp, Pid, Ppid) via Go template
-syntax {{.VarName}}
+syntax `{{.VarName}}`
 
 ## `terraformExecTimeout` (`string`)
 
-Overrides Terraform execution timeout (e.g. 30s)
-
-Path to the Terraform binary.
+Overrides Terraform execution timeout in [`time.ParseDuration`](https://pkg.go.dev/time#ParseDuration)
+compatible format (e.g. `30s`)
 
 ## `terraformExecPath` (`string`)
 
--- a/docs/SETTINGS.md
+++ b/docs/SETTINGS.md
@@ -4,6 +4,18 @@
 
 The language server supports the following configuration options:
 
+## `terraformLogFilePath` (`string`)
+
+Path to a file for Terraform executions to be logged into (TF_LOG_PATH)
+with support for variables (e.g. Timestamp, Pid, Ppid) via Go template
+syntax {{.VarName}}
+
+## `terraformExecTimeout` (`string`)
+
+Overrides Terraform execution timeout (e.g. 30s)
+
+Path to the Terraform binary.
+
 ## `terraformExecPath` (`string`)
 
 Path to the Terraform binary.
--- a/internal/langserver/handlers/service.go
+++ b/internal/langserver/handlers/service.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"time"
 
 	"github.com/creachadair/jrpc2"
 	"github.com/creachadair/jrpc2/code"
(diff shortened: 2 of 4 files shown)


internal/langserver/handlers/service.go
```diff
@@ -391,11 +392,30 @@ func (svc *service) configureSessionDependencies(cfgOpts *settings.Options) erro
...
+
+	timeout, ok := lsctx.TerraformExecTimeout(svc.srvCtx)
+	if ok {
+		if len(cfgOpts.TerraformExecTimeout) > 0 {
+			return fmt.Errorf("Terraform exec timeout can either be set via (-tf-exec-timeout) CLI flag " +
+				"or (terraformExecTimeout) LSP config option, not both")
+		}
 		execOpts.Timeout = timeout
+	} else if len(cfgOpts.TerraformExecTimeout) > 0 {
+		d, err := time.ParseDuration(cfgOpts.TerraformExecTimeout)
```
danishprakash, line 414: @radeksimko we can unmarshal into time.Duration but that requires writing a mapstructure hook since it's not supported out of the box, I'm trying this for now, lmk how it looks.

internal/settings/settings.go
```diff
@@ -21,10 +21,9 @@ type Options struct {
 	// ExperimentalFeatures encapsulates experimental features users can opt into.
 	ExperimentalFeatures ExperimentalFeatures `mapstructure:"experimentalFeatures"`
 
-	// TODO: Need to check for conflict with CLI flags
-	TerraformExecPath string `mapstructure:"terraformExecPath"`
-	// TerraformExecTimeout time.Duration
-	// TerraformLogFilePath string
+	TerraformExecPath        string `mapstructure:"terraformExecPath"`
+	TerraformExecTimeout     string `mapstructure:"terraformExecTimeout"`
+	TerraformExecLogFilePath string `mapstructure:"terraformExecLogFilePath"`
```
danishprakash, line 26: `TerraformExecLogFilePath` changed var name here to keep it consistent, feel free to update this though.

internal/langserver/handlers/service.go
radeksimko: I think this may actually be better as we have more control over the (user-facing) error message here.

internal/langserver/handlers/service.go
```diff
@@ -391,11 +392,30 @@ func (svc *service) configureSessionDependencies(cfgOpts *settings.Options) erro
...
+		if err != nil {
+			return fmt.Errorf("Failed to parse Terraform timeout: %s", err)
```
radeksimko, line 416:
```suggestion
			return fmt.Errorf("Failed to parse terraformExecTimeout LSP config option: %s", err)
```

radeksimko (changes requested):
Thanks for this 2nd PR as well.

Aside from my two inline comments it looks good.

We'll also need to document these new options in https://github.com/hashicorp/terraform-ls/blob/main/docs/SETTINGS.md but I'm happy to do that prior to merging.

radeksimko:
Actually one more thing:
Can you also prepend the relevant two CLI flags with `(DEPRECATED)`?
https://github.com/hashicorp/terraform-ls/blob/00594752ad33a748c1e8110d15a7b429bd77dd1e/internal/cmd/serve_command.go#L46-L49

Thanks.

internal/settings/settings.go
danishprakash: Oh, knowing [`terraform-exec`](https://github.com/hashicorp/terraform-exec) is a thing now, it clears a lot of things up. I was under the impression that they all are being passed on to the terraform binary one way or the other. Thanks for clearing this up.

docs/SETTINGS.md
```diff
@@ -4,6 +4,18 @@
 
 The language server supports the following configuration options:
 
+## `terraformLogFilePath` (`string`)
+
+Path to a file for Terraform executions to be logged into (TF_LOG_PATH)
```
danishprakash, line 9: @radeksimko I've used the same desc from the client side changes, If you follow a particular approach for user facing strings, kindly suggest the same. Thanks.

radeksimko (approved):
LGTM, I just made some minor changes to the docs and log message.

Thanks for the PR.

PR #621, opened 2021-08-17
Support standalone (not autoloaded) `*.tfvars` files

### LINKED ISSUE #562: Support standalone (not auto-loaded) tfvars
### Use-cases

The server currently supports only auto-loaded variable files

https://github.com/hashicorp/terraform-ls/blob/f027eab9047efd9324bb7e554b097ea8c111aff3/internal/terraform/module/module_ops.go#L227-L228

but users also often use "stand-alone" variable files which are then passed via CLI to `apply`, `plan` or `destroy`.

They would benefit from similar level of support as we provide for auto-loaded files.

### Proposal

 - Create new `DecodeVariablesFile` function (basically decouple `ParseVariables` operation) which can decode any given _bytes_
 - Modify `didOpen` to parse any such stand-alone variable files and publish any diagnostics
 - Modify `didChange` to re-parse any such stand-alone variable files and publish any diagnostics

Diagnostics for these files should _not_ be stored nor published as part of the module, because we cannot make any conclusions about how the file will be passed and e.g. whether duplicate variable name across two files is an issue.

radeksimko:
Closes #562

This PR decouples some logic of the overgrown `module` package into `ast` and `parser`:
 - `parser` is responsible for parsing `*.tf` and `*.tfvars` files - and potentially other future files we will support
 - `ast` is responsible for representing the files and diagnostics so that they can be filtered accordingly (e.g. we can distinguish between autoloaded and not-autoloaded tfvars files)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 *.go text eol=lf
+*.tf text eol=lf
--- a/internal/decoder/decoder.go
+++ b/internal/decoder/decoder.go
@@ -2,12 +2,11 @@ package decoder
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/hashicorp/hcl-lang/decoder"
 	"github.com/hashicorp/hcl-lang/lang"
-	"github.com/hashicorp/hcl/v2"
 	lsctx "github.com/hashicorp/terraform-ls/internal/context"
+	"github.com/hashicorp/terraform-ls/internal/terraform/ast"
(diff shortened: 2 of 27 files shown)


.gitattributes
```diff
@@ -1 +1,2 @@
 *.go text eol=lf
+*.tf text eol=lf
```
radeksimko, line 2: This is to avoid test failures (in CI) on Windows caused by converted line endings in test config files.

jpogran (approved):
After PR review this looks great!

PR #625, opened 2021-08-18
Format on save code action

### LINKED ISSUE #327: Support code-action-on-save for triggering formatting
### Current Version

```
0.10.0
```

### Use-cases

Being able to format the document using the "code-action-on-save" editor feature. That feature is not really part of the spec. More an editor feature that is currently implemented by VSCode, LSP for Sublime Text, and probably others.

It works by user defining a setting like this one in the editor:

```js
  "editor.codeActionsOnSave": {
    "source.fixAll.eslint": true,
    "source.organizeImports": true
  },
```

and the client requesting and applying such code actions from the server before saving.

Introducing a code action like `source.format.terraform` would allow to granularly enable "format-on-save" for that particular server rather than having to globally enable format-on-save (that utilizes formatting provider).

### Attempted Solutions

It's currently possible to enable format-on-save in various editors by using its built-in format-on-save functionality that utilizes the server's `documentFormattingProvider` capabilities. The problem with that approach is that enabling that feature enables it for all servers and the user when running multiple servers, doesn't always want to enable it for all of them.

### Related LSP methods

The feature would utilize [Code Action Request](https://microsoft.github.io/language-server-protocol/specification#textDocument_codeAction). The client makes a code action request with `context.only` set to `source.format.terraform` (for example), receives it back from the server and applies before saving the file. The code action could return a command to run (requires extra round trip) or ideally provide the edits directly.

The server capability `codeActionProvider.codeActionKinds` is also relevant and should be set by the server to include the `source.format.terraform` code action. That makes the code action available for both code-action-on-save and some context menus that editors might expose for the document.

jpogran:
This adds a code action to format a file based on the `editor.codeActionsOnSave` setting. This can either be a global setting or one specific to the terraform language.

Fixes #327

### DIFF
--- a/docs/language-clients.md
+++ b/docs/language-clients.md
@@ -55,6 +55,16 @@ Client is expected to always launch a single instance of the server and check fo
 It is assumed that paths to these folders will be provided as part of `workspaceFolders`
 in the `initialize` request per LSP.
 
+## Code Actions
+
+The server implements a set of opt-in code actions which perform different actions for the user. The code action request is sent from the client to the server to compute commands for a given text document and range. These commands are typically code fixes to either fix problems or to beautify/refactor code.
+
+### Format Document
+
+The server will format a given document according to Terraform formatting conventions.
+
+This action is available as `source.formatAll.terraform-ls` for clients which configure actions globally (such as Sublime Text LSP) and as `source.formatAll` for clients which allow languageID or server specific configuration (such as VS Code).
+
 ## Code Lens
 
 ### Reference Counts (opt-in)
--- a/internal/langserver/handlers/code_action.go
+++ b/internal/langserver/handlers/code_action.go
@@ -1,0 +1,75 @@
+package handlers
+
+import (
+	"context"
+	"fmt"
+
+	lsctx "github.com/hashicorp/terraform-ls/internal/context"
+	"github.com/hashicorp/terraform-ls/internal/langserver/errors"
+	ilsp "github.com/hashicorp/terraform-ls/internal/lsp"
+	lsp "github.com/hashicorp/terraform-ls/internal/protocol"
+	"github.com/hashicorp/terraform-ls/internal/terraform/module"
+)
+
+func (h *logHandler) TextDocumentCodeAction(ctx context.Context, params lsp.CodeActionParams) []lsp.CodeAction {
+	ca, err := h.textDocumentCodeAction(ctx, params)
+	if err != nil {
+		h.logger.Printf("code action failed: %s", err)
+	}
+
+	return ca
+}
+
+func (h *logHandler) textDocumentCodeAction(ctx context.Context, params lsp.CodeActionParams) ([]lsp.CodeAction, error) {
+	var ca []lsp.CodeAction
+
+	wantedCodeActions := ilsp.SupportedCodeActions.Only(params.Context.Only)
+	if len(wantedCodeActions) == 0 {
+		return nil, fmt.Errorf("could not find a supported code action to execute for %s, wanted %v",
+			params.TextDocument.URI, params.Context.Only)
+	}
+
+	fh := ilsp.FileHandlerFromDocumentURI(params.TextDocument.URI)
+
+	fs, err := lsctx.DocumentStorage(ctx)
+	if err != nil {
+		return ca, err
+	}
+	file, err := fs.GetDocument(fh)
+	if err != nil {
+		return ca, err
+	}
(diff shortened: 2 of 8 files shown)


internal/langserver/handlers/formatting.go
```diff
@@ -14,12 +14,23 @@ import (
...
 
+	edits, err := formatDocument(ctx, h, params.TextDocument.URI)
+	if err != nil {
+		return edits, err
+	}
+
+	return edits, nil
+}
+
+func formatDocument(ctx context.Context, h *logHandler, uri lsp.DocumentURI) ([]lsp.TextEdit, error) {
```
radeksimko, line 25: As discussed via Slack yesterday I think it would be better to only extract the minimum amount of logic here, for a couple of reasons:

 - the code action handler will likely need access to the original document and possibly tfExec as well for other future actions we may introduce later
 - smaller functions will less responsibilities are usually easier to read, maintain and test
 - we gain some extra safety with compile-time checks

With that in mind I'd propose something like the following

```go
func formatDocument(tfExec exec.TerraformExecutor, doc filesystem.DocumentHandler) (filesystem.DocumentChanges, error)
```
What do you think?

internal/langserver/handlers/code_action.go
```diff
@@ -0,0 +1,78 @@
...
+		},
+	}
+
+	x := lsp.CodeAction{
+		Title:       "Format Document",
+		Kind:        lsp.SourceFixAll,
+		IsPreferred: false,
+		Edit:        we,
+	}
+	return x, nil
```
radeksimko, line 77: Totally nitpick: the above blocks can be basically embedded into each other like this:

```go
	return lsp.CodeAction{
		Title:       "Format Document",
		Kind:        lsp.SourceFixAll,
		IsPreferred: false,
		Edit:        lsp.WorkspaceEdit{
			Changes: map[string][]lsp.TextEdit{
				string(uri): edits,
			},
		},
	}, nil
```

internal/langserver/handlers/code_action.go
```diff
@@ -0,0 +1,78 @@
...
+			}
+			ca = append(ca, c)
+		}
+	}
+
+	return ca, nil
+}
+
+func terraformFmt(ctx context.Context, h *logHandler, uri lsp.DocumentURI) (lsp.CodeAction, error) {
+	var ca lsp.CodeAction
```
radeksimko, line 58:
```suggestion
```
I think this variable isn't needed here?

radeksimko (commented):
I gave it spin and it worked nicely in both VS Code and Sublime Text. 👍🏻 

Aside from my inline comments:

We should also add an RPC-level test for this new handler, similar to the one we have in https://github.com/hashicorp/terraform-ls/blob/a7089c3ac5b63fdc2b0d4aae26b55fa5132d9fa0/internal/langserver/handlers/formatting_test.go#L33
and perhaps one for tfvars https://github.com/hashicorp/terraform-ls/blob/a7089c3ac5b63fdc2b0d4aae26b55fa5132d9fa0/internal/langserver/handlers/formatting_test.go#L194
I'd probably omit the version check as it's already heavily tested in multiple places, including `terraform-exec` itself.

Lastly it would be great to get this new feature documented. The best way to do so is probably adding `Code Actions` section similar to Code Lens we have here https://github.com/hashicorp/terraform-ls/blob/main/docs/language-clients.md#code-lens and also briefly explaining the expected use case for the two kinds (`source.formatAll` and `source.formatAll.terraform-ls`) - one serving the needs of a client which has languageID-based configuration and the other one client which has global configuration (such as Sublime Text).

docs/language-clients.md
```diff
@@ -55,6 +55,16 @@ Client is expected to always launch a single instance of the server and check fo
...
+## Code Actions
+
+The server implements a set of opt-in code actions which perform different actions for the user. The code action request is sent from the client to the server to compute commands for a given text document and range. These commands are typically code fixes to either fix problems or to beautify/refactor code.
+
+### Format Document
+
+The server will format a given document or text range according to Terraform formatting conventions.
+
+This can be enabled for a single language by specifying `source.formatAll.terraform-ls` or for all languages by specifying `source.formatAll`. Some editors like VS Code enable further customization of when code actions are performed.
+
```
jpogran, line 67: Not sure how much detail we want here since configuring this is very client specific

radeksimko (approved):
I left you just some very minor inline suggestions, but the PR looks pretty good in general and ready for merging.

PR #630, opened 2021-08-25
Enable Semantic Token Refresh

### LINKED ISSUE #344: Support workspace/semanticTokens/refresh
### Current Version

```
0.11.0
```

### Use-cases

Semantic token based highlighting is fully dependent on schema. Schema may change at runtime, even after it is initially obtained because, e.g.:

 - version of Terraform changes (or becomes known from being previously unknown)
 - new provider (schema) is added
 - existing provider (schema) is removed
 - version of existing provider (schema) changes (is upgraded or downgraded)

### Proposal

Add change hooks to the `StateStore` which can be added when we're setting up the session and state stores, somewhere here https://github.com/hashicorp/terraform-ls/blob/00594752ad33a748c1e8110d15a7b429bd77dd1e/internal/langserver/handlers/service.go#L406-L410

The hook will need to check (via metadata in `filesystem.DocumentStorage`) whether the module has any open files currently and if so, then send [`workspace/semanticTokens/refresh`](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#semanticTokens_refreshRequest) notification to prompt the client to recompute semantic tokens.

Hooks can be implemented within the `StateStore` using [memdb's `Defer`](https://pkg.go.dev/github.com/hashicorp/go-memdb#Txn.Defer) function on any relevant transactions.

jpogran:
Add change hooks to the Module state store to trigger token refresh after specified parts of the schema are discovered after initialization.

Fixes #344

### DIFF
--- a/internal/langserver/handlers/hooks_module.go
+++ b/internal/langserver/handlers/hooks_module.go
@@ -174,8 +174,6 @@ func refreshCodeLens(ctx context.Context, clientRequester session.ClientCaller) 
 
 func refreshSemanticTokens(ctx context.Context, svrCtx context.Context, logger *log.Logger) state.ModuleChangeHook {
 	return func(_, newMod *state.Module) {
-		logger.Printf("Request semantic token refresh for %s", newMod.Path)
-
 		jrpcsvc := jrpc2.ServerFromContext(ctx)
 		_, err := jrpcsvc.Callback(svrCtx, "workspace/semanticTokens/refresh", nil)
 		if err != nil {
--- a/internal/langserver/handlers/hooks_module.go
+++ b/internal/langserver/handlers/hooks_module.go
@@ -3,7 +3,9 @@ package handlers
 import (
 	"context"
 	"fmt"
+	"log"
 
+	"github.com/creachadair/jrpc2"
 	"github.com/hashicorp/terraform-ls/internal/langserver/diagnostics"
 	"github.com/hashicorp/terraform-ls/internal/langserver/session"
 	"github.com/hashicorp/terraform-ls/internal/state"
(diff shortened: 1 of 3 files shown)


internal/state/hooks.go
```diff
@@ -9,3 +9,9 @@ func (mh ModuleChangeHooks) notifyModuleChange(oldMod, newMod *Module) {
 		h(oldMod, newMod)
 	}
 }
+
+func (mh ModuleChangeHooks) notifyTokenRefresh(oldMod, newMod *Module) {
```
radeksimko, line 13: Is there a reason we can't just reuse the existing hook which notifies of a module change and just filter the changes to avoid refreshing e.g. when schema doesn't change?

jpogran: No reason other than I was stumbling through remembering how all this worked and piecing it together as I went. Will modify.

I am also thinking to review where/when we notify in light of how many times this is firing right now. I worry about cpu usage in larger projects. More testing is needed here.

Also am fighting some build errors, but haven't looked into that yet.

radeksimko: I wouldn't expect the hook firing itself and comparison of data within the hook (which then results to returning early and _not_ doing anything else) to be expensive in the grand scheme of things but I haven't done any extensive testing to confirm that theory admittedly.

Also I expect that the times when we want to fire the other two hooks (diagnostics and telemetry) will overlap a lot with this one anyway, so I'm not sure we'd gain all that much from spreading the logic through the `ModuleStore` methods.

jpogran: Awesome. Kept the hooks in the current places.

internal/langserver/handlers/hooks_module.go
```diff
@@ -169,3 +171,15 @@ func refreshCodeLens(ctx context.Context, clientRequester session.ClientCaller)
 		}
 	}
 }
+
+func refreshSemanticTokens(ctx context.Context, svrCtx context.Context, logger *log.Logger) state.ModuleChangeHook {
+	return func(_, newMod *state.Module) {
+		logger.Printf("Request semantic token refresh for %s", newMod.Path)
```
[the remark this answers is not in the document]
jpogran: With our recent experiences with high file count, and your recent work with addressing that, I'm loathe to add anything that might add more I/O. The main reason to have this is to 'see' the request being triggered client side and the accompanying result server side, but that seems more like 'debug' type logging than something that should be enabled by default. I'll remove it now and we can investigate debug type logging later.

radeksimko:
The mentioned upstream bug was addressed in https://github.com/hashicorp/terraform-ls/pull/787

Merge at will! :shipit:

Issue #599, opened 2021-08-02
Recognize inputs and outputs of uninitialized remote (Terraform Registry) modules


radeksimko:
## Use-cases

Users may expect completion, hover, go to definition etc. to work automatically for a `module` block calling any publicly available module they write in their configuration, straight away, e.g.

```hcl
module "eks" {
  source  = "terraform-aws-modules/eks/aws"
  version = "17.1.0"
}
```

### Attempted Solutions / Workarounds

```
terraform get
```

### Pre-requisites

 - [x] https://github.com/hashicorp/terraform-ls/issues/877
 - [x] https://github.com/hashicorp/terraform-ls/issues/882

### `hashicorp/terraform-ls`

 - [ ] Introduce new job `GetModuleMetadataFromRegistry` (alongside existing one such as `LoadModuleMetadata`) which will talk to the Registry API and cache responses keyed off the source address & version of a Registry module in memdb
 - [ ] Ensure `GetModuleMetadataFromRegistry` is scheduled from `didOpen` and `didChange` and depends on `LoadModuleMetadata` (as that job provides module call metadata)
 - [ ] `state`: Change `ModuleStore.ModuleMeta(modPath string)` to `ModuleStore.InstalledModuleMeta(modPath string)`
 - [ ] `state`: Implement `ModuleStore.DeclaredModuleMeta(modCall tfmod.DeclaredModuleCall)` which pulls out any previously cached Registry module metadata from memdb

```go
func GetModuleMetadataFromRegistry(ctx context.Context, modStore *state.ModuleStore) error {
  // TODO: loop over module calls
  // TODO: check if any source address is Registry address
  // TODO: check if that address was already cached
  // TODO: if not, cache it
}
```

### `hashicorp/terraform-schema`

 - [ ] `schema`: Introduce `DeclaredModuleMeta(modCall tfmod.DeclaredModuleCall) (*tfmod.RemoteModuleMeta, error)`
 - [ ] `schema`: Use `DeclaredModuleMeta()` within [`SchemaMerger`](https://github.com/hashicorp/terraform-schema/blob/1e3acbcfd5317a70f82008962162c1116ab7303c/schema/schema_merge.go#L183-L214)
[...]

Issue #598, opened 2021-08-02
Recognize inputs and outputs of _local_ uninitialized modules


radeksimko:
### Current Version

```
0.20.0
```

### Use-cases

Users may expect completion, hover, go to definition etc. to work automatically for a local `module` block they write in their configuration, straight away, e.g.

```hcl
module "ecs_service" {
  source = "./service-module"
}
```

### Attempted Solutions / Workarounds

```
terraform get
```

### Proposal

As proposed in https://github.com/hashicorp/terraform-ls/pull/551#discussion_r650832388 already we consider reading the `source` and load/track the module in question if the target path exists.

This will likely mean reimplementing a small part of `terraform get` within the LS, but we shouldn't need to actually "install" the modules onto the filesystem in `.terraform`.

PR #632, opened 2021-08-27
Module View Command Handler

jpogran:
Adds a handler to return a list of modules in the current workspace to display client side.

Completes: https://github.com/hashicorp/vscode-terraform/pull/746

### DIFF
--- a/internal/langserver/handlers/command/module_calls.go
+++ b/internal/langserver/handlers/command/module_calls.go
@@ -1,0 +1,139 @@
+package command
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"strings"
+
+	"github.com/creachadair/jrpc2/code"
+	lsctx "github.com/hashicorp/terraform-ls/internal/context"
+	"github.com/hashicorp/terraform-ls/internal/langserver/cmd"
+	"github.com/hashicorp/terraform-ls/internal/terraform/datadir"
+	"github.com/hashicorp/terraform-ls/internal/uri"
+)
+
+const moduleCallsVersion = 0
+
+type moduleCallsResponse struct {
+	FormatVersion int          `json:"v"`
+	ModuleCalls   []moduleCall `json:"module_calls"`
+}
+
+type moduleCall struct {
+	Name             string             `json:"name"`
+	SourceAddr       string             `json:"source_addr"`
+	Version          string             `json:"version,omitempty"`
+	SourceType       datadir.ModuleType `json:"source_type,omitempty"`
+	DocsLink         string             `json:"docs_link,omitempty"`
+	DependentModules []moduleCall       `json:"dependent_modules"`
+}
+
+func ModuleCallsHandler(ctx context.Context, args cmd.CommandArgs) (interface{}, error) {
+	response := moduleCallsResponse{
+		FormatVersion: moduleCallsVersion,
+		ModuleCalls:   make([]moduleCall, 0),
+	}
+
+	modUri, ok := args.GetString("uri")
+	if !ok || modUri == "" {
+		return response, fmt.Errorf("%w: expected module uri argument to be set", code.InvalidParams.Err())
+	}
+
+	if !uri.IsURIValid(modUri) {
+		return response, fmt.Errorf("URI %q is not valid", modUri)
+	}
+
+	modPath, err := uri.PathFromURI(modUri)
+	if err != nil {
+		return response, err
+	}
+
+	mm, err := lsctx.ModuleFinder(ctx)
+	if err != nil {
+		return response, err
+	}
+
+	found, _ := mm.ModuleByPath(modPath)
+	if found == nil {
+		return response, nil
+	}
+
+	if found.ModManifest == nil {
+		return response, nil
+	}
+
+	response.ModuleCalls = parseModuleRecords(found.ModManifest.Records)
+
+	return response, nil
+}
+
+func parseModuleRecords(records []datadir.ModuleRecord) []moduleCall {
+	// sort all records by key so that dependent modules are found
+	// after primary modules
+	sort.SliceStable(records, func(i, j int) bool {
+		return records[i].Key < records[j].Key
+	})
+
+	modules := make(map[string]moduleCall)
+	for _, manifest := range records {
+		if manifest.IsRoot() {
+			// this is the current directory, which is technically a module
+			// skipping as it's not relevant in the activity bar (yet?)
+			continue
+		}
+
+		moduleName := manifest.Key
+		subModuleName := ""
+
+		// determine if this module is nested in another module
+		// in the currecnt workspace by finding a period in the moduleName
+		// is it better to look at SourceAddr and compare?
+		if strings.Contains(manifest.Key, ".") {
+			v := strings.Split(manifest.Key, ".")
+			moduleName = v[0]
+			subModuleName = v[1]
+		}
+
+		// build what we know
+		moduleInfo := moduleCall{
+			Name:             moduleName,
+			SourceAddr:       manifest.SourceAddr,
+			DocsLink:         getModuleDocumentationLink(manifest),
+			Version:          manifest.VersionStr,
+			SourceType:       manifest.GetModuleType(),
+			DependentModules: make([]moduleCall, 0),
+		}
+
+		m, present := modules[moduleName]
+		if present {
+			// this module is located inside another so append
+			moduleInfo.Name = subModuleName
+			m.DependentModules = append(m.DependentModules, moduleInfo)
+			modules[moduleName] = m
+		} else {
+			// this is the first we've seen module
+			modules[moduleName] = moduleInfo
+		}
+	}
+
+	// don't need the map anymore, return a list of modules found
+	list := make([]moduleCall, 0)
+	for _, mo := range modules {
+		list = append(list, mo)
+	}
+
+	sort.SliceStable(list, func(i, j int) bool {
+		return list[i].Name < list[j].Name
+	})
+
+	return list
+}
+
+func getModuleDocumentationLink(record datadir.ModuleRecord) string {
+	if record.GetModuleType() != datadir.TFREGISTRY {
+		return ""
+	}
+
+	return fmt.Sprintf(`https://registry.terraform.io/modules/%s/%s`, record.SourceAddr, record.VersionStr)
+}
--- a/internal/langserver/handlers/command/module_calls_test.go
+++ b/internal/langserver/handlers/command/module_calls_test.go
@@ -1,0 +1,89 @@
+package command
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/hashicorp/terraform-ls/internal/terraform/datadir"
+)
+
+func Test_parseModuleRecords(t *testing.T) {
+	tests := []struct {
+		name    string
+		records []datadir.ModuleRecord
+		want    []moduleCall
+	}{
+		{
+			name: "detects terraform module types",
+			records: []datadir.ModuleRecord{
+				{
+					Key:        "ec2_instances",
+					SourceAddr: "terraform-aws-modules/ec2-instance/aws",
+					VersionStr: "2.12.0",
+					Dir:        ".terraform\\modules\\ec2_instances",
+				},
+				{
+					Key:        "web_server_sg",
+					SourceAddr: "github.com/terraform-aws-modules/terraform-aws-security-group",
+					VersionStr: "",
+					Dir:        ".terraform\\modules\\web_server_sg",
+				},
+				{
+					Key:        "eks",
+					SourceAddr: "terraform-aws-modules/eks/aws",
+					VersionStr: "17.20.0",
+					Dir:        ".terraform\\modules\\eks",
+				},
+				{
+					Key:        "eks.fargate",
+					SourceAddr: "./modules/fargate",
+					VersionStr: "",
+					Dir:        ".terraform\\modules\\eks\\modules\\fargate",
+				},
+			},
+			want: []moduleCall{
+				{
+					Name:             "ec2_instances",
+					SourceAddr:       "terraform-aws-modules/ec2-instance/aws",
+					Version:          "2.12.0",
+					SourceType:       "tfregistry",
+					DocsLink:         "https://registry.terraform.io/modules/terraform-aws-modules/ec2-instance/aws/2.12.0",
+					DependentModules: []moduleCall{},
+				},
+				{
+					Name:       "eks",
+					SourceAddr: "terraform-aws-modules/eks/aws",
+					Version:    "17.20.0",
+					SourceType: "tfregistry",
+					DocsLink:   "https://registry.terraform.io/modules/terraform-aws-modules/eks/aws/17.20.0",
+					DependentModules: []moduleCall{
+						{
+							Name:             "fargate",
+							SourceAddr:       "./modules/fargate",
+							Version:          "",
+							SourceType:       "local",
+							DocsLink:         "",
+							DependentModules: []moduleCall{},
+						},
+					},
+				},
+				{
+					Name:             "web_server_sg",
+					SourceAddr:       "github.com/terraform-aws-modules/terraform-aws-security-group",
+					Version:          "",
+					SourceType:       "github",
+					DocsLink:         "",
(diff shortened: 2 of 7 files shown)


internal/langserver/handlers/command/module_list.go
```diff
@@ -0,0 +1,145 @@
...
+)
+
+const moduleListVersion = 0
+
+type moduleListResponse struct {
+	FormatVersion int               `json:"v"`
+	Modules       []terraformModule `json:"modules"`
+}
+
+type terraformModule struct {
```
radeksimko, line 23:
```suggestion
const moduleCallsVersion = 0

type moduleCallsResponse struct {
	FormatVersion int               `json:"v"`
	ModuleCalls       []moduleCall `json:"module_calls"`
}

type moduleCall struct {
```

I think this better represents what the command returns as per https://www.terraform.io/docs/language/modules/syntax.html#calling-a-child-module

internal/langserver/handlers/command/module_list.go
```diff
@@ -0,0 +1,145 @@
...
+
+	list := parseModuleRecords(found.ModManifest.Records)
+
+	return moduleListResponse{
+		Modules:       list,
+		FormatVersion: moduleListVersion,
+	}, nil
+}
+
+func parseModuleRecords(records []datadir.ModuleRecord) []terraformModule {
```
radeksimko, line 65: I think it would be right to add a unit test to cover this fairly complex function.

The input should be relatively easy to mock as part of a test.

internal/langserver/handlers/command/module_list.go
```diff
@@ -0,0 +1,145 @@
...
+	modules := make(map[string]terraformModule)
+	for _, manifest := range records {
+		if manifest.Dir == "." {
```
radeksimko, line 68:
```suggestion
		if manifest.IsRoot() {
```
I _think_ this looks less magical here and you may not even need the comment below at that point.

internal/langserver/handlers/command/module_list.go
```diff
@@ -0,0 +1,145 @@
...
+		// in the currecnt workspace by finding a period in the moduleName
+		// is it better to look at SourceAddr and compare?
+		if strings.Contains(manifest.Key, ".") {
+			v := strings.Split(manifest.Key, ".")
+			moduleName = v[0]
+			subModuleName = v[1]
+		}
+
+		// github modules are put in `files` directory, but Manifest Record does not
+		// record the `name` in Key it puts it in SourceAddr
```
radeksimko, line 87: I think there's some misunderstanding as to what `Key` or `SourceAddr` are for.

`Key` is just locally-relevant reference name within the config (also known as the first and only label of the `module` block), `SourceAddr` is basically 1-1 copy of the raw `source` field within the `module` block.

For example:

```hcl
module "web_server" {
  source = "github.com/hashicorp/two-tier-infra"
}
```
this would be parsed as
 - `Key: "web_server"`
 - `SourceAddr: "github.com/hashicorp/two-tier-infra"`

Both are important and relevant, but not interchangeable.

internal/langserver/handlers/command/module_list.go
```diff
@@ -0,0 +1,145 @@
...
+		if strings.HasPrefix(manifest.Key, "files") {
+			moduleName = manifest.SourceAddr
+		}
+
+		// build what we know
+		moduleInfo := terraformModule{
+			Name:    moduleName,
+			Path:    strings.TrimSuffix(manifest.SourceAddr, "/"),
```
radeksimko, line 95: I think we're mixing two different things here.

As mentioned, `manifest.SourceAddr` is just the "human-typed" address, e.g. `github.com/hashicorp/example` or `hashicorp/consul/aws` (Registry address), it's not strictly speaking a path, but it's probably useful to display in the UI. I guess we can pass this as a separate field `SourceAddr` here?

`manifest.Path` is where the module was actually installed on the filesystem. That is probably _not_ something you would display in the UI, but it may be useful if we want to give users the option to "open installed module" for any local modules? Maybe we can omit `Path` initially.

internal/langserver/handlers/command/module_list.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	sort.SliceStable(list, func(i, j int) bool {
+		return list[i].Name < list[j].Name
+	})
+
+	return list
+}
+
+func getModuleType(manifest datadir.ModuleRecord) string {
```
radeksimko, line 129: It may be worth decoupling this into `internal/terraform/datadir/module_types.go` and relatedly turn the known types into constants.

These can be declared on `*ModuleRecord`.

I think that longer-term we might need to extract it out of the repo entirely, similar to https://github.com/hashicorp/terraform-registry-address but for now keeping it in a separate package should suffice. I can consider the extraction as part of https://github.com/hashicorp/terraform-ls/issues/599 and/or https://github.com/hashicorp/terraform-ls/issues/598

internal/langserver/handlers/execute_command.go
```diff
@@ -12,10 +12,11 @@ import (
...
 var handlers = cmd.Handlers{
-	cmd.Name("rootmodules"):        command.ModulesHandler,
-	cmd.Name("module.callers"):     command.ModuleCallersHandler,
-	cmd.Name("terraform.init"):     command.TerraformInitHandler,
-	cmd.Name("terraform.validate"): command.TerraformValidateHandler,
+	cmd.Name("rootmodules"):          command.ModulesHandler,
+	cmd.Name("module.callers"):       command.ModuleCallersHandler,
+	cmd.Name("terraform.init"):       command.TerraformInitHandler,
+	cmd.Name("terraform.validate"):   command.TerraformValidateHandler,
+	cmd.Name("terraform.modulelist"): command.ModuleListHandler,
```
radeksimko, line 19:
```suggestion
	cmd.Name("module.calls"): command.ModuleCallsHandler,
```
We only prefix the actual commands with `terraform`.

internal/langserver/handlers/command/module_list.go
```diff
@@ -0,0 +1,145 @@
...
+	if strings.HasPrefix(manifest.Dir, "modules") {
+		return "local"
+	} else if strings.HasPrefix(manifest.Key, "files") {
+		return "github"
+	} else {
+		return "tfregistry"
```
radeksimko, line 135: I'm not sure we can make assumptions like above based on the `Dir` or `Key`. I think ideally the only input for detection should be just `manifest.SourceAddr`.

We may be able to reuse some of the logic here https://github.com/hashicorp/terraform/blob/332ea1f233891e45bd274e714ab8e4e11c79dc95/internal/addrs/module_source.go#L191-L268 and relevant tests https://github.com/hashicorp/terraform/blob/332ea1f233891e45bd274e714ab8e4e11c79dc95/internal/addrs/module_source_test.go

or alternatively/additionally the logic in go-getter detectors, which core uses [internally](https://github.com/hashicorp/terraform/blob/332ea1f233891e45bd274e714ab8e4e11c79dc95/internal/getmodules/getter.go#L31-L50):
 - https://github.com/hashicorp/go-getter/blob/main/detect_github.go
 - https://github.com/hashicorp/go-getter/blob/main/detect_git.go
 - https://github.com/hashicorp/go-getter/blob/main/detect_bitbucket.go
 - https://github.com/hashicorp/go-getter/blob/main/detect_gcs.go
 - https://github.com/hashicorp/go-getter/blob/main/detect_s3.go

radeksimko (commented):
This is a good start and works for some simple examples.

Aside from my in-line comments/suggestions we'll also need an E2E test for the new command.

Do let me know if you wish to talk through any particular comment, I'm happy to do it more synchronously via Zoom.

internal/langserver/handlers/command/module_list.go
jpogran: Thanks for the clarification, this makes much more sense.

internal/langserver/handlers/command/module_list.go
jpogran: Moved to new file `internal/terraform/datadir/module_types.go` in this iteration

internal/langserver/handlers/command/module_list.go
jpogran: Used `go-getter` and `terraform-registry-address` to parse source addresses.

internal/langserver/handlers/command/module_list.go
jpogran: I renamed `path` to `source` to more accurately reflect the idea I was going for in the VS Code UI. Going to omit `manifest.Path` for now, as a TODO for later when we get some user feedback

internal/langserver/handlers/command/module_list.go
jpogran: Added tests in `module_calls_test.go`

internal/terraform/datadir/module_types.go
```diff
@@ -0,0 +1,36 @@
+package datadir
+
+import (
+	"github.com/hashicorp/go-getter"
```
radeksimko, line 4: I appreciate the convenience aspect of leveraging `go-getter` here directly, but it's worth noting that Terraform unfortunately doesn't use these detectors _directly_ and so its understanding of addresses may differ from go-getter for various reasons, some of which are documented [here](https://github.com/hashicorp/terraform/blob/51b0aee36cc2145f45f5b04051a01eb6eb7be8bf/internal/getmodules/package.go#L7-L36) and [here](https://github.com/hashicorp/terraform/blob/332ea1f233891e45bd274e714ab8e4e11c79dc95/internal/getmodules/getter.go#L13-L28).

All that said it shouldn't differ too much and we only use the detection to improve UX, so the cost of getting it wrong is relatively low and I think it's totally fine to use it for now.

It is likely though that in the long run we will maintain some custom logic which more accurately reflects TF Core implementation and doesn't have any side effects such as making network calls.

TL;DR Ignore the above, the use of go-getter is 👌🏻  in this PR 😄

internal/terraform/datadir/module_types.go
```diff
@@ -0,0 +1,36 @@
...
+func (r *ModuleRecord) GetModuleType() ModuleType {
+	if _, err := tfregistry.ParseRawProviderSourceString(r.SourceAddr); err == nil {
+		return TFREGISTRY
+	}
+	if _, ok, _ := new(getter.GitHubDetector).Detect(r.SourceAddr, ""); ok {
+		return GITHUB
+	}
+	if _, ok, _ := new(getter.GitLabDetector).Detect(r.SourceAddr, ""); ok {
+		return GITLAB
+	}
```
radeksimko, line 27: Terraform actually doesn't treat GitLab any differently - at least it doesn't use the GitLab detector at all:

https://github.com/hashicorp/terraform/blob/332ea1f233891e45bd274e714ab8e4e11c79dc95/internal/getmodules/getter.go#L30-L52

```suggestion
```

internal/terraform/datadir/module_types.go
```diff
@@ -0,0 +1,36 @@
...
+	if _, ok, _ := new(getter.GitDetector).Detect(r.SourceAddr, ""); ok {
+		return GITHUB
```
radeksimko, line 29:
```suggestion
		return GIT
```
Typo? 👀

internal/terraform/datadir/module_types.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+	if _, ok, _ := new(getter.FileDetector).Detect(r.SourceAddr, ""); ok {
```
radeksimko, line 31: Terraform itself doesn't actually use this particular detector - it instead implements its own in https://github.com/hashicorp/terraform/blob/332ea1f233891e45bd274e714ab8e4e11c79dc95/internal/getmodules/file_detector.go#L9-L31 but similar to overall use of go-getter I think the difference is minor enough that we can use it here initially. 👌🏻

internal/langserver/handlers/command/module_calls_test.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+func Test_parseModuleRecords(t *testing.T) {
+	type args struct {
+		records []datadir.ModuleRecord
+	}
+	tests := []struct {
+		name string
+		args args
+		want []moduleCall
```
radeksimko, line 18:
```suggestion
	tests := []struct {
		name string
		records []datadir.ModuleRecord
		want []moduleCall
```

Nitpick, but I think there is no need for that extra type definition?

radeksimko (commented):
Thank you for making these updates, I think it's pretty close to a merge-able state!

I left you some more in-line comments, majority of which are relatively minor/stylistic.

As always - do let me know if you wish to discuss any particular comment/feedback more synchronously via Zoom, I'm happy to do it!

internal/langserver/handlers/command/module_calls_test.go
jpogran: This is from the neat Go extension generated test cases, and works well with multiple parameter functions but not so well with single parameters.

internal/terraform/datadir/module_types.go
jpogran: This was a placeholder until we could decide on icon for git repo based modules. We have landed on `git-branch`, so we can now correctly report the type

internal/terraform/datadir/module_types.go
jpogran: 👍🏼 Like you say, we aren't interested in validity here, we're interested in figuring out whether we use the Terraform icon or a folder icon or a git icon.

internal/terraform/datadir/module_types.go
radeksimko: Yep, agreed. 👍🏻

internal/langserver/handlers/command/module_calls.go
```diff
@@ -0,0 +1,144 @@
...
+			moduleName = v[0]
+			subModuleName = v[1]
+		}
+
+		t := manifest.GetModuleType()
+
+		// build what we know
+		moduleInfo := moduleCall{
+			Name:             moduleName,
+			Source:           strings.TrimSuffix(manifest.SourceAddr, "/"),
```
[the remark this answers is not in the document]
jpogran: When I was making the treeitem description `sourceAdde@version` the trailing slash didn't match. This is not longer needed as we are not displaying that information that way any longer, as well as a display-centric bit of logic should not be in the LS

internal/terraform/datadir/module_types.go
```diff
@@ -0,0 +1,36 @@
...
+const (
+	NOTFOUND   ModuleType = "notfound"
+	TFREGISTRY ModuleType = "tfregistry"
+	LOCAL      ModuleType = "local"
+	GITHUB     ModuleType = "github"
+	GITLAB     ModuleType = "gitlab"
+)
+
+func (r *ModuleRecord) GetModuleType() ModuleType {
+	if _, err := tfregistry.ParseRawProviderSourceString(r.SourceAddr); err == nil {
```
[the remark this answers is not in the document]
jpogran: After discussion, we are going to keep the code as is and ticket future work at building a parser for Terraform module source addresses. Building such a parser is out of scope for this set of work and while technically incorrect `ParseRawProviderSourceString` works well enough to figure out if we are using the Terraform registry.

internal/langserver/handlers/command/module_calls_test.go
```diff
@@ -0,0 +1,90 @@
...
+					SourceType:       "github",
+					DocsLink:         "",
+					DependentModules: []moduleCall{},
+				},
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := parseModuleRecords(tt.records); !reflect.DeepEqual(got, tt.want) {
```
radeksimko, line 84: nitpick: [`cmp.Diff`](https://github.com/google/go-cmp) provides a better (more human-readable) way of comparing complex structures in tests and we already use it in many other tests, but it's certainly not a blocker for merging the PR!

internal/terraform/datadir/module_types.go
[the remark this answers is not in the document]
radeksimko: Thank you for capturing all the useful context in that new comment! 👍🏻

radeksimko (approved):
🚢 🚀

internal/langserver/handlers/command/module_calls_test.go
jpogran: omg that is so much nicer!

PR #652, opened 2021-09-27
schemas: exclude netapp provider from schema generation

radeksimko:
Related to https://github.com/hashicorp/terraform-ls/runs/3719625555?check_suite_focus=true#step:6:64 and https://github.com/NetApp/terraform-provider-netapp-cloudmanager/issues/74


jpogran (approved):

radeksimko:
This is actually no longer needed as maintainers released new versions correctly.

PR #657, opened 2021-09-29
Prefill Required Attributes

jpogran:


internal/langserver/handlers/complete.go
```diff
@@ -47,6 +47,13 @@ func (h *logHandler) TextDocumentComplete(ctx context.Context, params lsp.Comple
 	}
 	d.SetSchema(schema)
 
+	expFeatures, err := lsctx.ExperimentalFeatures(ctx)
+	if err != nil {
+		d.PrefillRequiredAttributes = false
```
radeksimko, line 52: You may know this already, but in Go you can safely assume that any `bool` struct field always defaults to `false`, so this has basically no effect.

Also the condition under which this errors out is pretty rare, but when it does it almost certainly points to a bug which will need to be fixed, so I'd return the error here, just like we do it in the `DidSave` handler:
https://github.com/hashicorp/terraform-ls/blob/8840e7ba573d1e431898779b80a9ba2d464ce74d/internal/langserver/handlers/did_save.go#L14-L17

internal/settings/settings.go
```diff
@@ -9,7 +9,8 @@ import (
 )
 
 type ExperimentalFeatures struct {
-	ValidateOnSave bool `mapstructure:"validateOnSave"`
+	ValidateOnSave            bool `mapstructure:"validateOnSave"`
+	PrefillRequiredAttributes bool `mapstructure:"prefillRequiredAttributes"`
```
radeksimko, line 13: In line with my comments to your other PR, I think that the name here should reflect that we're completing both attributes and blocks.

radeksimko:
Oh, one last thing - I think we should also document this new experiment here:

https://github.com/hashicorp/terraform-ls/blob/main/docs/SETTINGS.md#experimentalfeatures

It doesn't have to be part of this PR, but we should do it before releasing it, so that we can also link to it from the changelog.

PR #680, opened 2021-10-18
Fix Code Action Resolution

### LINKED ISSUE #679: Issue with running the source.formatAll.terraform-ls code action on save
I've only got around to testing the #327 feature now and noticed that the behavior doesn't seem to be as expected in regards to the `kind` sent in the response. If the client is asking for the `source.formatAll.terraform-ls` kind then that is the kind that should also be included in the response, otherwise the client will ignore the response (at least ST implementation but I'm fairly certain that VSCode would do the same).

### Server Version

```
0.23.0
```

### Terraform Version

```
Terraform v1.0.4
on darwin_amd64
```

### Client Version

```
Sublme Text 4
```

```hcl
variable "spin_autoscaling_config" {
  type = map(number)
  default = {
    target_value         = 40,
    scaleout_cooldown = 60,
    scalein_cooldown  = 120,
  }
}
```

#### textDocument/codeAction Request:

```json
{
  "context": {
    "diagnostics": [],
    "only": [
      "source.formatAll.terraform-ls"
    ]
  },
  "range": {
    "end": {
      "character": 0,
      "line": 8
    },
    "start": {
      "character": 0,
      "line": 0
    }
  },
  "textDocument": {
    "uri": "file:///Users/rafal/xxx/app.tf"
  }
}
```

#### textDocument/codeAction Response:

```json
[
  {
    "edit": {
      "changes": {
        "file:///Users/rafal/xxx/app.tf": [
          {
            "newText": "    target_value      = 40,\n",
            "range": {
              "end": {
                "character": 0,
                "line": 4
              },
              "start": {
                "character": 0,
                "line": 3
              }
            }
          }
        ]
      }
    },
    "kind": "source.fixAll",
    "title": "Format Document"
  }
]
```

### Expected Behavior

The response includes the `source.formatAll.terraform-ls` kind.
The document is formatted.

### Actual Behavior

The response includes the `source.fixAll` kind.
The document is not formatted.

### Steps to Reproduce

1. Add `source.formatAll.terraform-ls` to the list of code actions triggered on save (done differently depending on the editor of choice)
2. Trigger save with the hcl above

jpogran:
This fixes several parts of the code action handler workflow.

- Remove `source` as it is a base type and not a specific action we can use.
- Remove `source.fixAll` because the action implies fixing errors, but terraform fmt only formats code style.
- Remove `source.formatAll` to allow fine grained selection of actions.
- Rename `source.formatAll.terraform-ls` to `source.formatAll.terraform` to follow existing Code Action conventions.
- Correctly set the code action value in the returned `lsp.CodeAction` response to the one chosen by the user.
- Exit early so as to not format a document without an explicit request. When additional code actions are added, we can modify this to return a merged list of actions to apply.

To confirm this works as intended, the following should be tested:

Configure your client with the following settings or follow the usage section of code-actions.md:

```json
"editor.formatOnSave": true,
"[terraform]": {
  "editor.defaultFormatter": "hashicorp.terraform",
  "editor.formatOnSave": false,
  "editor.codeActionsOnSave": {
    "source.formatAll.terraform": true
  },
},
```

closes #679


rchl:
I do think that the formatting doesn't quite fit within the `source.fixAll` code action because formatting and fixing are IMO two different concepts.

Also, producing multiple code actions with formatting changes doesn't quite make sense because those have to be run sequentially so if the first one formats the document then the second one will likely fail in mysterious way since it will be applying the changes to already formatted document.

radeksimko:
> Using the versioned workspace edits (new in 3.16 LSP spec) would be preferred since that would guard against applying changes to already changed document.

I don't disagree but I would hope we can address the problem without rearchitecting it completely.

> I do think that the formatting doesn't quite fit within the source.fixAll code action because formatting and fixing are IMO two different concepts.

I don't disagree here either, but I struggle to find any documentation or established convention to support this opinion, as is unfortunately common with many features in LSP.

rchl:
> > Using the versioned workspace edits (new in 3.16 LSP spec) would be preferred since that would guard against applying changes to already changed document.
> 
> I don't disagree but I would hope we can address the problem without rearchitecting it completely.

Sure. It's not a must have but good to have in general wherever text edits are created.

> > I do think that the formatting doesn't quite fit within the source.fixAll code action because formatting and fixing are IMO two different concepts.
> 
> I don't disagree here either, but I struggle to find any documentation or established convention to support this opinion, as is unfortunately common with many features in LSP.

I tend to think of "fix" addressing an existing diagnostics while "format" formatting unrelated to any existing diagnostics.

jpogran:
I tried several approaches to test what would happen if multiple code actions are received, but could not reproduce this case. In each attempt, only one code action was received by the LS from VS Code.

I tried using the all languages `editor.CodeActionsOnSave` entry.:

```json
"editor.codeActionsOnSave": {
  "source.formatAll": true,
  "source.formatAll.terraform-ls": true
},
"[terraform]": {
  "editor.defaultFormatter": "hashicorp.terraform"
},
```

I also tried variations of the language specific `editor.CodeActionsOnSave`  entry:

```json
"[terraform]": {
  "editor.codeActionsOnSave": {
    "source.formatAll": true,
    "source.formatAll.terraform-ls": true
  },
  "editor.defaultFormatter": "hashicorp.terraform"
},
```

Each time, no matter the order, the LS would only receive one code action to process.

PR #689, opened 2021-10-28
refactor: Reflect upstream `hcl-lang` changes

### LINKED ISSUE #658: Provide workspace-wide symbols for variables in `tfvars` files
### Use-cases

When navigating through the whole module (represented as "workspace" in LSP) we currently provide navigation through any symbols representing blocks and/or attributes in `*.tf` files - such as providers, resources and data sources.

Users would benefit from being able to quickly navigate to variables defined in `*.tfvars` and `*.tfvars.json` (see also https://github.com/hashicorp/terraform-ls/issues/519)

### Proposal

The `WorkspaceSymbol` (representing `workspace/symbol` RPC method) will need to be updated, specifically this part

https://github.com/hashicorp/terraform-ls/blob/8840e7ba573d1e431898779b80a9ba2d464ce74d/internal/langserver/handlers/workspace_symbol.go#L31-L42

We can reuse this function https://github.com/hashicorp/terraform-ls/blob/8840e7ba573d1e431898779b80a9ba2d464ce74d/internal/decoder/decoder.go#L43-L55 and call the same `Symbols` method on the decoder and append found symbols to the slice.

radeksimko:
Depends on https://github.com/hashicorp/hcl-lang/pull/92

This is to prepare for module enhancements as described in https://github.com/hashicorp/vscode-terraform/issues/715

Closes #658

One other minor side effect of the refactoring is that we now highlight the `DefRange` (if available) of a definition instead of the whole `Range`. For example, when you click to `var.foobar` it takes you to the same location, but `variable "foobar"` is highlighted, instead of the whole `variable` block.

--- 

In addition to decoder changes described already in the upstream PR there is a few more changes here.

 - This represents first steps towards getting rid of the `context` package which makes the codebase more prone to import cycles and goes against best practices. The remaining necessary context-related logic is broken down into the packages which produce the relevant data.
 - Similarly passing dependencies via context goes against best practices in Go and reduces the chance of catching bugs early via compile-time checks
 - This also represents first steps towards slimming down and getting rid of the `ModuleManager`


jpogran (approved):
After Radek walked me through the changes, and explained the refactoring, this looks like it will be a great change to make. I think this will not only help for modules, but also for the Prefill Required Fields CodeAction I want to implement later.

radeksimko:
The CI failure is unrelated to the PR:

> Error: Failed to query available provider packages

> Could not retrieve the list of available versions for provider
> thalesgroup/ciphertrust: no available releases match the given constraints

PR #692, opened 2021-11-01
schemas: exclude ciphertrust

radeksimko:
This is to address the failing CI build which is caused by the fact that the provider only has a single release, which is a pre-release (beta1) and since we don't provide any version constraints in our generated configs it implicitly excludes prereleases.

https://github.com/hashicorp/terraform-ls/runs/4065743149?check_suite_focus=true#step:6:65

### DIFF
--- a/internal/schemas/gen/gen.go
+++ b/internal/schemas/gen/gen.go
@@ -245,15 +245,16 @@ func listProviders(tier string) ([]provider, error) {
 // registry for registry.terraform.io/jfrog/artifactory: failed to retrieve
 // authentication checksums for provider: 404 Not Found
 var ignore = map[string]bool{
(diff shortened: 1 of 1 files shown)


dbanck (approved):
Can we somehow track changes on this provider to get notified when a non-beta version is released so that we can remove this from the ignore list?

radeksimko:
That is a good question, I guess the best way would be one/any of us watching the repo for releases.

However I always considered pre-building the schemas an imperfect solution and so I think that a few missing providers is not going to make a huge difference in grand scheme of things.

I guess some of the providers on the current ignore list could be removed now too. We just haven't invested the time into investigating whether they download correctly now.

PR #696, opened 2021-11-04
Implement ignorelist for module indexing

### LINKED ISSUE #670: Make ignorelist for module indexing configurable
### Use-cases

When editor launches the LS on behalf of the user, it will attempt to index all initialized modules within the hierarchy.
Some users open directories which are relatively high up in the hierarchy, meaning that there is many directories to walk through, which in turn may consume CPU and filesystem I/O resources.

This is why we also avoid descending into certain directories which will never contain _editable_ Terraform modules.

https://github.com/hashicorp/terraform-ls/blob/ac2194d310d5ccc6362c3d12dd900eb1a7c6634c/internal/terraform/module/walker.go#L22-L29

Users may however have additional directories in their workspace which they would like to ignore.

### Proposal

Make the ignorelist configurable via options, similar to how other aspects of LS are configurable

https://github.com/hashicorp/terraform-ls/blob/ac2194d310d5ccc6362c3d12dd900eb1a7c6634c/internal/settings/settings.go#L16-L28

dbanck:
This PR adds a new setting called `ignoreDirectoryNames` to the language server. It's now possible to send a list of directories on `initialize`, which should be ignored by the walker and never be indexed.

Closes #670

### DIFF
--- a/docs/SETTINGS.md
+++ b/docs/SETTINGS.md
@@ -68,6 +68,18 @@ Or if left empty
 This setting should be deprecated once the language server supports multiple workspaces,
 as this arises in VS code because a server instance is started per VS Code workspace.
 
+## `ignoreDirectoryNames` (`[]string`)
+
+This allows excluding directories from being checked by the walker by passing a static list of directory names.
+
+The following list of directories will always be ignored:
+
+- `.git`
+- `.idea`
+- `.vscode`
+- `terraform.tfstate.d`
+- `.terragrunt-cache`
+
 ## `experimentalFeatures` (object)
 
 This object contains inner settings used to opt into experimental features not yet ready to be on by default.
--- a/docs/SETTINGS.md
+++ b/docs/SETTINGS.md
@@ -70,7 +70,7 @@ as this arises in VS code because a server instance is started per VS Code works
 
 ## `ignoreDirectoryNames` (`[]string`)
 
-This allows excluding directories from being checked by the walker by passing a static list of directory names.
+This allows excluding directories from being indexed upon initialization by passing a list of directory names.
 
 The following list of directories will always be ignored:
 
--- a/internal/langserver/handlers/initialize_test.go
+++ b/internal/langserver/handlers/initialize_test.go
@@ -134,7 +134,7 @@ func TestInitialize_ignoreDirectoryNames(t *testing.T) {
 			PerWorkDir: map[string][]*mock.Call{
 				pluginDir: validTfMockCalls(),
 				emptyDir: {
-					// TODO? can this be empty?
+					// TODO! improve mock and remove entry for `emptyDir` here afterwards
 					{
 						Method:        "GetExecPath",
 						Repeatability: 1,
--- a/internal/langserver/handlers/initialize_test.go
+++ b/internal/langserver/handlers/initialize_test.go
@@ -2,6 +2,7 @@ package handlers
 
 import (
 	"fmt"
+	"path/filepath"
 	"testing"
 
 	"github.com/creachadair/jrpc2/code"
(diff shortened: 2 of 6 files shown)


internal/settings/settings.go
```diff
@@ -46,6 +49,18 @@ func (o *Options) Validate() error {
 		}
 	}
 
+	if len(o.IgnoreDirectoryNames) > 0 {
+		for _, directory := range o.IgnoreDirectoryNames {
+			if directory == datadir.DataDirName {
+				return fmt.Errorf("cannot ignore data directory %q", datadir.DataDirName)
```
radeksimko, line 55:
```suggestion
				return fmt.Errorf("cannot ignore directory %q", datadir.DataDirName)
```

radeksimko: AFAIK we never refer to that directory as "data directory" in any user-facing docs, this is just internal name, so I'd keep it simpler.

internal/settings/settings_test.go
```diff
@@ -40,3 +43,40 @@ func TestDecodeOptions_success(t *testing.T) {
...
+		input  string
+		result string
+	}{
+		{datadir.DataDirName, `cannot ignore data directory ".terraform"`},
+		{filepath.Join("path", "path"), fmt.Sprintf(`expected directory name, got a path: %q`, filepath.Join("path", "path"))},
+	}
+
+	for _, table := range tables {
+		out, err := DecodeOptions(map[string]interface{}{
+			"IgnoreDirectoryNames": []string{table.input},
```
radeksimko, line 58:
```suggestion
			"ignoreDirectoryNames": []string{table.input},
```
We may want to use the field tag names, to exercise the mapping in the test too.

internal/settings/settings_test.go
```diff
@@ -40,3 +43,40 @@ func TestDecodeOptions_success(t *testing.T) {
...
+
+		result := out.Options.Validate()
+		if result.Error() != table.result {
+			t.Fatalf("expected error: %s, got: %s", table.result, result)
+		}
+	}
+}
+func TestValidate_IgnoreDirectoryNames_success(t *testing.T) {
+	out, err := DecodeOptions(map[string]interface{}{
+		"IgnoreDirectoryNames": []string{"directory"},
```
radeksimko, line 72:
```suggestion
		"ignoreDirectoryNames": []string{"directory"},
```
We may want to use the field tag names, to exercise the mapping in the test too.

internal/langserver/handlers/initialize_test.go
```diff
@@ -119,3 +120,42 @@ func TestInitialize_multipleFolders(t *testing.T) {
...
+
+	InitPluginCache(t, pluginDir)
+	InitPluginCache(t, emptyDir)
+
+	ls := langserver.NewLangServerMock(t, NewMockSession(&MockSessionInput{
+		TerraformCalls: &exec.TerraformMockCalls{
+			PerWorkDir: map[string][]*mock.Call{
+				pluginDir: validTfMockCalls(),
+				emptyDir: {
+					// TODO! improve mock and remove entry for `emptyDir` here afterwards
```
radeksimko, line 137: Do you mind creating an issue to track this and link it from this comment here?

docs/SETTINGS.md
```diff
@@ -68,6 +68,18 @@ Or if left empty
 This setting should be deprecated once the language server supports multiple workspaces,
 as this arises in VS code because a server instance is started per VS Code workspace.
 
+## `ignoreDirectoryNames` (`[]string`)
+
+This allows excluding directories from being checked by the walker by passing a static list of directory names.
```
radeksimko, line 73:
```suggestion
This allows excluding directories from being indexed upon initialization by passing a list of directory names.
```
Nitpick - Just a small suggestion to reflect that this is a user-facing doc and users may not know what a walker is.

radeksimko (approved):
This looks pretty good, well done! I only left a few minor comments/suggestions. Feel free to merge once addressed.

You may want to use the squash option when merging, or rebase to squash the commits.

internal/langserver/handlers/initialize_test.go
dbanck: I've created issue #701 for this

PR #697, opened 2021-11-05
Implement support for JSON variable files

### LINKED ISSUE #519: Support JSON variable files (`*.tfvars.json`)

dbanck:
This PR adds support for collecting variables from `*.tfvars.json` and `*.auto.tfvars.json` files in addition to `*.tfvars`.

Closes #519.

--- 

### Symbols

Users can find variables declared in JSON files when seeking symbols across the workspace.

CleanShot 2021-11-08 at 15 18 29@2x

### Problems

Users can find a list of issues for JSON files in the problems tab.

CleanShot 2021-11-09 at 10 38 40@2x

### DIFF
--- a/internal/terraform/ast/variables.go
+++ b/internal/terraform/ast/variables.go
@@ -31,7 +31,10 @@ func (vf VarsFilename) IsJSON() bool {
 
 func (vf VarsFilename) IsAutoloaded() bool {
 	name := string(vf)
-	return strings.HasSuffix(name, ".auto.tfvars") || name == "terraform.tfvars"
+	return strings.HasSuffix(name, ".auto.tfvars") ||
+		strings.HasSuffix(name, ".auto.tfvars.json") ||
+		name == "terraform.tfvars" ||
+		name == "terraform.tfvars.json"
 }
 
 type VarsFiles map[VarsFilename]*hcl.File
--- a/internal/terraform/ast/variables.go
+++ b/internal/terraform/ast/variables.go
@@ -16,11 +16,17 @@ func NewVarsFilename(name string) (VarsFilename, bool) {
 }
 
 func IsVarsFilename(name string) bool {
-	return strings.HasSuffix(name, ".tfvars") && !isIgnoredFile(name)
+	return (strings.HasSuffix(name, ".tfvars") ||
+		strings.HasSuffix(name, ".tfvars.json")) &&
+		!isIgnoredFile(name)
 }
 
 func (vf VarsFilename) String() string {
 	return string(vf)
+}
+
+func (vf VarsFilename) IsJSON() bool {
+	return strings.HasSuffix(string(vf), ".json")
 }
 
 func (vf VarsFilename) IsAutoloaded() bool {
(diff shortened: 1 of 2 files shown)


internal/terraform/ast/variables.go
```diff
@@ -16,13 +16,19 @@ func NewVarsFilename(name string) (VarsFilename, bool) {
...
 	return string(vf)
 }
 
+func (vf VarsFilename) IsJSON() bool {
+	return strings.HasSuffix(string(vf), ".json")
+}
+
 func (vf VarsFilename) IsAutoloaded() bool {
 	name := string(vf)
 	return strings.HasSuffix(name, ".auto.tfvars") || name == "terraform.tfvars"
```
radeksimko, line 34: I think we'll need to reflect the JSON extensions here as well, otherwise we would never present any diagnostics to the user from JSON tfvars.

For example
https://github.com/hashicorp/terraform-ls/blob/fd0fc089835f3d0c5e23c3ca2a80bc9f27ea2121/internal/langserver/handlers/did_open.go#L81-L84

The high-level idea (reason behind this filtering) is that we present "module-wide" diagnostics more prominently whenever any file from the module is opened - as these are the kind of diagnostics that user would face also when running `validate` or `plan` or `apply` without any additional flags.

The user may also have many individual tfvars files, but we don't present diagnostics from those automatically as we cannot tell which ones will they pick when running validate/plan/apply. Such diagnostics are only presented when these individual files are opened.

radeksimko (changes requested):

internal/terraform/ast/variables.go
dbanck: Oh, I missed that. I extended the check and verified that diagnostics are published.

dbanck: CleanShot 2021-11-08 at 15 21 34@2x

radeksimko: 👍🏻  It would be amazing if you could grab a screenshot of this - the diagnostics for `*.tfvars.json` - rendered in the "Problems" pane in VS Code and add it above to symbols as a 2nd use case, but it's not a blocker for merging! 😃

radeksimko (approved):
LGTM

Issue #241, opened 2020-07-22
Follow symlinks when discovering root modules


radeksimko:
### Current Version

```
0.5.3
```

### Use-cases

The discovery of root modules may not work reliably for users which use symlinks in the paths leading to root modules as we do not currently resolve symlinks. Discovered paths get compared with paths in module manifest files, which may or may not be resolved symlinks and therefore may or may not match.

### Attempted Solutions

Always use resolved paths everywhere, or avoid symlinks.

### Proposal

Resolve symlinks where necessary:

https://github.com/hashicorp/terraform-ls/blob/a1ff20f9f5378eaf2b224f84c8bc05b2028da18f/internal/terraform/rootmodule/root_module_manager.go#L84-L88

https://github.com/hashicorp/terraform-ls/blob/0503882c8bc2be9a5de5c2547be5050fa8a6af3f/internal/terraform/rootmodule/root_module_manager.go#L120-L124

https://github.com/hashicorp/terraform-ls/blob/0503882c8bc2be9a5de5c2547be5050fa8a6af3f/internal/terraform/rootmodule/module_manifest.go#L73

and add relevant test cases to verify functionality.

PR #699, opened 2021-11-08
Avoid duplicate state entries (by avoiding symlink evaluation)

radeksimko:
We use module paths in many different places and we this ensure we will compare the same paths consistently and avoid duplicate entries in memdb.

There is unfortunately no easy way to test this as the potential problem would arise as part of a module operation which is executed asynchronously, and/or when publishing diagnostics, which is in itself hard to test currently.

This bug is more visible in https://github.com/hashicorp/terraform-ls/pull/714 and https://github.com/hashicorp/terraform-ls/pull/700 when you open a folder that has symlink somewhere in the path.

--- 

We will likely re-evaluate this decision more holistically (in the context of the whole codebase) in https://github.com/hashicorp/terraform-ls/issues/241

### DIFF
--- a/internal/langserver/handlers/initialize.go
+++ b/internal/langserver/handlers/initialize.go
@@ -266,7 +266,7 @@ func resolvePath(rootDir, rawPath string) (string, error) {
 }
 
 func cleanupPath(path string) (string, error) {
-	absPath, err := filepath.EvalSymlinks(path)
+	absPath, err := filepath.Abs(path)
 	return toLowerVolumePath(absPath), err
 }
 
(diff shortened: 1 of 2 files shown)


radeksimko:
Actually this is not as significant after all because we only evaluate symlinks for paths passed as part of the (hopefully soon to be deprecated) `rootModules` and `excludeRootModules` config options. It does not affect #700

radeksimko:
Actually it is used by nearly handles, I just didn't read the code properly... 🤦🏻

dbanck (approved):

PR #707, opened 2021-11-09
fix: Avoid crash in telemetry on module removal

radeksimko:
I was able to reproduce this by removing a folder from a workspace, but I also noticed that the VS Code client in particular sends `shutdown` to the server instead of `didChangeWorkspaceFolder` with the removed folder, so we can't _really_ track module removals _yet_ until that client-side bug is addressed.

https://github.com/hashicorp/terraform-ls/blob/055cf0226f209e9b5cbfa95a9ec59d0fe1928c41/internal/state/module.go#L252-L255

I also noticed that the notifications for folder change would never work because we weren't passing the module manager down to the handler, so it would always error out 🙈  This was there since the beginning https://github.com/hashicorp/terraform-ls/pull/502

### DIFF
--- a/internal/langserver/handlers/did_change_workspace_folders_test.go
+++ b/internal/langserver/handlers/did_change_workspace_folders_test.go
@@ -1,0 +1,53 @@
+package handlers
+
+import (
+	"fmt"
+	"testing"
+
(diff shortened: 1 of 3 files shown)


dbanck (approved):
LGTM

Issue #458, opened 2021-04-15
Convert legacy provider addresses via Registry API


radeksimko:
### Current Version

```
0.15.0
```

### Background

As part of https://github.com/hashicorp/terraform-ls/pull/454 we start to represent provider addresses as special types (as opposed to opaque strings) since that makes it possible to reliably compare and store these addresses. i.e. to tell that for example `registry.terraform.io/hashicorp/aws` == `hashicorp/aws` and to be able to tell whether the provider comes from the official registry (in which case we can make assumptions about URLs for docs etc.).

### Problem Definition

There are two scenarios where we may run into an address that isn't valid.

#### Migrated providers

Some providers which previously lived under the single global namespace were migrated out to their own namespace.

For example the `terraform-providers/grafana` provider was migrated to `grafana/grafana`.

#### 0.12

Terraform 0.12 users would end up with "legacy" short address format in multiple places

 - `terraform version` output
 - `terraform providers schema -json`
 - `required_providers` block with simple entries such as `aws = "~> 3.0"` as opposed to whole object declaring the `source`

### Attempted Solutions

Attempt to match providers by swapping "legacy" (unknown) namespace for `hashicorp`, or lookup the first provider of the same type name (i.e. the last segment of the address after `/`).

This isn't accurate, but probably accurate enough for the time being, given that we're looking in just ~100-150 preloaded providers + whatever was sourced via CLI. This certainly won't scale though.

### Proposal

Introduce a new module operation which can request details about a provider from the Registry API, such as whether it was migrated and where. See https://github.com/hashicorp/terraform-registry-address#what-to-do-with-a-legacy-address for more.

PR #712, opened 2021-11-12
Introduce list module providers command

dbanck:
This PR adds a new workspace command `module.providers`.

### DIFF
--- a/internal/langserver/handlers/command/module_providers.go
+++ b/internal/langserver/handlers/command/module_providers.go
@@ -1,0 +1,69 @@
+package command
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/creachadair/jrpc2/code"
+	lsctx "github.com/hashicorp/terraform-ls/internal/context"
+	"github.com/hashicorp/terraform-ls/internal/langserver/cmd"
+	"github.com/hashicorp/terraform-ls/internal/uri"
+)
+
+const moduleProvidersVersion = 0
+
+type moduleProvidersResponse struct {
+	FormatVersion        int                            `json:"v"`
+	ProviderRequirements map[string]providerRequirement `json:"provider_requirements"`
+	InstalledProviders   map[string]string              `json:"installed_providers"`
+}
+
+type providerRequirement struct {
+	DisplayName       string `json:"display_name"`
+	VersionConstraint string `json:"version_constraint"`
+}
+
+func ModuleProvidersHandler(ctx context.Context, args cmd.CommandArgs) (interface{}, error) {
+	response := moduleProvidersResponse{
+		FormatVersion:        moduleProvidersVersion,
+		ProviderRequirements: make(map[string]providerRequirement),
+		InstalledProviders:   make(map[string]string),
+	}
+
+	modUri, ok := args.GetString("uri")
+	if !ok || modUri == "" {
+		return response, fmt.Errorf("%w: expected module uri argument to be set", code.InvalidParams.Err())
+	}
+
+	if !uri.IsURIValid(modUri) {
+		return response, fmt.Errorf("URI %q is not valid", modUri)
+	}
+
+	modPath, err := uri.PathFromURI(modUri)
+	if err != nil {
+		return response, err
+	}
+
+	mm, err := lsctx.ModuleFinder(ctx)
+	if err != nil {
+		return response, err
+	}
+
+	found, _ := mm.ModuleByPath(modPath)
+	if found == nil {
+		return response, nil
+	}
+
+	for provider, version := range found.Meta.ProviderRequirements {
+		response.ProviderRequirements[provider.String()] = providerRequirement{
+			DisplayName:       provider.ForDisplay(),
+			VersionConstraint: version.String(),
+		}
+	}
+
+	for provider, version := range found.InstalledProviders {
+		response.InstalledProviders[provider.String()] = version.String()
+	}
+
+	return response, nil
+}
--- a/internal/langserver/handlers/command/module_providers.go
+++ b/internal/langserver/handlers/command/module_providers.go
@@ -8,6 +8,7 @@ import (
 	lsctx "github.com/hashicorp/terraform-ls/internal/context"
 	"github.com/hashicorp/terraform-ls/internal/langserver/cmd"
 	"github.com/hashicorp/terraform-ls/internal/uri"
+	tfaddr "github.com/hashicorp/terraform-registry-address"
 )
 
 const moduleProvidersVersion = 0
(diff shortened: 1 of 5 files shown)


internal/langserver/handlers/command/module_providers.go
```diff
@@ -0,0 +1,69 @@
...
+	if err != nil {
+		return response, err
+	}
+
+	found, _ := mm.ModuleByPath(modPath)
+	if found == nil {
+		return response, nil
+	}
+
+	for provider, version := range found.Meta.ProviderRequirements {
```
radeksimko, line 57: I think we'll need to account for the possibility of the module not being parsed (yet) here, as this request may be called before the background operation collecting the data finishes. I guess that a simple `nil` check on `.Meta` may do the job for now.

In the long run we'll also address the lack of data by triggering update of the data from server to client. I will create a separate issue for that as we need to address the same problem in `module.calls`.

internal/langserver/handlers/command/module_providers.go
```diff
@@ -0,0 +1,69 @@
...
+	if err != nil {
+		return response, err
+	}
+
+	mm, err := lsctx.ModuleFinder(ctx)
+	if err != nil {
+		return response, err
+	}
+
+	found, _ := mm.ModuleByPath(modPath)
```
radeksimko, line 52:
```suggestion
	mod, _ := mm.ModuleByPath(modPath)
```
Nitpick - I think it would be nice if the variable name could communicate _what_ was found.

internal/langserver/handlers/execute_command_module_providers_test.go
```diff
@@ -0,0 +1,168 @@
...
+	}))
+	stop := ls.Start(t)
+	defer stop()
+
+	ls.Call(t, &langserver.CallRequest{
+		Method: "initialize",
+		ReqParams: fmt.Sprintf(`{
+	    "capabilities": {},
+	    "rootUri": %q,
+		"processId": 12345
```
radeksimko, line 38:
```suggestion
		"processId": 12345
```
Nitpick - this looks a bit misaligned, I think there's a mixture of spaces and tabs here.

internal/langserver/handlers/execute_command_module_providers_test.go
```diff
@@ -0,0 +1,168 @@
...
+	return constraints
+}
+
+func testVersion(t *testing.T, v string) *version.Version {
+	ver, err := version.NewVersion(v)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return ver
+}
```
radeksimko, line 168: Nitpick - you could probably replace this with the upstream `version.Must()` which will `panic` on error and the panic would show up in test (should the version be unparseable), maybe not as nicely, but it will be pretty obvious where it comes from. The main benefit is that it's less code to maintain here 🤷🏻‍♂️  😄

radeksimko (changes requested):
Just a few comments in-line, most are minor - except for the `nil` check.

All-in-all good work!

internal/langserver/handlers/command/module_providers.go
dbanck: Meta is a `ModuleMetadata` struct. Can a struct be `nil`? How check for that?

internal/langserver/handlers/execute_command_module_providers_test.go
dbanck: I guess gofmt won't format things here, because it's inside a string?

internal/langserver/handlers/command/module_providers.go
radeksimko: Oops, sorry for the confusion - I missed that fact entirely, I thought it was a pointer.

In that case we could either turn it into a pointer, or - easier/quicker solution in the context of this PR - we can check that `MetaState != op.OpStateUnknown`. The parsing may still fail (in which case `MetaState = OpStateLoaded`), so if we would also need to check for `MetaErr == nil`.

The only downside with introducing a pointer is that we'd need to carefully check all references and make sure these also contain nil checks, so that we don't introduce a crash.

internal/langserver/handlers/execute_command_module_providers_test.go
radeksimko: Yes, that is one of the reasons, but also this is JSON, not Go, so I'm not sure gofmt would be able to handle it anyway. It doesn't need to be formatted the same throughout the codebase, although if you wanted to take care of that (in a separate PR) that would be nice! 😄 I was more referring to the fact that it stands out from the diff here in this particular test.

internal/langserver/handlers/command/module_providers.go
dbanck: Okay, I've added a check for `MetaState`  and `MetaErr`. This seems a lot easier for now than converting `ModuleMetadata` to a pointer.

internal/langserver/handlers/command/module_providers.go
```diff
@@ -0,0 +1,85 @@
...
+
+	for provider, version := range mod.InstalledProviders {
+		response.InstalledProviders[provider.String()] = version.String()
+	}
+
+	return response, nil
+}
+
+func getProviderDocumentationLink(provider tfaddr.Provider) string {
+	if provider.IsLegacy() || provider.IsBuiltIn() {
```
radeksimko, line 80:
```suggestion
	if provider.IsLegacy() || provider.IsBuiltIn() || provider.Hostname != "registry.terraform.io" {
```

We cannot tell where the docs are if the provider isn't hosted by the official Registry, so we should also exclude linking to these. Such providers can either be side-loaded or hosted in a private registry.

btw. we make the same decision in https://github.com/hashicorp/terraform-schema/blob/188460bcdf1add8d8d7d06fe7aa05ff2cb035a0d/schema/convert_json.go#L349-L366 when attaching document links to the schema for providers.

radeksimko (approved):
Just one last comment about the docs URL generation, feel free to merge once that is addressed. 🚀

internal/langserver/handlers/command/module_providers.go
dbanck: Good catch.

Would it make sense to move `providerHasDocs` at some point into the [terraform-registry-address](https://github.com/hashicorp/terraform-registry-address/blob/main/provider.go) package? Since we're starting to do the same check in different other packages?

radeksimko: I'm not sure as I don't have much visibility into the plans around private registry myself, I think the duplication is not necessarily bad in this case. I reckon private registry will eventually have docs too, but I have no idea how that would work in terms of URLs. I reckon some of these assumptions may be challenged as part of https://github.com/hashicorp/vscode-terraform/issues/737 and https://github.com/hashicorp/terraform-ls/issues/458 so I would be hesitant to bake them into a public API of some kind.

Similarly the builtin provider actually does have docs, it's just that we don't have _versioned_ docs for TF Core and we preferred to not make a blind assumption by linking to https://www.terraform.io/docs/language/state/remote-state-data.html

So many things are still "WIP", unfortunately.

PR #718, opened 2021-11-18
ModuleLoader: Run defered operations in the same routine

radeksimko:
Previously we ran `Defer` as part of a separate goroutine because there were occasionally blocking operations scheduled and waited for as part of `Defer`, which could in turn result in a deadlock with the originating operation.

As all tasks are now async by default (outside of tests) and tests always block on *every* operation until it is finished, this presents an opportunity to clean up some tech debt and remove `time.Sleep`.

--- 

The second commit further prevents deadlocks by marking duplicate operations automatically as done.


radeksimko:
Unfortunately this won't be as simple as I thought - some tests still seem to end up in a deadlock 😑 

Filed https://github.com/hashicorp/terraform-ls/issues/719 and closing this for now.

PR #721, opened 2021-11-19
Update docs with new workspace commands

dbanck:
#632 and #712 introduced two new workspace commands. This PR updates the docs for both commands

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -143,4 +143,38 @@ List of module which are called from the current module.
 }
 ```
 
+### `module.providers`
+
+Provides information about the providers of the current module, including requirements and
+installed version.
+
+**Arguments:**
+
+ - `uri` - URI of the directory of the module in question, e.g. `file:///path/to/network`
+
+**Outputs:**
+
+ - `v` - describes version of the format; Will be used in the future to communicate format changes.
+ - `provider_requirements` - map of provider FQN string to requirements object
+   - `display_name` - a user-friendly FQN string, simplified for readability
+   - `version_constraint` - a comma-separated list of version constraints
+   - `docs_link` - a link to the provider documentation; if available
+ - `installed_providers` - map of provider FQN string to installed version string; can be empty if none installed
+
+```json
+{
+  "v": 0,
+  "provider_requirements": {
+    "registry.terraform.io/hashicorp/aws": {
+      "display_name": "hashicorp/aws",
+      "version_constraint": "~> 3.64.0",
+      "docs_link": "https://registry.terraform.io/providers/hashicorp/aws/latest"
+    }
+  },
+  "installed_providers": {
+    "registry.terraform.io/hashicorp/aws": "3.64.2"
+  }
+}
+```
+
 ### `rootmodules` (DEPRECATED, use `module.callers` instead)
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -104,7 +104,15 @@ can be used to hint the user e.g. where to run `init` or `validate` from.
 
 ### `module.calls`
 
-List of module which are called from the current module.
+List of modules called by the module under the given URI.
+
+Empty array may be returned when e.g.
+  - the URI doesn't represent a module
+  - the configuration is invalid
+  - modules are not installed
+  - there are no module calls
+
+The data is sourced from the local cache within `.terraform` and so it may not necessarily represent newly added module calls in the configuration until they're installed via `get` or `init`.
 
 **Arguments:**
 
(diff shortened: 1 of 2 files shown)


docs/commands.md
```diff
@@ -102,4 +102,79 @@ can be used to hint the user e.g. where to run `init` or `validate` from.
 }
 ```
 
+### `module.calls`
+
+List of module which are called from the current module.
```
radeksimko, line 107:
```suggestion
List of modules called by the module under the given URI.

Empty array may be returned when e.g.
  - the URI doesn't represent a module
  - the configuration is invalid
  - modules are not installed
  - there are no module calls

The data is sourced from the local cache within `.terraform` and so it may not necessarily represent newly added module calls in the configuration until they're installed via `get` or `init`.
```

docs/commands.md
```diff
@@ -102,4 +102,79 @@ can be used to hint the user e.g. where to run `init` or `validate` from.
...
+
+**Arguments:**
+
+ - `uri` - URI of the directory of the module in question, e.g. `file:///path/to/network`
+
+**Outputs:**
+
+ - `v` - describes version of the format; Will be used in the future to communicate format changes.
+ - `module_calls` - array of modules which are called from the module in question
+   - `name` - a unique identifier for this particular module
```
radeksimko, line 117:
```suggestion
   - `name` - the reference name of this particular module (i.e. `network` from `module "network" { ...`)
```

docs/commands.md
```diff
@@ -102,4 +102,79 @@ can be used to hint the user e.g. where to run `init` or `validate` from.
...
+   - `source_addr` - the source address given for this module in configuration
```
radeksimko, line 118:
```suggestion
   - `source_addr` - the source address given for this module call (e.g. `terraform-aws-modules/eks/aws`)
```

docs/commands.md
```diff
@@ -102,4 +102,79 @@ can be used to hint the user e.g. where to run `init` or `validate` from.
...
+   - `version` - the exact version of the module
```
radeksimko, line 119:
```suggestion
   - `version` - version constraint of the module call; applicable to modules hosted by the Terraform Registry (e.g. `~> 1.0`
```

docs/commands.md
```diff
@@ -102,4 +102,79 @@ can be used to hint the user e.g. where to run `init` or `validate` from.
...
+   - `source_type` - source of the Terraform module, e.g. github or tfregistry
```
radeksimko, line 120:
```suggestion
   - `source_type` - source of the Terraform module, e.g. `github` or `tfregistry`
```

docs/commands.md
```diff
@@ -102,4 +102,79 @@ can be used to hint the user e.g. where to run `init` or `validate` from.
...
+
+**Arguments:**
+
+ - `uri` - URI of the directory of the module in question, e.g. `file:///path/to/network`
+
+**Outputs:**
+
+ - `v` - describes version of the format; Will be used in the future to communicate format changes.
+ - `provider_requirements` - map of provider FQN string to requirements object
+   - `display_name` - a user-friendly FQN string, simplified for readability
```
radeksimko, line 159:
```suggestion
   - `display_name` - a human-readable name of the provider (e.g. `hashicorp/aws`)
```

docs/commands.md
radeksimko: In retrospect I guess we should've called this `version_constraint`, but that's totally unrelated to this docs PR 😄

docs/commands.md
```diff
@@ -102,4 +102,79 @@ can be used to hint the user e.g. where to run `init` or `validate` from.
...
+   - `version_constraint` - a comma-separated list of version constraints
```
radeksimko, line 160:
```suggestion
   - `version_constraint` - a comma-separated list of version constraints (e.g. `>= 1.0, < 1.2`)
```

docs/commands.md
```diff
@@ -102,4 +102,79 @@ can be used to hint the user e.g. where to run `init` or `validate` from.
...
+   - `docs_link` - a link to the provider documentation; if available
+ - `installed_providers` - map of provider FQN string to installed version string; can be empty if none installed
```
radeksimko, line 162:
```suggestion
 - `installed_providers` - map where _key_ is the provider FQN and _value_ is the installed version of the provider; can be empty if none are installed
```

internal/langserver/handlers/command/module_calls.go
```diff
@@ -87,7 +87,7 @@ func parseModuleRecords(records []datadir.ModuleRecord) []moduleCall {
 		subModuleName := ""
 
 		// determine if this module is nested in another module
-		// in the currecnt workspace by finding a period in the moduleName
+		// in the current workspace by finding a period in the moduleName
```
radeksimko, line 90: Good catch! 👍🏻

radeksimko (commented):
Thank you for adding these docs.

I left some mostly minor suggestions with examples etc. to hopefully make it a little easier to understand for future readers.

dbanck:
Thanks a lot for your helpful suggestions. I've applied them all.

radeksimko (approved):

Issue #618, opened 2021-08-12
Provide `tfvars` references for `variable` blocks


radeksimko:
### Current Version

```
v0.20.1
```

### Current State

Users currently benefit from being able to lookup references via a context menu by right-clicking to a `variable` block, e.g. finding all references (`var.example`) for

```hcl
variable "example" {
}
```

Users can also see the total number of references within a module via code lens above the `variable` block in VS Code or other supported clients.

### Use-cases

Variables can be referenced in variable (`tfvars`) files and users would benefit from seeing variable references in these files as well, both when looking via the context menu and code lens.

Additionally when requesting definition for a `tfvars` entry, the corresponding `variable` block is returned.

#### terraform-schema

 - Modify [`schema.SchemaForVariables()`](https://github.com/hashicorp/terraform-schema/blob/c8db2640719fe08cf64091ffb68c788a3dedeabe/schema/variable_schema.go#L10) such that uses `OriginForTarget` as introduced in https://github.com/hashicorp/hcl-lang/pull/94 and used already for module references in https://github.com/hashicorp/terraform-schema/pull/82

PR #727, opened 2021-11-26
Introduce go-to-variable from `tfvars` files

dbanck:
Closes https://github.com/hashicorp/terraform-ls/issues/618

---

Depends on:
* https://github.com/hashicorp/terraform-schema/pull/84

### DIFF
--- a/internal/decoder/decoder.go
+++ b/internal/decoder/decoder.go
@@ -66,7 +66,7 @@ func varsPathContext(mod *state.Module) (*decoder.PathContext, error) {
 		Files:            make(map[string]*hcl.File, 0),
 	}
 
-	for _, origin := range mod.RefOrigins {
+	for _, origin := range mod.VarsRefOrigins {
 		if ast.IsVarsFilename(origin.OriginRange().Filename) {
 			pathCtx.ReferenceOrigins = append(pathCtx.ReferenceOrigins, origin)
 		}
--- a/internal/decoder/decoder.go
+++ b/internal/decoder/decoder.go
@@ -63,17 +63,12 @@ func varsPathContext(mod *state.Module) (*decoder.PathContext, error) {
 		Schema:           schema,
 		ReferenceOrigins: make(reference.Origins, 0),
 		ReferenceTargets: make(reference.Targets, 0),
-		Files:            make(map[string]*hcl.File, 0),
+		Files:            make(map[string]*hcl.File),
 	}
 
 	for _, origin := range mod.VarsRefOrigins {
 		if ast.IsVarsFilename(origin.OriginRange().Filename) {
 			pathCtx.ReferenceOrigins = append(pathCtx.ReferenceOrigins, origin)
-		}
-	}
-	for _, target := range mod.RefTargets {
-		if target.RangePtr != nil && ast.IsVarsFilename(target.RangePtr.Filename) {
-			pathCtx.ReferenceTargets = append(pathCtx.ReferenceTargets, target)
 		}
 	}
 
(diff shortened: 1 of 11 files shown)


radeksimko (changes requested):
This is a good start and thank you for capturing the GIFs!

internal/terraform/module/module_ops.go
```diff
@@ -299,17 +299,32 @@ func DecodeReferenceOrigins(ctx context.Context, modStore *state.ModuleStore, sc
...
+	})
+	if err != nil {
+		return err
+	}
+
+	// TODO! merge errors?
+	moduleOrigins, _ := moduleDecoder.CollectReferenceOrigins()
+	varOrigins, rErr := varsDecoder.CollectReferenceOrigins()
+
+	origins := append(moduleOrigins, varOrigins...)
```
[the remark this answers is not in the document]
dbanck: Thanks for your comment. I'm also not happy with my current implementation. Let's say this is instead a PoC.

I assumed we have to add the `tfvars` reference origins to the module `RefOrigins` field since there is no other field for storing origins coming from var files. After adding them, the loop in `varsPathContext`  is working, adding them to the `pathCtx`.
https://github.com/hashicorp/terraform-ls/blob/517a7aae497790131d498aa732d011cd682705d3/internal/decoder/decoder.go#L69-L73

Are you suggesting altering the loops in `varsPathContext` so we don't need to modify `RefOrigins`?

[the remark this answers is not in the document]
radeksimko: What I'm trying to say is that `state.Module` should IMO have a dedicated field for tfvars references and this field should be returned as part of the `PathContext` only for tfvars language ID.

Perhaps more universal pattern would be a map where key is the language ID and value is reference origins+targets for that language ID, but that may also be premature optimization at this point as we only have two IDs so far. 🤔 

You could just go with `VarsRefOrigins`+`VarsRefOriginsErr`+`VarsRefOriginsState` and relevant two methods for updating the state + result. Relatedly I suppose the existing `RefOrigins*` should at that point be renamed to `ModuleRefOrigins*`.

internal/decoder/decoder.go
```diff
@@ -66,7 +66,7 @@ func varsPathContext(mod *state.Module) (*decoder.PathContext, error) {
 		Files:            make(map[string]*hcl.File, 0),
 	}
 
-	for _, origin := range mod.RefOrigins {
+	for _, origin := range mod.VarsRefOrigins {
```
radeksimko, line 69: Oops, I didn't realize we were already using the _wrong_ references here. 🙈  I think that LOC `74-78` can be deleted in that case.

dbanck: That's why I thought I had to merge them, but it's much cleaner with new fields on `state.Module`.

internal/terraform/module/module_loader_test.go
```diff
@@ -64,6 +64,11 @@ func TestModuleLoader_referenceCollection(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	varsOriginsOp := NewModuleOperation(modPath, op.OpTypeDecodeVarsReferences)
+	err = ml.EnqueueModuleOp(varsOriginsOp)
```
radeksimko, line 68: I _think_ this part of the test isn't going to be very useful unless we also add tfvars file(s) to the test module in `testdata/single-root-no-modules`. It may be fine just to leave this test as it was though. The logic seems well tested in isolation.

radeksimko (commented):
Just one note about the test, but LGTM overall!

I'm however currently investigating [a bug](https://github.com/hashicorp/terraform-ls/issues/733) which may result in a patch release and so if you can avoid merging this until I'm done investigating that would be great!

radeksimko (approved):

radeksimko (approved):
Feel free to merge once conflicts are resolved

PR #729, opened 2021-11-30
Update neovim formatting example to use the sync version

kevinrambaud:
Neovim now exposes a sync version of the formatting method. With the current example, in some cases, formatting was happening after a file was saved as the `formatting` method is async.


radeksimko (approved):
LGTM, thanks.

PR #737, opened 2021-12-07
Switch to hc-install from tfinstall

### LINKED ISSUE #711: Switch to hc-install from tfinstall
### Use-cases

The part of tfexec library we use - tfinstall - responsible for installing Terraform in a few contexts - was extracted into a more generalized library https://github.com/hashicorp/hc-install

Upcoming release of tfexec will have tfinstall removed and we should migrate to hc-install for installation.

### Proposal

Find usage of tfinstall and replace the logic with equivalent hc-install logic.

rtim75:
Closes #711

---

- replaced all `tfinstall` usage with `hc-install` which first looks for tf in filesystem and if didn't find, installs specified/latest version
- in `exec` moved tf version into function arguments

### DIFF
--- a/internal/langserver/handlers/complete_test.go
+++ b/internal/langserver/handlers/complete_test.go
@@ -851,7 +851,9 @@ output "test" {
 func tfExecutor(t *testing.T, workdir, tfVersion string) exec.TerraformExecutor {
 	ctx := context.Background()
 	installDir := filepath.Join(workdir, "hcinstall")
-	err := os.MkdirAll(installDir, 0o755)
+	if err := os.MkdirAll(installDir, 0o755); err != nil {
+		t.Fatal(err)
+	}
 
 	i := hcinstall.NewInstaller()
 	v := version.Must(version.NewVersion(tfVersion))
--- a/internal/langserver/handlers/complete_test.go
+++ b/internal/langserver/handlers/complete_test.go
@@ -10,7 +10,11 @@ import (
 	"time"
 
 	"github.com/hashicorp/go-version"
-	"github.com/hashicorp/terraform-exec/tfinstall"
+	hcinstall "github.com/hashicorp/hc-install"
+	"github.com/hashicorp/hc-install/fs"
+	"github.com/hashicorp/hc-install/product"
+	"github.com/hashicorp/hc-install/releases"
+	"github.com/hashicorp/hc-install/src"
 	tfjson "github.com/hashicorp/terraform-json"
 	"github.com/hashicorp/terraform-ls/internal/langserver"
 	"github.com/hashicorp/terraform-ls/internal/langserver/session"
@@ -603,7 +607,7 @@ func TestVarsCompletion_withValidData(t *testing.T) {
 						"detail": "required, string",
 						"insertTextFormat":1,
 						"textEdit": {
-							"range": {"start":{"line":0,"character":0}, "end":{"line":0,"character":0}}, 
+							"range": {"start":{"line":0,"character":0}, "end":{"line":0,"character":0}},
 							"newText":"test"
 						}
 					}
@@ -844,14 +848,25 @@ output "test" {
 		}`)
 }
 
-func tfExecutor(t *testing.T, workdir, version string) exec.TerraformExecutor {
+func tfExecutor(t *testing.T, workdir, tfVersion string) exec.TerraformExecutor {
 	ctx := context.Background()
-	installPath := filepath.Join(t.TempDir(), "tfinstall")
-	err := os.MkdirAll(installPath, 0o755)
-	if err != nil {
-		t.Fatal(err)
-	}
-	execPath, err := tfinstall.Find(ctx, tfinstall.ExactVersion(version, installPath))
+	installDir := filepath.Join(workdir, "hcinstall")
+	err := os.MkdirAll(installDir, 0o755)
+
+	i := hcinstall.NewInstaller()
+	v := version.Must(version.NewVersion(tfVersion))
+
+	execPath, err := i.Ensure(ctx, []src.Source{
+		&fs.ExactVersion{
+			Product: product.Terraform,
+			Version: v,
+		},
+		&releases.ExactVersion{
+			Product:    product.Terraform,
+			Version:    v,
+			InstallDir: installDir,
+		},
+	})
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/internal/langserver/handlers/complete_test.go
+++ b/internal/langserver/handlers/complete_test.go
@@ -854,6 +854,11 @@ func tfExecutor(t *testing.T, workdir, tfVersion string) exec.TerraformExecutor 
 	if err := os.MkdirAll(installDir, 0o755); err != nil {
 		t.Fatal(err)
 	}
+	t.Cleanup(func() {
+		if err := os.Remove(installDir); err != nil {
+			t.Fatal(err)
+		}
+	})
 
 	i := hcinstall.NewInstaller()
 	v := version.Must(version.NewVersion(tfVersion))
@@ -872,6 +877,12 @@ func tfExecutor(t *testing.T, workdir, tfVersion string) exec.TerraformExecutor 
 	if err != nil {
 		t.Fatal(err)
 	}
+
+	t.Cleanup(func() {
+		if err := i.Remove(ctx); err != nil {
+			t.Fatal(err)
+		}
+	})
 
 	tfExec, err := exec.NewExecutor(workdir, execPath)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


internal/langserver/handlers/complete_test.go
```diff
@@ -844,14 +848,25 @@ output "test" {
...
+func tfExecutor(t *testing.T, workdir, tfVersion string) exec.TerraformExecutor {
 	ctx := context.Background()
-	installPath := filepath.Join(t.TempDir(), "tfinstall")
-	err := os.MkdirAll(installPath, 0o755)
-	if err != nil {
-		t.Fatal(err)
-	}
-	execPath, err := tfinstall.Find(ctx, tfinstall.ExactVersion(version, installPath))
+	installDir := filepath.Join(workdir, "hcinstall")
+	err := os.MkdirAll(installDir, 0o755)
```
radeksimko, line 854:
```suggestion
	if err := os.MkdirAll(installDir, 0o755); err != nil {
		t.Fatal(err)
	}
	t.Cleanup(func() {
		if err := os.Remove(installDir); err != nil {
			t.Fatal(err)
		}
	})
```

I think we'll need some error handling and cleanup here.

internal/langserver/handlers/complete_test.go
```diff
@@ -844,14 +848,25 @@ output "test" {
...
+		&fs.ExactVersion{
+			Product: product.Terraform,
+			Version: v,
+		},
+		&releases.ExactVersion{
+			Product:    product.Terraform,
+			Version:    v,
+			InstallDir: installDir,
+		},
+	})
```
radeksimko, line 869:
```suggestion
	})

	t.Cleanup(func() {
		if err := i.Remove(ctx); err != nil {
			t.Fatal(err)
		}
	})
```
The installer provides a method we can use to cleanup downloaded/installed files after testing.

internal/schemas/gen/gen.go
```diff
@@ -81,13 +85,22 @@ func gen() error {
...
 
-	tmpDir, err := ioutil.TempDir("", "tfinstall")
+	installDir, err := ioutil.TempDir("", "hcinstall")
 	if err != nil {
 		return err
 	}
-	defer os.RemoveAll(tmpDir)
-
-	execPath, err := tfinstall.Find(ctx, tfinstall.LookPath(), tfinstall.LatestVersion(tmpDir, false))
+	defer os.RemoveAll(installDir)
```
radeksimko, line 92:
```suggestion
	t.Cleanup(func() {
		if err := os.Remove(installDir); err != nil {
			t.Fatal(err)
		}
	})
```
`defer` is not wrong, but in the context of tests `t.Cleanup` is a bit more flexible especially once/if we add subtests and/or decide to run tests in parallel. It also allows us to fail if some files are left behind which would likely help us discover a bug upstream in hc-install `Remove`.

internal/schemas/gen/gen.go
```diff
@@ -81,13 +85,22 @@ func gen() error {
...
+	i := hcinstall.NewInstaller()
+	execPath, err := i.Ensure(ctx, []src.Source{
+		&fs.AnyVersion{
+			Product: &product.Terraform,
+		},
+		&releases.LatestVersion{
+			Product:    product.Terraform,
+			InstallDir: installDir,
+		},
+	})
```
radeksimko, line 103:
```suggestion
	})
	t.Cleanup(func() {
		if err := i.Remove(ctx); err != nil {
			t.Fatal(err)
		}
	})
```

internal/terraform/exec/exec_test.go
```diff
@@ -57,21 +62,29 @@ func TestExec_cancel(t *testing.T) {
...
-func newExecutor(t *testing.T) TerraformExecutor {
-	tmpDir := TempDir(t)
-	workDir := filepath.Join(tmpDir, "workdir")
-	err := os.MkdirAll(workDir, 0755)
-	if err != nil {
-		t.Fatal(err)
-	}
-	installDir := filepath.Join(tmpDir, "installdir")
-	err = os.MkdirAll(installDir, 0755)
+func newExecutor(t *testing.T, tfVersion string) TerraformExecutor {
```
radeksimko, line 65: Is there a particular reason for adding the version as an argument here? IMO we always want to test everything within this (`exec`) package against the same version, so I'd just set the version in one place.

internal/terraform/exec/exec_test.go
```diff
@@ -57,21 +62,29 @@ func TestExec_cancel(t *testing.T) {
...
+	ctx := context.Background()
+	workDir := TempDir(t)
+	installDir := filepath.Join(workDir, "hcinstall")
+	err := os.MkdirAll(installDir, 0755)
 	if err != nil {
 		t.Fatal(err)
 	}
```
radeksimko, line 72:
```suggestion
	if err := os.MkdirAll(installDir, 0755); err != nil {
		t.Fatal(err)
	}
	t.Cleanup(func() {
		if err := os.Remove(installDir); err != nil {
			t.Fatal(err)
		}
	})
```

radeksimko (commented):
Thank you for the PR!

I left you some suggestions in-line. I'm happy to discuss any point further.

internal/langserver/handlers/complete_test.go
rtim75: will we? the `installDir` will always be under the `workDir` which is created with the `TempDir` function(internal/langserver/handlers/handlers_test.go#222). The TempDir has a cleanup function already which means the whole `workDir` will be deleted when tests complete.

rtim75: I realise it's a bit obscure and relies on `TempDir()` usage, though

internal/schemas/gen/gen.go
rtim75: It's not a test function here.

internal/schemas/gen/gen.go
rtim75: It's not a test function here.

internal/terraform/exec/exec_test.go
rtim75: Oh, now I understand the reason to have version hardcoded in the `newExecutor` function. I only made it a variable to be aligned with another function that creates executors. I will update this one to have a static version.

internal/terraform/exec/exec_test.go
rtim75: `installDir` is under the `workDir`. It's created using the `TempDir()` function which registers a cleanup function the will delete the whole `workDir`. Do you want to have another cleanup function anyway?

internal/langserver/handlers/complete_test.go
rtim75: installDir is under the `workDir`. It's created using the `TempDir()` function which registers a cleanup function that will delete the whole `workDir`. Do you want to have another cleanup function anyway?

internal/langserver/handlers/complete_test.go
rtim75: the error handling was definitely missing here

internal/schemas/gen/gen.go
radeksimko: Ooops, yes - you're right, this isn't a test 😄  my bad

internal/langserver/handlers/complete_test.go
radeksimko: You're not wrong here in that it would get cleaned anyway but the benefit of explicit cleanup function is that we can catch test failures when removal of any individual file fails and we can investigate (as that would likely indicate a bug).

Also I think that the previous `t.TempDir()` was preferable over `workdir` as calling `TempDir()` in this context would create test-specific directory which will _not_ be shared with any other test, which in turn allows us to run these tests in parallel later. Also workdir would often point to some git-versioned test data and I think it's best to avoid creating files under test in places which are not git-ignored.

internal/terraform/exec/exec_test.go
radeksimko: See my comment above.

radeksimko (commented):
Thank you for making the changes. I'm happy to merge once the last two comments about cleanup are resolved.

internal/langserver/handlers/complete_test.go
radeksimko: See my comment in the other test - I agree it is is a bit obscure.

rtim75:
Thanks @radeksimko 
I've updated all the files to have explicit cleanups for install dirs as well as for installers.

rtim75:
I'm looking into the failing build

rtim75:
@radeksimko should the hpegl be excluded as well? It appears to have the same issue as in #692

radeksimko:
I have just addressed a few conflicts (mainly in `go.mod` & `go.sum`) and one remaining point about where Terraform gets installed, so we can get this merged - I hope you don't mind. Your original authorship remains intact.

Thank you for the contribution!

radeksimko (approved):

rtim75:
> I hope you don't mind
Not at all 😄 Thanks)

Issue #549, opened 2021-06-11
TestLangServer_workspaceExecuteCommand_modules_multiple fails on aarch64


06kellyjac:
### Server Version

```
v0.18.0
```

### Terraform Version

```
None
```

### Client Version

```
None
```

### Terraform Configuration Files

```hcl
None
```

### Log Output

Gist contains: Just the error section, full aarch64 failing build, and a full x86 passing build for comparison

https://gist.github.com/06kellyjac/a5699a23c1745911eaf981a28c14a467

### Expected Behavior

Tests should pass

### Actual Behavior

`TestLangServer_workspaceExecuteCommand_modules_multiple` fails but only on aarch64

### Steps to Reproduce

Either run the tests normally from this repo on an **aarch64** machine (`go test ./...` or `go test github.com/hashicorp/terraform-ls/internal/langserver/handlers`)
Or clone my nixpkgs repo and use branch `terraform-ls` on an **aarch64** machine and run `nix-build -A terraform-ls`

radeksimko:
Hi @06kellyjac 
Thanks for the report.

Unfortunately it seems that GitHub does not provide any ARM environment to run the tests in currently: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources which seems to be mostly blocked on lack of ARM support in Azure per https://github.com/actions/virtual-environments/issues/2552#issuecomment-771478000

We could in theory self host the ARM runners as per https://github.blog/changelog/2019-12-03-github-actions-self-hosted-runners-on-arm-architectures/ but that is obviously not as trivial (from ops perspective) as using GitHub-hosted runners.

I'm asking around internally (at HashiCorp) about that possibility, but generally until we have a way of testing this in CI it's unlikely that we'll find time to fix it.

That said if you can come up with a PR that fixes the test I'd be happy to review it!

06kellyjac:
Might be able to set up the tests in actions using QEMU like deno does using this: https://github.com/denoland/rusty_v8/blob/main/.github/workflows/ci.yml#L81-L96

I imagine it'll need some adapting to work for golang rather than rust but I might play with it

radeksimko:
I had a spare moment today to reproduce this on my Raspberry Pi and I have a hunch of what the root cause is.

I believe it's merely test data being limited to 3 combinations of OS & architecture:

[...]

PR #739, opened 2021-12-13
skip and fix failing tests on Apple Silicon

radeksimko:
Closes https://github.com/hashicorp/terraform-ls/issues/549

~The `build` CI failure can be fixed by https://github.com/hashicorp/terraform-ls/pull/740~

### DIFF
--- a/internal/terraform/exec/exec_test.go
+++ b/internal/terraform/exec/exec_test.go
@@ -70,7 +70,7 @@ func newExecutor(t *testing.T) TerraformExecutor {
 		t.Fatal(err)
 	}
 
-	opts := tfinstall.ExactVersion("0.13.1", installDir)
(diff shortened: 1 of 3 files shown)


internal/terraform/exec/exec_test.go
```diff
@@ -70,7 +70,7 @@ func newExecutor(t *testing.T) TerraformExecutor {
 		t.Fatal(err)
 	}
 
-	opts := tfinstall.ExactVersion("0.13.1", installDir)
+	opts := tfinstall.ExactVersion("1.1.0", installDir)
```
radeksimko, line 73: For posterity `1.0.2` was the first version to support `darwin/arm64`:
https://releases.hashicorp.com/terraform/1.0.2/

dbanck (approved):
LGTM 👍

PR #740, opened 2021-12-13
schemas: revise list of ignored providers

radeksimko:
This is to stop ignoring providers which are now `init`-able without issues and to add `nullstone-io/ns` which has invalid schema that Terraform v1.1 rejects when running most schema-related commands, such as `terraform providers schema -json`.


dbanck (approved):
LGTM

PR #749, opened 2021-12-30
Plumb a context in to the server.Loop call.

creachadair:


jpogran:
@creachadair can you describe what this PR fixes/solves? There is little detail or context here. Is there a reason you're targeting this at a branch and not main?

creachadair:
> @creachadair can you describe what this PR fixes/solves? There is little detail or context here. Is there a reason you're targeting this at a branch and not main?

Hi @jpogran, sorry for not including more detail. This patch is updating the API usage for the github.com/creachadair/jrpc2 dependency, so that Dependabot can merge its change. That's why it's targeted at the dependabot PR, though it could just as well be done separately (and I would be glad to send such a PR if you'd prefer it).

radeksimko (approved):
The "context plumbing" still needs some tidying in various places and I reckon that we'll want to pass some cancellable context here eventually, but the TCP mode is used only rarely, for debugging and the TODO() is fine for now anyway as it doesn't change the existing implementation anyway from what I can tell.

Thanks for the PR!

creachadair:
> the TODO() is fine for now anyway as it doesn't change the existing implementation anyway from what I can tell.

That’s correct. A background context preserves the existing behaviour without a context.

The main reason to add a non-trivial context here would be if you need to stop a loop and cancel all its running instances. When the listener closes, the loop stops accepting new connections but does not return until the servers that are already running have terminated normally. The context ending explicitly stops those servers first.

Issue #733, opened 2021-12-03
High CPU usage since 0.25.0


ElvenSpellmaker:
### Server Version

```
0.25.0
```

### Terraform Version

```
v1.0.2
```

### Client Version

```
VSCode 1.62.3
```

### Terraform Configuration Files

N/A perhaps?

### Log Output

There's lot of 
`2021/12/03 10:09:32 provider_schema.go:196: PSS: getting provider schema` in the vscode console log and it seems to be going over and over the repo again and again looking at the whole repo (which has a lot of Terraform in it spread out all over the place)

### Expected Behavior

It should work without high CPU

### Actual Behavior

![image](https://user-images.githubusercontent.com/2286713/144583595-ad50fb73-2218-49ea-bf9a-42d1ddc7ee5c.png)

The CPU varies from 50-100% consistently while in TF files.

The log when I last copied it had 184491 lines.

### Steps to Reproduce
Load some TF files in vscode since the latest update

radeksimko:
@ElvenSpellmaker Thanks for the report and sorry for the experience.

> looking at the whole repo (which has a lot of Terraform in it spread out all over the place)

This indexing is generally by design e.g. to enable discovery of cross-module references.

It has a cap on parallelism, but that is mostly just to reduce the peak CPU usage, it just means the indexing will take longer at the cost of lower CPU usage. We have plans to potentially allow this cap to be configurable as per 

I would say the easiest short-term solution, while we investigate, would be to open just a few relevant individual folders as opposed to the parent folder which contains many Terraform modules, or to set `"requiredVersion": "0.24.0"` in `"terraform.languageServer"` settings block in VS Code.

> The log when I last copied it had 184491 lines.

I would be very interested in the log as it could show us the sequence of operations that happen as part of the indexing. Are you willing to share it? Feel free to encrypt it with [my GPG key](https://keybase.io/radeksimko) and send to `radek<at>hashicorp.com` if it's sensitive and you don't want to post it publicly.

ElvenSpellmaker:
> @ElvenSpellmaker Thanks for the report and sorry for the experience.
> 
> > looking at the whole repo (which has a lot of Terraform in it spread out all over the place)
> 
> This indexing is generally by design e.g. to enable discovery of cross-module references.
Yeah I'd understand if it stopped after a while but it never stops, it's been at 50-100% CPU since I upgraded this morning at 8:42AM (now 10:40AM)

ElvenSpellmaker:
[...]

Issue #150, opened 2020-06-10
Report progress when obtaining schema


radeksimko:
### Current Version

```
0.3.2
```

### Use-cases

The language server has to obtain schema in order to have data for completion candidates and other features from whatever providers the user has in their configuration.

This currently happens literally behind the scenes, asynchronously, without the user knowing, which means the user isn't literally "blocked" while we're obtaining schema, but it does lead to a suboptimal UX:

1. user requests completion at pos `line:col`
2. server responds with no candidates and error "schema temporarily unavailable"
3. client caches this "no candidates" response
4. user retries few seconds later, when schema is available, but still gets "no candidates", because that's what's in cache, so client doesn't ask again for the same position, until the file is changed

Generally the user expects to be able to get completion straight away and it's not obvious why they aren't able to get it, or what the error ("schema temporarily unavailable") could mean and when they could retry.

### Proposal

Report progress while obtaining schema to the client/UI to better inform the user when this background operation is running and when they could expect completion to work.

### Related LSP methods

https://microsoft.github.io/language-server-protocol/specifications/specification-current/#window_workDoneProgress_create

https://microsoft.github.io/language-server-protocol/specifications/specification-current/#window_workDoneProgress_cancel

radeksimko:
Closing in favour of https://github.com/hashicorp/terraform-ls/issues/5 which now tracks progress reporting for the major time-consuming operations.

Issue #5, opened 2020-02-21
Implement progress reporting


radeksimko:
Per LSP spec:

https://microsoft.github.io/language-server-protocol/specifications/specification-current/#progress

MartinLoeper:
Just noticed the described behaviour in nvim too. It takes some time until all lsp features are up and running. Would be great to get some minimal progress hints :)

PR #752, opened 2022-01-04
module loader: Reduce parallelism to 1/1 in all cases

radeksimko:
This will cause indexing to take longer to finish on machines with >2 CPUs, but it should flatten the CPU spikes some users have reported in https://github.com/hashicorp/terraform-ls/issues/733

It is worth noting that this puts a bit more pressure on resolving https://github.com/hashicorp/terraform-ls/issues/5 / https://github.com/hashicorp/terraform-ls/issues/150

### DIFF
--- a/internal/terraform/module/module_loader.go
+++ b/internal/terraform/module/module_loader.go
@@ -3,7 +3,6 @@ package module
 import (
 	"context"
 	"log"
-	"runtime"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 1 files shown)


dbanck (approved):
LGTM!

Were you able to identify the module loader as source of the high cpu usage from the profiling results?

radeksimko:
> Were you able to identify the module loader as source of the high cpu usage from the profiling results?

In broader sense yes, based on a provided CPU profile from a user I believe the main root cause is the command(s) we execute for each (initialized) module. I reckon the ultimate solution will be sourcing the schemas from a (future) Registry API, which should consume far less CPU than running `terraform providers schema -json`.

Screenshot 2022-01-05 at 10 32 23

PR #755, opened 2022-01-06
fix: Catch OS agnostic interrupt signal

radeksimko:
While debugging https://github.com/hashicorp/terraform-ls/issues/733 Windows users have reported that CPU profile never gets written in their environment.

This is because CPU profile only gets written upon (graceful) shutdown which we perform by catching particular interrupt/termination signals. Prior to this patch however we were only catching Unix-native signals, which Windows never sends, resulting in dirty shutdowns and never giving the opportunity for a CPU profile to be written to disk.

See also: https://pkg.go.dev/os/signal#hdr-Windows and https://stackoverflow.com/a/35683558

### DIFF
--- a/internal/cmd/inspect_module_command.go
+++ b/internal/cmd/inspect_module_command.go
@@ -98,7 +98,7 @@ func (c *InspectModuleCommand) inspect(rootPath string) error {
 	walker.SetLogger(c.logger)
 
 	ctx, cancel := ictx.WithSignalCancel(context.Background(),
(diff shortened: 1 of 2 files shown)


dbanck (approved):
LGTM!

I found that it's still possible to get a 0kb file when writing CPU profiles. I assume that sometimes VSCode will relaunch the process, overwriting the "valid" profile with an empty 0kb one. The safest way to avoid this is to use a timestamp in the filename.

PR #757, opened 2022-01-07
fix: avoid code lens updates when disabled

radeksimko:
Previously we'd always send a notification to the client whenever reference counts changed. This would be ignored by the client if the code lens was disabled, so this patch is basically no-op from UX perspective, but it should save some overhead by not sending the notification at all when it's not needed.


dbanck (approved):
Cool! 👍

PR #758, opened 2022-01-10
telemetry: Only send requests if data has changed

radeksimko:
Depends on https://github.com/hashicorp/terraform-schema/pull/89
Depends on https://github.com/hashicorp/terraform-ls/pull/761

According to manual testing with a single installed module from https://registry.terraform.io/modules/terraform-aws-modules/eks/aws/latest this can reduce the amount of telemetry requests to about half (from 32 to 14).

There's probably more we can do here - e.g. avoid sending telemetry for unsaved data. This is not only extra overhead, but also pollutes the data with "unfinished" provider or backend names. However this would require more effort - probably queueing the telemetry data somewhere and then adding a hook which gets triggered by `textDocument/didSave`. Either way not something I wanted to tackle in this PR.

### DIFF
--- a/internal/langserver/handlers/hooks_module.go
+++ b/internal/langserver/handlers/hooks_module.go
@@ -12,86 +12,117 @@ import (
 )
 
 func sendModuleTelemetry(ctx context.Context, store *state.StateStore, telemetrySender telemetry.Sender) state.ModuleChangeHook {
-	return func(_, newMod *state.Module) {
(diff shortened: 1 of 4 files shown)


dbanck (approved):
Implementation LGTM! 🌟 

I wondered if we prefer to send a "full" properties body when something changes or just the change? For example, we're sending `providerRequirements` and `tfVersion` in the current implementation, even if just `tfRequirements` has changed.

PR #759, opened 2022-01-11
fix: Return correct target selection range for definition/declaration

### LINKED ISSUE #735: textDocument/definition returns targetSelectionRange with a start and end of (0,0).
Apologies for the lack of detail; I encountered this on a machine I cannot log into my personal Github account from.

### Server Version
0.25.0

### Terraform Version
1.0.3

### Client Version
emacs-lsp

Shouldn't be relevant

### Log Output
Hopefully not needed

### Expected Behavior
According to the spec, targetSelectionRange must contain targetRange. So I expect that to be the case.

### Actual Behavior
targetSelectionRange has a start and end at line 0 column 0, which does not contain targetRange.

### Steps to Reproduce
Set up terraform-ls with emacs-lsp and try to go to a definition with `lsp-find-definition`.

radeksimko:
Fixes #735 

Also fixes another bug where in `textDocument/definition` we previously checked `Declaration.LinkSupport` instead of `Definition.LinkSupport` client capability.

Lastly there was a (minor) bug in tests where we missed `textDocument` nesting, so link support was never tested.

### DIFF
--- a/internal/lsp/location_links.go
+++ b/internal/lsp/location_links.go
@@ -31,6 +31,7 @@ func refTargetToLocationLink(target *decoder.ReferenceTarget) lsp.LocationLink {
 		OriginSelectionRange: HCLRangeToLSP(target.OriginRange),
 		TargetURI:            lsp.DocumentURI(targetUri),
 		TargetRange:          HCLRangeToLSP(target.Range),
+		TargetSelectionRange: HCLRangeToLSP(target.Range),
 	}
 
 	if target.DefRangePtr != nil {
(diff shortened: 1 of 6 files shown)


internal/lsp/location_links.go
```diff
@@ -31,6 +31,7 @@ func refTargetToLocationLink(target *decoder.ReferenceTarget) lsp.LocationLink {
 		OriginSelectionRange: HCLRangeToLSP(target.OriginRange),
 		TargetURI:            lsp.DocumentURI(targetUri),
 		TargetRange:          HCLRangeToLSP(target.Range),
+		TargetSelectionRange: HCLRangeToLSP(target.Range),
 	}
 
 	if target.DefRangePtr != nil {
```
dbanck, line 37: Do we want to add a test case for the situation when `TargetSelectionRange` gets overwritten by this condition?

dbanck (approved):
LGTM! Good job catching the other two bugs, too!

internal/lsp/location_links.go
radeksimko: Added a test case.

PR #760, opened 2022-01-11
schemas: Ignore HewlettPackard/hpegl

radeksimko:
This is to address the fact that the provider doesn't have any release other than pre-release, which makes it impossible to install without explicit version (which is how we install it internally).

```
Error: Failed to query available provider packages

Could not retrieve the list of available versions for provider
hewlettpackard/hpegl: no available releases match the given constraints 
```

Related: https://github.com/hashicorp/terraform/issues/30337


dbanck (approved):
LGTM!

PR #773, opened 2022-01-25
Sign releases using signore

### LINKED ISSUE #688: Change checksum signing to use signore
Currently we use `gpg` directly with a provided GPG key to sign the archive checksums:

https://github.com/hashicorp/terraform-ls/blob/36e53874b6518e3cbe756d80230c09d6f7ace02e/.goreleaser.yml#L76-L88

The key we use for signing will expire in April 2022.

We should use the internal HashiCorp signing service (signore). The client can be installed either via https://github.com/hashicorp/setup-signore or https://github.com/hashicorp/setup-signore-package - we can use the latter as we release on Linux.

https://github.com/hashicorp/signore/blob/main/docs/SIGNING_IN_CI.md

dbanck:
This PR updates the release signing to use the internal HashiCorp signing service (signore).

Closes #688.

Update secrets before merging:
- [x] add `SIGNORE_CLIENT_ID`
- [x] add `SIGNORE_CLIENT_SECRET`
- [x] add `SIGNORE_SIGNER`

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -76,7 +76,8 @@ checksum:
 signs:
   -
     id: with_key_id
-    signature: "${artifact}.{{ .Env.GPG_KEY_ID }}.sig"
+    # TODO: Replace with variable once signore provides the primary_key_id
+    signature: "${artifact}.72D7468F.sig"
     cmd: signore
     args: ["sign", "--dearmor", "--file", "${artifact}", "--out", "${signature}"]
     artifacts: checksum
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -76,8 +76,7 @@ checksum:
 signs:
   -
     id: with_key_id
-    # TODO: Replace with variable once goreleaser supports variables here
-    signature: "${artifact}.72D7468F.sig"
+    signature: "${artifact}.{{ .Env.GPG_KEY_ID }}.sig"
     cmd: signore
     args: ["sign", "--dearmor", "--file", "${artifact}", "--out", "${signature}"]
     artifacts: checksum
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -78,12 +78,14 @@ signs:
     id: with_key_id
     # TODO: Replace with variable once goreleaser supports variables here
     signature: "${artifact}.72D7468F.sig"
-    args: ["--batch", "--no-tty", "-u", "{{ .Env.GPG_KEY_ID }}", "--output", "${artifact}.{{ .Env.GPG_KEY_ID }}.sig", "--detach-sign", "${artifact}"]
+    cmd: signore
+    args: ["sign", "--dearmor", "--file", "${artifact}", "--out", "${signature}"]
     artifacts: checksum
   -
     id: default
     signature: "${artifact}.sig"
-    args: ["--batch", "--no-tty", "-u", "{{ .Env.GPG_KEY_ID }}", "--output", "${artifact}.sig", "--detach-sign", "${artifact}"]
+    cmd: signore
+    args: ["sign", "--dearmor", "--file", "${artifact}", "--out", "${signature}"]
     artifacts: checksum
 
 brews:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -65,6 +65,9 @@ jobs:
           role-skip-session-tagging: true
           role-duration-seconds: 3600
       -
+        name: Install signore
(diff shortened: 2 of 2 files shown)


.goreleaser.yml
```diff
@@ -76,14 +76,15 @@ checksum:
 signs:
   -
     id: with_key_id
-    # TODO: Replace with variable once goreleaser supports variables here
-    signature: "${artifact}.72D7468F.sig"
-    args: ["--batch", "--no-tty", "-u", "{{ .Env.GPG_KEY_ID }}", "--output", "${artifact}.{{ .Env.GPG_KEY_ID }}.sig", "--detach-sign", "${artifact}"]
+    signature: "${artifact}.{{ .Env.GPG_KEY_ID }}.sig"
```
dbanck, line 79: Would it be possible to get the key id from signore, so it can be removed from our repository secrets?

milescrabill: This is a request other folks have made, and a good call. I've been mildly mixed on how to handle this because we're identifying signatures in [releases](https://releases.hashicorp.com) with the key id of the primary key but signing with a subkey (which is trusted by the primary key). I can add something like `key_id` and `primary_key_id` to the signers metadata, I just don't want people to get confused and use the subkey key id.

The key id for the current HashiCorp Security key is [`72D7468F`](https://keys.openpgp.org/vks/v1/by-fingerprint/C874011F0AB405110D02105534365D9472D7468F).

dbanck: Thanks for the explanation! Having something like `key_id` and `primary_key_id` sounds useful.

I intend to reduce the number of things that need to be updated when a key expires or needs rotation. So, if we can avoid having a hard-coded value here, there would be one less file to update.

How does signore handle the expiration of subkeys? For example, do we need to create a new client in April to use a new signer with the new subkey? Or is signore taking care of this for us?

milescrabill: My plan is to add a new signer for the "long-term" signing subkey and alias the current interim signing subkey signer to use that. The signore client will be changed to add a soft deprecation notice that users should update the signer in their config. We're planning on re-signing everything in the releases S3 bucket signed with the interim key with the long-term key before then as well.

So essentially, there shouldn't be any need for signore users to change anything come April, except to shush the deprecation notice.

dbanck: Great! Having an alias to the new signing subkey simplifies things.

I'll keep `72D7468F` in here until we can get the ID from signore. Thanks again for taking the time to explain things and reviewing the PR.

.goreleaser.yml
```diff
@@ -76,14 +76,16 @@ checksum:
 signs:
   -
     id: with_key_id
-    # TODO: Replace with variable once goreleaser supports variables here
+    # TODO: Replace with variable once signore provides the primary_key_id
```
radeksimko, line 79: This is a good idea in principle but I assume that we'd have to obtain the key ID in a separate step and _then_ pass it down as ENV variable to goreleaser, right?

Presumably we'd want to be sure that the key obtained is the same key actually used for signing - so I suppose signore would then need to accept the key ID as a flag and reject signing if it doesn't match or something - just to prevent a rare case where the active key is changed between the two steps and we'd then end up publishing `sig` file under key ID which doesn't reflect the real key used.

radeksimko (approved):
LGTM, I'm assuming you tested this in a separate GHA run in a fork?

If so - feel free to merge! 🚢

.goreleaser.yml
dbanck: Yes, I would obtain the id separately and pass it down as an ENV variable to goreleaser.

I assume that obtaining the ID would work by using the `signore` binary together with the `SIGNORE_SIGNER` secret, like `signore get_signer $SIGNORE_SIGNER`. The returned object will have metadata containing a `primary_key_id` (the primary key) and a `key_id` (the subkey used for signing). I understand that there could be a rare case that the signer gets updated after obtaining the key id and before signing the binary. But I think it is much more likely having the id hardcoded here and forgetting to change it.

A way to avoid this case would be to have signore return some metadata with the `sign` call. Then, we can rename the `.sig` file with the correct key id from the metadata.

radeksimko: > Then, we can rename the .sig file with the correct key id from the metadata.

That could be a little tricky because AFAIK GoReleaser uses the pre-declared `signature` under `signs` as a way of tracking the artifacts which are then passed e.g. to `hc-releases` later.

However I was merely curious about the context behind the comment at this point. We can cross the bridge when we come to it. 🤷🏻‍♂️ 😄

PR #781, opened 2022-02-03
Speed up CI testing

jpogran:
Previously we ran tests and built official and partner provider schemas on every PR. This list now tops over 200 providers to download and parse.

This commit adds a new CI_PRIMARY_PROVIDERS environment variable which when set to `true` will only download providers for AWS, Google, and Azure. This will provide enough of a canary to check if any changes break schema parsing while providing fast PR runs. The full list of schemas will be downloaded and parsed in the release.yml file.


radeksimko (commented):
> This commit adds a new CI_PRIMARY_PROVIDERS environment variable which when set to true will only download providers for AWS, Google, and Azure. This will provide enough of a canary to check if any changes break schema parsing while providing fast PR runs. The full list of schemas will be downloaded and parsed in the release.yml file.

I guess it depends on what we're trying to check - what you're proposing would only check whether the schema obtaining logic is _entirely broken_. What we actually need to catch are the individual providers which have broken releases either because the maintainers broke it somehow (e.g. somehow didn't upload a build for linux/amd64 or didn't upload checksum or uploaded wrong checksum) or because there's a new provider which only has pre-releases and we cannot download these with the existing logic because `terraform init` by assumes latest _stable_ release if `version` is not declared.
Historically the providers which broke were never the official ones (and certainly not the top 3) but rather partner ones.

https://github.com/hashicorp/terraform-ls/blob/b13641c6b00755dc78c46b540befa7b2059cf39f/internal/schemas/gen/gen.go#L247-L254

> This commit also changes the build steps to build against a matrix of only windows, ubuntu and macos. These platforms are our primary targets. We don't need to build all targets for every PR.

While broken builds outside of the three targets are relatively rare, they tend to be the worst to debug and so I'd prefer to keep running these for every PR to avoid last-minute debugging after merge.

GoReleaser already builds multiple targets in parallel, the only limitation is available CPU (parallelisation defaults to no of CPUs) - I'm assuming your motivation for the use of the matrix feature is that each run within the matrix is a separate container/VM, so it gets dedicated CPU? If so - that sounds like a good idea - but we can as well just set different `GOOS` & `GOARCH` and run everything on `ubuntu-latest` - and avoid the extra licensing costs and any likely overhead of macOS/Windows compared to Linux?

Some alternative ideas:

 - Removing go-getter as a dependency would reduce the compilation time to about half (while still keeping all targets) - see https://github.com/hashicorp/terraform-ls/runs/5063860108?check_suite_focus=true (6mins vs last `main` build of ~12mins)
 - We could remove `go test` from goreleaser, assuming that we run tests in a dedicated workflow anyway
 - Decoupling the schema obtaining into its own workflow would allow us to [filter changes](https://github.com/hashicorp/terraform-provider-aws/blob/bb560f77070a5b5aada8acd1d58df7a8b25c1f71/.github/workflows/acctest-terraform-lint.yml#L7-L14) to only run it on PRs which change `./internal/schemas` so we avoid running it entirely on the majority of PRs which don't change this codepath
   - the same workflow can also run on push to `main` and/or nightly - to keep checking the _external factors_

.github/workflows/ci.yml
```diff
@@ -68,26 +91,34 @@ jobs:
...
+          # In order:
+          # * Module download cache
+          # * Build cache (Linux)
+          # * Build cache (Mac)
+          # * Build cache (Windows)
+          path: |
+            ~/go/pkg/mod
+            ~/.cache/go-build
+            ~/Library/Caches/go-build
+            ~/AppData/Local/go-build
```
radeksimko, line 118: I wish this was available as some kind of variable from somewhere - it must have been lots of trial & error and/or documentation reading to put all these paths together! 😅 

Am I right that `%LocalAppData%\go-build` ~= `~/AppData/Local/go-build` but we just assume the variable is never going to be set to anything else within the Windows GHA environment? https://github.com/actions/cache/blob/main/examples.md#windows-1

.github/workflows/ci.yml
```diff
@@ -68,26 +91,34 @@ jobs:
...
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
       -
```
radeksimko, line 122:
```suggestion
          key: ${{ runner.os }}-${{ runner.arch }}-go-${{ hashFiles('**/go.sum') }}
          restore-keys: |
            ${{ runner.os }}-${{ runner.arch }}-go-
      -
```
Nitpick - I believe the cache _could_ differ by architecture as well. Admittedly we don't run anywhere else other than on Intel right now, but this minor change would allow us to do in the future.

.github/workflows/ci.yml
```diff
@@ -18,6 +18,26 @@ jobs:
...
+            goarch: 'amd64'
+          - goos: 'darwin'
+            goarch: 'arm64'
+
+          - goos: 'linux'
+            goarch: 'amd64'
+          - goos: 'linux'
+            goarch: 'arm64'
+          - goos: 'linux'
+            goarch: 'arm'
```
radeksimko, line 40: I'm fine with this shorter list for a PR-based workflow, but I think we should continue building all platforms on push to `main` (i.e. after PR is merged). Unfortunately both of these are now handled by the same workflow - I'm not sure if there's any better solution other than duplicating the workflow into a separate YAML? 😞

.github/workflows/ci.yml
jpogran: I did start out with that example, but https://github.com/mvdan/github-actions-golang#how-do-i-set-up-caching-between-builds lumped them all together to reduce multiple filters. This seems to work in a 'if this folder doesn't exist then don't cache it' sort of way.

The `~/` path thing came from https://github.com/actions/cache/issues/705, which details how little this action is documented for different platforms. I wasn't seeing anything cached for `%LocalAppData%`, but did get for `~/AppData` so I switched. This does kinda make sense, since the default shell on Windows is now pwsh for Github Actions, and pwsh does not use % for environment variables.

We could dig further into this, but https://github.com/actions/cache/issues/301, makes it pretty obvious this action wasn't designed for windows so we should probably abandon using it until that is fixed. It currently takes longer to download and extract the cache than it does to run `go mod download`

.github/workflows/ci.yml
jpogran: We have a [10GB max limit](https://github.com/actions/cache#cache-limits) of the size of things we can cache, so we'll have to monitor it. My local `go-build` is 1GB, not including the other locations, so multiplying that by the amount of platforms/architectures and we could hit that limit soon.

.github/workflows/ci.yml
jpogran: We can use reusable workflows to split out the core parts of this one workflow, then have the scheduled events or triggers handle executing each extracted workflow. Another approach would be splitting up the goreleaser file into multiple files and having different scheduled actions.

There will be duplication any way we go, but this might not really be a limitation per se. We have different purposes for each part, and clearly defined 'rules' on when they should run. Splitting these out into 'unit' or 'acceptance' with scheduled or triggered would reduce resource costs and time the user is waiting.

.github/workflows/ci.yml
radeksimko: Makes sense! Thanks for the extra context.

.github/workflows/ci.yml
radeksimko: Yep - I don't think that would affect the existing workflow though - because `runner.arch` will be `amd64` in all cases - so the same amount of data would just get cached under a different key.

Again - not a big deal - not a blocker for the PR!

PR #783, opened 2022-02-04
Remove go-getter & simplify module source detection

radeksimko:
This simplifies the relatively non-critical part of code which is responsible for detecting module source (which has the effect of displaying different icons in the UI).

Here I'm trading-off possibly less accuracy for faster compile times by removing go-getter as a dependency.

In GitHub CI, the cross-compilation was previously around 8-11mins:

 - https://github.com/hashicorp/terraform-ls/runs/5336086783?check_suite_focus=true
 - https://github.com/hashicorp/terraform-ls/runs/5336082442?check_suite_focus=true
 - https://github.com/hashicorp/terraform-ls/runs/5335277276?check_suite_focus=true
 - https://github.com/hashicorp/terraform-ls/runs/5336105445?check_suite_focus=true

This should shave off at least a few minutes on average - the CI run associated with this PR reported `5m 33s`.

### DIFF
--- a/internal/terraform/datadir/module_types.go
+++ b/internal/terraform/datadir/module_types.go
@@ -1,7 +1,8 @@
 package datadir
 
 import (
-	"github.com/hashicorp/go-getter"
+	"strings"
+
 	tfregistry "github.com/hashicorp/terraform-registry-address"
 )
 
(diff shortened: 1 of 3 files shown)


jpogran (approved):
Go go speed racer!

internal/terraform/datadir/module_types.go
```diff
@@ -15,6 +16,13 @@ const (
 	GIT        ModuleType = "git"
 )
 
+var moduleSourceLocalPrefixes = []string{
+	"./",
+	"../",
+	".\\",
+	"..\\",
+}
+
```
jpogran, line 25: 👍🏼

PR #789, opened 2022-02-11
Add String() for `OpTypeDecodeVarsReferences`

radeksimko:
This is mostly for convenience in debug messages, which otherwise come through as not-very-descriptive `OpType(9)`.


dbanck (approved):
LGTM!

I think I missed running some code generation when adding the variable reference back then?

radeksimko:
> I think I missed running some code generation when adding the variable reference back then?

```
go generate internal/terraform/module/operation
```
is the magic command.

And yeah - it's easy to miss - I suppose we could add some checks in CI, but given how often we add new types I'm not sure it's worth the effort. 🤷🏻

Issue #724, opened 2021-11-22
Index uninitalized modules


radeksimko:
### Background

The server implements an "indexer" (internally called "walker") which indexes modules by walking the file hierarchy. This allows the server to maintain data about modules which can be used for sourcing (more) relevant schema, providing cross-module references etc.

### Use-cases

We have historically only indexed initialized modules (i.e. those which have `.terraform` directory). This was partially to reduce performance impact in the early versions of LS when we were indexing "synchronously" as part of responding to the `initialize` request and secondly because we only used the module manifest and provider binaries, not the configuration itself.

A new use case comes into play with cross-module "go-to-definition" and "go-to-references" (introduced in https://github.com/hashicorp/terraform-ls/pull/700), where it would be beneficial to index all modules, even uninitialized ones.

### Proposal

 - consider deprecating watcher in favour of LSP-native watching capability https://github.com/hashicorp/terraform-ls/issues/15
 - consider refactoring module loader as a more generic "task scheduler" which can be more easily tested
 - remove module manager
 - consider refactoring walker as "indexer"
 - Introduce `ast.IsFilenameIndexable(file string) bool` to [`ast` package](https://github.com/hashicorp/terraform-ls/tree/main/internal/terraform/ast)
 - replace `.terraform` conditional logic with some logic which detects any first indexable file in a directory and schedules the module for indexing and skips any other files within that directory

danishprakash:
Hey @radeksimko, is there anything I can pick up? I see a few recent enhancement-related issues filed by you, do you have a plan in mind for those or can I look into one of them? Please reply only when you've some time. Thanks.

PR #790, opened 2022-02-12
internal/service: add support for didChangeWatchedFiles

### LINKED ISSUE #15: Support workspace/didChangeWatchedFiles
## Background

LSP method `textDocument/didChange` (which is already implemented) can be used to find out about changes made through the IDE, but sometimes files are changed outside of the IDE.

These can be detected through [`workspace/didChangeWatchedFiles`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workspace_didChangeWatchedFiles) which the LS doesn't implement yet.

Clients may choose to send pretty much arbitrary updates to the server and it is up to the server to filter them. Ideally though, the server should document what clients should watch so it doesn't have to filter ~garbage~ unnecessary updates.

Server may also _add_ to the list of watched globs, which is a feature described under https://github.com/hashicorp/terraform-ls/issues/867

## UX Impact

Users will see completion, go-to-definition or go-to-references etc. involving indexed but unopened modules to be accurate even when these modules change outside of the editor - e.g. when git branch is switched or files are changed in another editor.

## Proposal

 - [ ] Implement/register new handler for `workspace/didChangeWatchedFiles`
 - [ ] Filter files based on appropriate criteria
     - `*.tf` and `*.tfvars` files
     - directory is already tracked as a module, i.e. needs updating https://github.com/hashicorp/terraform-ls/blob/c277ec6e4daf7225fbcd655cd444a251ad3572d8/internal/langserver/handlers/did_change.go#L49-L53
  - [ ] Re-parse the directory in which relevant files were changed/added/delete https://github.com/hashicorp/terraform-ls/blob/c277ec6e4daf7225fbcd655cd444a251ad3572d8/internal/langserver/handlers/did_change.go#L55-L58
  - [ ] Document what files does the server expect to be watched by default in https://github.com/hashicorp/terraform-ls/blob/main/docs/language-clients.md

danishprakash:
Closes #15

### DIFF
--- a/internal/langserver/handlers/did_change_watched_files.go
+++ b/internal/langserver/handlers/did_change_watched_files.go
@@ -1,0 +1,71 @@
+package handlers
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/creachadair/jrpc2"
+	"github.com/hashicorp/terraform-ls/internal/job"
+	ilsp "github.com/hashicorp/terraform-ls/internal/lsp"
+	"github.com/hashicorp/terraform-ls/internal/protocol"
+	lsp "github.com/hashicorp/terraform-ls/internal/protocol"
+	"github.com/hashicorp/terraform-ls/internal/terraform/ast"
+)
+
+func (svc *service) DidChangeWatchedFiles(ctx context.Context, params lsp.DidChangeWatchedFilesParams) error {
+	var ids job.IDs
+
+	for _, change := range params.Changes {
+		uri := string(change.URI)
+		_, err := os.Stat(uri)
+		if err != nil {
+			jrpc2.ServerFromContext(ctx).Notify(ctx, "window/showMessage", &lsp.ShowMessageParams{
+				Type:    lsp.Warning,
+				Message: fmt.Sprintf("Unable to update: %s, Failed to open directory", uri),
+			})
+			continue
+		}
+
+		// `uri` can either be a file or a director baed on the spec.
+		// We're not making any assumptions on the above and passing
+		// the uri as the module path itself for validation.
+		//
+		// https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeWatchedFiles
+		if !ast.IsModuleFilename(uri) && !ast.IsVarsFilename(uri) {
+			jrpc2.ServerFromContext(ctx).Notify(ctx, "window/showMessage", &lsp.ShowMessageParams{
+				Type:    lsp.Warning,
+				Message: fmt.Sprintf("Unable to update file: %s, filetype not supported", uri),
+			})
+			continue
+		}
+
+		// only handle `Changed` event type
+		if change.Type == protocol.Changed {
+			dh := ilsp.HandleFromDocumentURI(change.URI)
+
+			// check existence
+			_, err = svc.modStore.ModuleByPath(dh.Dir.Path())
+			if err != nil {
+				continue
+			}
+
+			jobIds, err := svc.parseAndDecodeModule(dh.Dir)
+			if err != nil {
+				continue
+			}
+
+			ids = append(ids, jobIds...)
+
+		}
+
+	}
+
+	// wait for all jobs (slowest usually) to complete
+	err := svc.stateStore.JobStore.WaitForJobs(ctx, ids...)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/internal/langserver/handlers/did_change_watched_files.go
+++ b/internal/langserver/handlers/did_change_watched_files.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"path"
+	"path/filepath"
 
 	"github.com/creachadair/jrpc2"
 	"github.com/hashicorp/terraform-ls/internal/document"
@@ -37,7 +37,7 @@ func (svc *service) DidChangeWatchedFiles(ctx context.Context, params lsp.DidCha
 			}
 
 			// 2nd we try again assuming it is a file
-			parentDir := path.Dir(rawPath)
+			parentDir := filepath.Dir(rawPath)
 			_, err = svc.modStore.ModuleByPath(parentDir)
 			if err != nil {
 				svc.logger.Printf("error finding module (%q deleted): %s", parentDir, err)
--- a/internal/langserver/handlers/did_change_watched_files.go
+++ b/internal/langserver/handlers/did_change_watched_files.go
@@ -6,62 +6,64 @@ import (
 	"os"
 
 	"github.com/creachadair/jrpc2"
+	"github.com/hashicorp/terraform-ls/internal/document"
 	"github.com/hashicorp/terraform-ls/internal/job"
-	ilsp "github.com/hashicorp/terraform-ls/internal/lsp"
 	"github.com/hashicorp/terraform-ls/internal/protocol"
 	lsp "github.com/hashicorp/terraform-ls/internal/protocol"
 	"github.com/hashicorp/terraform-ls/internal/terraform/ast"
+	"github.com/hashicorp/terraform-ls/internal/uri"
 )
 
 func (svc *service) DidChangeWatchedFiles(ctx context.Context, params lsp.DidChangeWatchedFilesParams) error {
 	var ids job.IDs
 
 	for _, change := range params.Changes {
-		uri := string(change.URI)
-		_, err := os.Stat(uri)
+		rawURI := string(change.URI)
+
+		fullPath, err := uri.PathFromURI(rawURI)
 		if err != nil {
-			jrpc2.ServerFromContext(ctx).Notify(ctx, "window/showMessage", &lsp.ShowMessageParams{
-				Type:    lsp.Warning,
-				Message: fmt.Sprintf("Unable to update: %s, Failed to open directory", uri),
-			})
+			svc.logger.Printf("Unable to update %q: %s", rawURI, err)
 			continue
 		}
 
-		// `uri` can either be a file or a director baed on the spec.
-		// We're not making any assumptions on the above and passing
-		// the uri as the module path itself for validation.
-		//
-		// https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeWatchedFiles
-		if !ast.IsModuleFilename(uri) && !ast.IsVarsFilename(uri) {
-			jrpc2.ServerFromContext(ctx).Notify(ctx, "window/showMessage", &lsp.ShowMessageParams{
-				Type:    lsp.Warning,
-				Message: fmt.Sprintf("Unable to update file: %s, filetype not supported", uri),
-			})
+		fi, err := os.Stat(fullPath)
+		if err != nil {
+			svc.logger.Printf("Unable to update %q: %s ", fullPath, err)
 			continue
 		}
 
-		// only handle `Changed` event type
+		// URI can either be a file or a directory based on the LSP spec.
+		var dirHandle document.DirHandle
+		if !fi.IsDir() {
+			if !ast.IsModuleFilename(fi.Name()) && !ast.IsVarsFilename(fi.Name()) {
+				jrpc2.ServerFromContext(ctx).Notify(ctx, "window/showMessage", &lsp.ShowMessageParams{
+					Type: lsp.Warning,
+					Message: fmt.Sprintf("Unable to update %q: filetype not supported. "+
+						"This is likely a bug which should be reported.", fullPath),
+				})
+				continue
+			}
+			docHandle := document.HandleFromPath(fullPath)
+			dirHandle = docHandle.Dir
+		} else {
+			dirHandle = document.DirHandleFromPath(fullPath)
+		}
+
 		if change.Type == protocol.Changed {
-			dh := ilsp.HandleFromDocumentURI(change.URI)
-
-			// check existence
-			_, err = svc.modStore.ModuleByPath(dh.Dir.Path())
+			_, err = svc.modStore.ModuleByPath(dirHandle.Path())
 			if err != nil {
 				continue
 			}
 
-			jobIds, err := svc.parseAndDecodeModule(dh.Dir)
+			jobIds, err := svc.parseAndDecodeModule(dirHandle)
 			if err != nil {
 				continue
 			}
 
 			ids = append(ids, jobIds...)
-
 		}
-
 	}
 
-	// wait for all jobs (slowest usually) to complete
 	err := svc.stateStore.JobStore.WaitForJobs(ctx, ids...)
 	if err != nil {
 		return err
--- a/internal/langserver/handlers/did_change_watched_files.go
+++ b/internal/langserver/handlers/did_change_watched_files.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"path"
 
 	"github.com/creachadair/jrpc2"
 	"github.com/hashicorp/terraform-ls/internal/document"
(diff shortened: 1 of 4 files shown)


internal/langserver/handlers/did_change_watched_files.go
```diff
@@ -0,0 +1,40 @@
...
+		switch change.Type {
+		case protocol.Created:
+			err := fs.CreateAndOpenDocument(fh, fh.Filename(), []byte(""))
+			if err != nil {
+				return err
+			}
+
+		case protocol.Changed:
+			// fs.ChangeDocument(..)
+			// fs.GetDocument(..)
```
danishprakash, line 29: I'm not exactly clear if it's just okay to load this "changed" document. Since there's no content or other meta for the document, I'm guessing this is the way to move forward? If yes, then should we check if it's part of a module and accordingly load the doc again?

internal/langserver/handlers/did_change_watched_files.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"context"
+
+	lsctx "github.com/hashicorp/terraform-ls/internal/context"
+	ilsp "github.com/hashicorp/terraform-ls/internal/lsp"
+	"github.com/hashicorp/terraform-ls/internal/protocol"
+	lsp "github.com/hashicorp/terraform-ls/internal/protocol"
+)
+
+func DidChangeWatchedFiles(ctx context.Context, params lsp.DidChangeWatchedFilesParams) error {
```
danishprakash, line 12: I've missed doing module-level configuration here for the passed document but since this notification spans across multiple documents, will it be a good idea to handle them based on the `FileChangeType`?

internal/langserver/handlers/did_change_watched_files.go
danishprakash: cc/ @radeksimko

radeksimko:
Hi @danishprakash 
Sorry for the delay in reviewing this PR.

There is a few PRs I'd like to merge first before looking at this one. Specifically https://github.com/hashicorp/terraform-ls/pull/839 and https://github.com/hashicorp/terraform-ls/pull/843

Thank you for the patience.

radeksimko:
With regards to the implementation and to answer some of your questions above:

I'd have to test this for real but my assumption is that `workspace/didChangeWatchedFiles` is perceived more like a "trigger" notification, primarily for files which are _not_ open. Files which _are_ open would presumably still receive `textDocument/didChange` with the new content. Closed files are just assumed to be saved to disk by the time we receive the `workspace/didChangeWatchedFiles` notification - so we shouldn't need to be updating anything internally.

So with that in mind I _think_ that the implementation could be simplified to basically this

https://github.com/hashicorp/terraform-ls/blob/8529ae5deb9c2574511327f07eb5953715910fe0/internal/langserver/handlers/did_change.go#L55-L60

but of course we'd first have to filter/deduplicate it down to just directories (module paths) and decide what do we do with files which the server doesn't treat as editable (i.e. anything other than `*.tf` or `*.tfvars`). I think we could probably just ignore and raise warnings for those - as [per docs](https://github.com/hashicorp/terraform-ls/blob/99e30e1ac0f0c58b495c4e0b9d27fabbab657857/docs/language-clients.md#language-ids) clients aren't supposed to send it to server anyway.

radeksimko:
@danishprakash The mentioned PRs were both now merged. When you have a moment - do you mind rebasing your PR?

Let me know if you need any further help beyond the answers I provided.

danishprakash:
@radeksimko thanks for the explanation. That helped a lot, however, I still have a few doubts/questions about some of those points, please correct me If I'm wrong or point me to the right resource If I seem to have missed something but basically
> workspace/didChangeWatchedFiles is perceived more like a "trigger" notification, primarily for files which are not open.

I understand we have `textDocument/didChange` for file changes but the spec doesn't mention the above explicitly, right? It simply states the following:
```
The watched files notification is sent from the client to the server 
when the client detects changes to files and folders watched by the 
language client
```

> Closed files are just assumed to be saved to disk by the time we receive the workspace/didChangeWatchedFiles notification - so we shouldn't need to be updating anything internally.

Again, I might be missing something here but based on the spec, can we say the above is an assumption at best?

danishprakash:
@radeksimko thanks for the explanation. That helped a lot, however, I still have a few doubts/questions about some of those points, please correct me If I'm wrong or point me to the right resource If I seem to have missed something but basically
> workspace/didChangeWatchedFiles is perceived more like a "trigger" notification, primarily for files which are not open.

I understand we have `textDocument/didChange` for file changes but the spec doesn't mention the above explicitly, right? It simply states the following:
```
The watched files notification is sent from the client to the server 
when the client detects changes to files and folders watched by the 
language client
```

> Closed files are just assumed to be saved to disk by the time we receive the workspace/didChangeWatchedFiles notification - so we shouldn't need to be updating anything internally.

Again, I might be missing something here but based on the spec, can we say the above is an assumption at best?

Finally, I'm also assuming based on the discussion above that we're going to ignore `CREATE` type event since we're receiving `workspace/didCreateFiles` event with the required metadata?

radeksimko:
You are right in that there are unfortunately parts of LSP which are not explicitly documented within the spec but servers pretty much rely on certain client implementations, most often the VS Code client library. The assumptions I made above are based on the VS Code client library's behaviour.

In general the spec doesn't seem to provide a lot of details on what the server should do with received documents from `textDocument/didOpen` and other text sync notifications, but AFAICT most servers do what this LS does - correct me if I'm wrong.

All I can find is [this](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_didOpen)

> The document open notification is sent from the client to the server to signal newly opened text documents. The document’s content is now managed by the client and the server must not try to read the document’s content using the document’s Uri.

and "read the document’s content using the document’s Uri" I'd interpret as "read the document's content from the disk/filesystem".

The spec also doesn't explicitly call out how remote environments should be handled on either client or server side. I'm guessing the assumption is mostly that LSP generally won't be used in a remote environment but I haven't researched this too much.

If you have some other assumptions/interpretations mismatching mine I'd be happy to discuss them! 😄 Relatedly I'm sure the LSP folks would welcome PRs with clarifications at https://github.com/microsoft/language-server-protocol

> I'm also assuming based on the discussion above that we're going to ignore CREATE type event since we're receiving workspace/didCreateFiles event with the required metadata?

That would be my assumption as well. The only edge case I would check however is how renames/moves are handled by the client - for example:

 - `/tmp/alpha/main.tf` is open
 - `/tmp/alpha/main.tf` is moved to `/tmp/beta/main.tf` & `/tmp/charlie/main.tf` is moved to `/tmp/alpha/main.tf`
 - assuming the rename is sent as part of the same single notification, this could either be 1 `Change` event or multiple `Create` and `Delete` events

btw. when the spec nor the VS Code client implementation provides us sufficient clarity, we consult the `gopls` codebase. A quick search through their codebase suggests they make similar assumptions: https://github.com/golang/tools/blob/e998cd2c59314075639579879898df1c61b57e2e/internal/lsp/cache/view.go#L441-L446

They also seem to implement a delay (default 100ms) and process notifications in batches, but I wouldn't implement it here _yet_ unless we have more context and real evidence it affects our users too.
https://github.com/golang/tools/blob/78ff15e68069f59866a005585edc7469f8671fc0/gopls/doc/settings.md#experimentalwatchedfiledelay-timeduration

internal/langserver/handlers/did_change_watched_files.go
```diff
@@ -0,0 +1,49 @@
...
+			// TODO: raise warning; unsupported file
+			return fmt.Errorf("Unable to update %s, filetype not supported.", filename)
+		}
+
+		// only handle Changed event type
+		if change.Type == protocol.Changed {
+			dh := ilsp.HandleFromDocumentURI(change.URI)
+
+			// check existence
+			_, err = svc.modStore.ModuleByPath(dh.Dir.Path())
```
danishprakash, line 34: Based on your comment on the issue page; we're essentially using this to ascertain whether a module associated with the given URI exists or not, right? And since we're not handling events other than Changed type, we're going to do an early return with an error here. Are these assumptions correct? 

https://github.com/hashicorp/terraform-ls/blob/70bb9e5545894cd629a16cf1d3e754fcd9c5d4c1/internal/state/module.go#L352-L354

internal/langserver/handlers/did_change_watched_files.go
```diff
@@ -0,0 +1,49 @@
...
+		URI := string(change.URI)
+		dir, err := os.Stat(URI)
+		if err != nil {
+			return fmt.Errorf("Unable to update %s. Failed to open directory, error: %w", URI, err)
+		}
+
+		filename := path.Base(URI)
+		if !dir.IsDir() && !ast.IsModuleFilename(filename) {
+			// TODO: raise warning; unsupported file
+			return fmt.Errorf("Unable to update %s, filetype not supported.", filename)
```
danishprakash, line 26: Since we want to show a warning/error notification, do you think we should return a custom error type and also emit a `window/showMessage` here?

internal/langserver/handlers/did_change_watched_files.go
```diff
@@ -0,0 +1,49 @@
...
+
+func (svc *service) DidChangeWatchedFiles(ctx context.Context, params lsp.DidChangeWatchedFilesParams) error {
+	for _, change := range params.Changes {
+		URI := string(change.URI)
+		dir, err := os.Stat(URI)
+		if err != nil {
+			return fmt.Errorf("Unable to update %s. Failed to open directory, error: %w", URI, err)
+		}
+
+		filename := path.Base(URI)
```
radeksimko, line 23:
```suggestion
		uri := string(change.URI)
		fi, err := os.Stat(uri)
		if err != nil {
			// TODO: raise warning via window/showMessage
			continue
		}
		if fi.IsDir() {
			// TODO: raise warning via window/showMessage
			continue
		}

		filename := path.Base(uri)
```
Usually in Go internal variables are lowercased/camelcased.

radeksimko (changes requested):
This is getting quite close. I left you some comments in-line.

Also feel free to checkout this PR branch if you need to test it: https://github.com/hashicorp/vscode-terraform/pull/1095

internal/langserver/handlers/did_change_watched_files.go
danishprakash: Thanks for the tip, I've been following [these](https://github.com/golang/go/wiki/CodeReviewComments#initialisms) guidelines from Go so far which states the following:
```
Words in names that are initialisms or acronyms (e.g. "URL" or "NATO")
have a consistent case. For example, "URL" should appear as "URL" or "url"
(as in "urlPony", or "URLPony"), never as "Url".
```

But I agree to your point on having that distinction for internal variables and otherwise. Thanks

internal/langserver/handlers/did_change_watched_files.go
```diff
@@ -0,0 +1,49 @@
...
+			if err != nil {
+				return err
+			}
+
+			jobIds, err := svc.parseAndDecodeModule(dh.Dir)
+			if err != nil {
+				return err
+			}
+
+			return svc.stateStore.JobStore.WaitForJobs(ctx, jobIds...)
```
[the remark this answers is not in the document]
danishprakash: Oh, thanks for pointing that out. It was an oversight, sorry about that.

internal/langserver/handlers/did_change_watched_files.go
radeksimko: > "URL" should appear as "URL" or "url" (as in "urlPony", or "URLPony"), never as "Url"

That seems right and not in conflict :) it's just that `URL` is more often used e.g. for constants or exported variables, `url` is what I'd use for internal variable.

danishprakash: I completely understand, I was just trying to understand the intent/reasoning. This actually makes more sense as I said and I'd personally be using this rule going forward. Thanks for taking the time to clear this up, really appreciate it :)

internal/langserver/handlers/did_change_watched_files.go
```diff
@@ -0,0 +1,79 @@
...
+		fi, err := os.Stat(uri)
+		if err != nil {
+			jrpc2.ServerFromContext(ctx).Notify(ctx, "window/showMessage", &lsp.ShowMessageParams{
+				Type:    lsp.Warning,
+				Message: fmt.Sprintf("Unable to update: %s, Failed to open directory", uri),
+			})
+			continue
+		}
+
+		if !fi.IsDir() {
```
danishprakash, line 31: Ok, so I pushed this change but I'm a little unsure of what's going on here. Can you corroborate the following? We're not handling single file changes here, although the spec says it can be part of the request, and we [_have_](https://github.com/hashicorp/terraform-ls/pull/843) preliminary support for the same (#843), and so we return with a warning if it's not a directory?

internal/langserver/handlers/did_change_watched_files.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+
+		// only handle `Changed` event type
+		if change.Type == protocol.Changed {
+			dh := ilsp.HandleFromDocumentURI(change.URI)
+
+			// check existence
+			_, err = svc.modStore.ModuleByPath(dh.Dir.Path())
+			if err != nil {
+				return err
```
danishprakash, line 55: I believe we should just warn and move ahead with other changes here as well? Similarly for the block below.

internal/langserver/handlers/did_change_watched_files.go
```diff
@@ -0,0 +1,79 @@
...
+			ids = append(ids, jobIds...)
+
+		}
+
+	}
+
+	// range over all ids corresponding
+	// to all changes in current request
+	for _, id := range ids {
+		err := svc.stateStore.JobStore.WaitForJobs(ctx, id)
```
danishprakash, line 72: Just curious, since we're "waiting" for jobs here, and given `go-memdb` is concurrent-safe for read ops, is it possible for us to do wait for jobs concurrently? Again, just a thought, I'm not sure how the internal semantics are just yet and I don't think there'll be any considerable difference in perceived performance/experience given we don't usually expect a lot of changes as part of a single request (or do we?).

internal/langserver/handlers/did_change_watched_files.go
radeksimko: I think this one we could just ignore/skip without warning as there can be genuinely valid reason for an error which is outside of user's control.

internal/langserver/handlers/did_change_watched_files.go
radeksimko: You can replace the for loop entirely with just the call, as the 2nd argument is variadic, i.e.

```go
err := svc.stateStore.JobStore.WaitForJobs(ctx, ids...)
if err != nil {
	return err
}
```

[It will loop over all IDs and wait for each](https://github.com/hashicorp/terraform-ls/blob/16c915b9fb13fb5061fc1a6c71959e3d9421af18/internal/state/jobs.go#L249-L256). There is no benefit in waiting concurrently since we're waiting for the slowest one anyway and the ones which are already done by the time we get to them will just return straight away.

internal/langserver/handlers/did_change_watched_files.go
radeksimko: I think the linked single file support won't affect the events we receive here. You are right however that we may receive file changes. IMO the best course of action would be simply using that path straight away as a module path, i.e. avoid calling `path.Base()`

internal/langserver/handlers/did_change_watched_files.go
danishprakash: Thanks for the clarification, waiting for the slowest one sounds right.

internal/langserver/handlers/did_change_watched_files.go
danishprakash: Agreed, the server should still process the other changes as part of the same request. But, should we not log it internally somewhere, skipping/ignoring it just means it's completely suppressed, right?

internal/langserver/handlers/did_change_watched_files.go
danishprakash: Got it, also doing away with the `IsDir()` check here now.

internal/langserver/handlers/did_change_watched_files.go
```diff
@@ -0,0 +1,71 @@
...
+	var ids job.IDs
+
+	for _, change := range params.Changes {
+		uri := string(change.URI)
+		_, err := os.Stat(uri)
+		if err != nil {
+			jrpc2.ServerFromContext(ctx).Notify(ctx, "window/showMessage", &lsp.ShowMessageParams{
+				Type:    lsp.Warning,
+				Message: fmt.Sprintf("Unable to update: %s, Failed to open directory", uri),
+			})
```
jpogran, line 26: I'm not sure we should be opening message windows for errors here. This pops for any change to any file/folder, so it could open many message windows at once. What is the user supposed to do if there is an error updating our cache? How will sending this error help the user?

Thinking about this, if we can't stat a directory we can't 'watch' it, therefore there is nothing for us to do. Likewise for a file. I would rather log here than send messages the user can't do anything about.

This also ties into the larger debate we have about returning errors from these handlers and how that affects client experience, but that isn't something you are required to deal with here

internal/langserver/handlers/did_change_watched_files.go
```diff
@@ -0,0 +1,71 @@
...
+		// `uri` can either be a file or a director baed on the spec.
+		// We're not making any assumptions on the above and passing
+		// the uri as the module path itself for validation.
+		//
+		// https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeWatchedFiles
+		if !ast.IsModuleFilename(uri) && !ast.IsVarsFilename(uri) {
+			jrpc2.ServerFromContext(ctx).Notify(ctx, "window/showMessage", &lsp.ShowMessageParams{
+				Type:    lsp.Warning,
+				Message: fmt.Sprintf("Unable to update file: %s, filetype not supported", uri),
+			})
```
jpogran, line 39: If this isn't a Terraform file, why is it an error that we should send to the user? Isn't this a case where we ignore the file if it's not a file we can do something with? As with the file access error above, I'd rather log here than interrupt the user with an error message they have no control over.

internal/langserver/handlers/did_change_watched_files.go
radeksimko: I think the main point here is that clients shouldn't be sending unsupported files, so this indicates client-side bug, which we'd like the user to report. We can make that clearer in the message though.

internal/langserver/handlers/did_change_watched_files.go
radeksimko: File access I'd be more inclined to ignore but the thing here is that clients shouldn't be sending unsupported files, so this clearly indicates client-side bug, which we'd like the user to report. We can make that clearer in the message though.

internal/langserver/handlers/did_change_watched_files.go
danishprakash: That's a very valid point. I was also thinking about how it can quickly become unwanted noise for the user. Logging makes sense here, especially since this is supposed to be an implicit request to the server unlike other events such as "go to definition", in which case, telling the user that the particular file isn't supported makes a lot more sense.

internal/langserver/handlers/did_change_watched_files.go
danishprakash: @radeksimko do you think this can be solved if and when we have the clients have the ability to specify file patterns to watch for? Not maybe for this change right away but longer term. I'm not sure but I feel aligned with what @jpogran mentioned above. Having to show errors to the user and expect them to report them to the client seems counterintuitive. What do you think?

danishprakash:
@radeksimko sure, but if it's okay with you, I'd love to know what all is pending and required If I were to complete this PR? I know tests are one. Apart from that, I'd really like to understand the complications if there are any. Thanks :)

radeksimko:
Also after testing it more I just realized I was wrong about the need of checking just `Changed` - we'd also need to account for deletions and creations of files within existing (tracked) directories - I'll address that.

radeksimko:
This turns out to be more complex than I initially anticipated, but I believe my last commit finally accounts for the remaining cases (deletions and creations).

@danishprakash let me know if you have any feedback or questions.

radeksimko (approved):

danishprakash:
@radeksimko hey, I have a question about the recent updates:
- We initially decided not to handle created and deleted events with the assumption that `didOpen` is triggered at least for the former. Were you able to reproduce a scenario wherein this was not the case? What did you test?

internal/langserver/handlers/did_change_watched_files.go
```diff
@@ -0,0 +1,181 @@
...
+			if err != nil {
+				svc.logger.Printf("error finding module (%q deleted): %s", parentDir, err)
+				continue
+			}
+
+			// and check the parent directory still exists
+			fi, err := os.Stat(parentDir)
+			if err != nil {
+				if os.IsNotExist(err) {
+					// if not, we remove the indexed module
```
danishprakash, line 51: When can this happen? The user deleted the only file and then subsequently the directory, too? If yes, then those would also be different events here and that's probably why you're just ensuring the directory is still present or not.

internal/langserver/handlers/did_change_watched_files.go
```diff
@@ -0,0 +1,181 @@
...
+					svc.removeIndexedModule(ctx, rawURI)
+					continue
+				}
+				svc.logger.Printf("error checking existence (%q deleted): %s", parentDir, err)
+				continue
+			}
+			if !fi.IsDir() {
```
danishprakash, line 58: Curious as to why this check is here since we're doing this earlier:
```
parentDir := filepath.Dir(rawPath)
```

radeksimko:
> We initially decided not to handle created and deleted events with the assumption that didOpen is triggered at least for the former. Were you able to reproduce a scenario wherein this was not the case? What did you test?

So `didOpen` is triggered for any _open_ files. What I didn't realize initially was that we actually use the parsed AST and other related metadata of unopened modules e.g. for completion within `module` block, which will be even more important when https://github.com/hashicorp/terraform-ls/issues/724 is addressed.

internal/langserver/handlers/did_change_watched_files.go
radeksimko: This accounts for the deletion of a whole directory.

> If yes, then those would also be different events here

Frankly, I don't know, since LSP doesn't go into detail about how clients should or shouldn't group file events. In fact when I tried to `rm -rf submodule` in VS Code where submodule had some `*.tf` files, no events were fired at all - which may be a bug in the VS Code language client - unsure.

So in the face of uncertainty I'm just being a little more defensive here.

internal/langserver/handlers/did_change_watched_files.go
radeksimko: Right, this should basically never happen - I was just trying to be a bit more defensive here too.

The only realistic scenarios which come to mind involve cases where it's not a file nor a directory, but e.g. symlink or any special type of file - which is mostly relevant on Unix-based systems:
https://pkg.go.dev/io/fs#FileMode

danishprakash: Okay, so just yesterday, I got bit by a severe production bug wherein I was copying files from one directory to another. Without a safety check of whether the file is a regular file or not, we ended up doing a cp on a file which was a UNIX pipe and the application just froze indefinitely waiting for the copy to finish would _never_ be the case. I can totally understand the need for this check here howsoever obvious it might seem.

internal/langserver/handlers/did_change_watched_files.go
danishprakash: Got it, thanks for the explanation and yes, it's still a little nebulous when it comes to understanding client-side event emits.

> may be a bug in the VS Code language client
Do you mean https://github.com/Microsoft/vscode-languageserver-node?

radeksimko: It could be either VS Code itself or the language client you linked - it would take some debugging to find out.

jpogran (approved):
I'm still nervous about opening so many message windows that they can't silence, for things the user can't do anything about, but this all works for the cases I could think of so I don't have any objections to merging this. Thank you both for this work!

PR #798, opened 2022-02-22
goreleaser: Enable pre-release publishing

### LINKED ISSUE #456: Avoid publishing pre-releases to Homebrew tap
Currently we have the ability to cut pre-releases, such as `0.12.0-beta1` or `0.12.0-rc1`. However such releases would likely make their way into the Homebrew tap as well, which is considered stable.

In order to streamline consumption of pre-releases for interested parties who opt-in it would be useful to investigate how to prevent publishing of such releases to Homebrew.

One possible way might be templated `skip_upload` under the `brews` item:
https://goreleaser.com/customization/homebrew/

Also we have never actually attempted to cut any prerelease, so this needs to be checked too.

Related: https://github.com/hashicorp/vscode-terraform/issues/605

radeksimko:
TODO:
 - [x] Verify impact on https://github.com/hashicorp/vscode-terraform (earlier versions which do not bundle LS)
 - [ ] Docs to cover expectations of pre-releases and how to obtain them
 - [x] Linux packages - do other product publish them for pre-releases? Do we need to do anything extra?

--- 

Closes #456

See relevant docs at https://goreleaser.com/customization/release/#github

```yaml
  # If set to auto, will mark the release as not ready for production
  # in case there is an indicator for this in the tag e.g. v1.0.0-rc1
  # If set to true, will mark the release as not ready for production.
  # Default is false.
  prerelease: auto
```

and https://goreleaser.com/customization/homebrew/

```yaml
    # If set to auto, the release will not be uploaded to the homebrew tap
    # in case there is an indicator for prerelease in the tag e.g. v1.0.0-rc1
    # Default is false.
    skip_upload: true
```

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -94,6 +94,7 @@ brews:
       owner: hashicorp
       name: homebrew-tap
       token: "{{ .Env.HOMEBREW_TAP_TOKEN }}"
+    skip_upload: auto
     url_template: "https://releases.hashicorp.com/terraform-ls/{{ .Version }}/{{ .ArtifactName }}"
     commit_author:
       name: hc-espd-packager
@@ -123,6 +124,9 @@ publishers:
 milestones:
   - name_template: "{{ .Tag }}"
     close: true
+
+release:
+  prerelease: auto
 
 changelog:
   skip: true


jpogran:
We specify `latest` by default using https://github.com/hashicorp/vscode-terraform/blob/ddddaf39c5970266a2d5ea21c75e573e31cb5152/src/installer/updater.ts#L36. `DEFAULT_LS_VERSION` is defined at https://github.com/hashicorp/vscode-terraform/blob/v2.19.0/src/installer/detector.ts#L6. This means the default action is to download only the latest stable version.

I verified this by setting up a test repo with the methods we use and pointed it to the terraform project which produces pre-release versions.

A user would have to manually specify a pre-release version using the deprecated `terraform.languageServer.requiredVersion` to purposely download a pre-release version.

jpogran (approved):

radeksimko:
I will cover docs update in a separate PR and I confirmed that Linux packaging for pre-releases is handled and we don't need to do anything extra for that.

PR #802, opened 2022-02-23
Update semantic token block label type

### LINKED ISSUE #777: Token identification for resource block label
In researching https://github.com/hashicorp/vscode-terraform/issues/710 and concluded it was a bug and not a feature request because we have both an `entity.name.type.terraform` and an `entity.name.label.terraform` matcher. I modified the matcher with something good enough (but not production worthy yet) to get it to match and the semantic tokens coming from the LS override it with a string (as shown).

image

@radeksimko pointed out this is coming from: https://github.com/hashicorp/terraform-ls/blob/2bdb4197f5c10431461c1cfb468ca13ebd0b97cf/internal/lsp/token_encoder.go#L36-L37

https://github.com/hashicorp/terraform-ls/blob/2bdb4197f5c10431461c1cfb468ca13ebd0b97cf/internal/lsp/token_types.go#L63-L117

It seems like VS Code does support a `label` type: https://github.com/microsoft/vscode/blob/de9ab35ce9da7cb3c850a1b570876263a472174b/src/vs/platform/theme/common/tokenClassificationRegistry.ts#L570

We need to research if using a label type will be possible, or if that conflicts with other parts of the grammar.

dbanck:
This PR updates the type for block labels from string to enum member.

It's the first step in providing more contextual information on block labels. With a schema update, it will be possible to differentiate between multiple block labels. This will allow us to report two different tokens for block labels where the first one stays enum member and the second one becomes a string.

We've tried different token types, like `class`, `macro`, `method`, and `enumMember` and settled on the latter. `enumMember` makes the most sense from a semantic perspective and has good highlighting in the default themes.

Having semantic tokens for block labels enables better highlighting for nested blocks.

Before (no LS)
CleanShot 2022-02-23 at 15 23 06

After (with LS)
CleanShot 2022-02-23 at 15 26 55

For a smoother user experience, we propose to update the static grammar for block labels from `entity.name.tag.terraform` to `variable.other.enummember`. This change would update the raw grammar version (no LS) to this:
CleanShot 2022-02-23 at 15 28 43

Closes #777

### DIFF
--- a/internal/langserver/handlers/semantic_tokens_test.go
+++ b/internal/langserver/handlers/semantic_tokens_test.go
@@ -71,7 +71,8 @@ func TestSemanticTokensFull(t *testing.T) {
 					"tokenTypes": [
 						"type",
 						"property",
-						"string"
+						"string",
+						"enumMember"
 					],
 					"tokenModifiers": [
 						"deprecated",
@@ -113,7 +114,7 @@ func TestSemanticTokensFull(t *testing.T) {
 			"result": {
 				"data": [
 					0,0,8,0,0,
-					0,9,6,1,2
+					0,9,6,3,2
 				]
 			}
 		}`)
@@ -178,7 +179,8 @@ func TestSemanticTokensFull_clientSupportsDelta(t *testing.T) {
 					"tokenTypes": [
 						"type",
 						"property",
-						"string"
+						"string",
+						"enumMember"
 					],
 					"tokenModifiers": [
 						"deprecated",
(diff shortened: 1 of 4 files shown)


jpogran:
Depending on your theme, the 'no LS' does highlight differently than strings.

Dark+ (default): 
image

Light+:
image

I do agree on aligning the scopes between the textmate grammar and the semantic token, just pointing out it does do something :)

jpogran (approved):

PR #807, opened 2022-02-25
Update help for `serve` command arguments

dbanck:
This PR updates the help section for the `serve` command.

Variables in arguments must be lower-case and without a `.`, e.g.
```
terraform-ls serve -cpuprofile='/Users/dbanck/tmp/profile-{{timestamp}}.pprof'`
```


radeksimko (approved):
Thanks, confirmed from https://github.com/hashicorp/terraform-ls/blob/7c44da68086bde8b24d052f8497200e7a326990a/internal/pathtpl/pathtpl.go#L19-L23

We could go further in clarifying that these are in fact templating functions, rather than variables (variables do use the `.CamelCase` syntax, functions are without the dot and typically lowercase).

That said don't treat it as blocker - this is already improving things significantly! 😄

radeksimko:
Totally fair - merge at will! 👍🏻

PR #817, opened 2022-03-02
internal/lsp: report dependent modifiers as `defaultLibrary`

radeksimko:
Closes #710

The (LSP) modifier `defaultLibrary` better represents the meaning of dependent labels or attributes, as opposed to `modification` which was chosen mostly arbitrarily as part of the initial implementation in https://github.com/hashicorp/terraform-ls/pull/331

Highlighting dependent labels was already possible prior to this patch but the change of the modifier here reflects the commitment to support custom themes in a more sensible (and hopefully stable) way.

Default VSCode themes [do not assign meaning to any modifiers](https://github.com/microsoft/vscode/blob/main/src/vs/platform/theme/common/tokenClassificationRegistry.ts#L572-L594) but it is possible to take advantage of this in a custom theme or simply by modifying the VSCode settings such as

```json
    "editor.semanticTokenColorCustomizations": {
        "rules": {
            "enumMember.defaultLibrary": {
                "bold": true
            }
        }
    },
```

## Before

(assuming customized settings as above & default theme)

Screenshot 2022-03-02 at 12 20 41 Screenshot 2022-03-02 at 12 20 31

## After

(assuming customized settings as above & default theme)

Screenshot 2022-03-02 at 12 16 20 Screenshot 2022-03-02 at 12 16 37

cc @tringuyen-yw

--- 

`variable` name getting also reported as dependent is not intended, but I opened a separate issue for that problem: https://github.com/hashicorp/terraform-schema/issues/96

### DIFF
--- a/internal/langserver/handlers/semantic_tokens_test.go
+++ b/internal/langserver/handlers/semantic_tokens_test.go
@@ -76,7 +76,7 @@ func TestSemanticTokensFull(t *testing.T) {
 					],
 					"tokenModifiers": [
 						"deprecated",
-						"modification"
+						"defaultLibrary"
 					],
 					"requests": {
 						"full": true
@@ -184,7 +184,7 @@ func TestSemanticTokensFull_clientSupportsDelta(t *testing.T) {
 					],
 					"tokenModifiers": [
 						"deprecated",
-						"modification"
+						"defaultLibrary"
 					],
 					"requests": {
 						"full": {
@@ -293,7 +293,7 @@ func TestVarsSemanticTokensFull(t *testing.T) {
 					],
 					"tokenModifiers": [
 						"deprecated",
-						"modification"
+						"defaultLibrary"
 					],
 					"requests": {
 						"full": true
--- a/internal/lsp/token_encoder.go
+++ b/internal/lsp/token_encoder.go
@@ -69,10 +69,10 @@ func (te *TokenEncoder) encodeTokenOfIndex(i int) []uint32 {
 	for _, m := range token.Modifiers {
 		switch m {
 		case lang.TokenModifierDependent:
-			if !te.tokenModifierSupported(TokenModifierModification) {
+			if !te.tokenModifierSupported(TokenModifierDefaultLibrary) {
 				continue
 			}
-			modifiers = append(modifiers, TokenModifierModification)
+			modifiers = append(modifiers, TokenModifierDefaultLibrary)
 		case lang.TokenModifierDeprecated:
 			if !te.tokenModifierSupported(TokenModifierDeprecated) {
 				continue
--- a/internal/lsp/token_types.go
+++ b/internal/lsp/token_types.go
@@ -113,7 +113,7 @@ var (
 	}
 	serverTokenModifiers = TokenModifiers{
 		TokenModifierDeprecated,
-		TokenModifierModification,
+		TokenModifierDefaultLibrary,
 	}
 )
 


jpogran:
Can we contribute this inside the extension as shown in https://code.visualstudio.com/api/language-extensions/semantic-highlight-guide#custom-textmate-scope-mappings?

tringuyen-yw:
Although bold is better than nothing. In my opinion, it would be even better if the user can define the font attribute directly. A possible example of this kind of color syntax is possible with Intelli-J. See screenshot in this issue [vscode-terraform, 710, Color syntax: add scope for resource_type and resource_name](https://github.com/hashicorp/vscode-terraform/issues/710)

In the case of VSCode, I can edit `settings.json` and add/override the `textMateRules` on the scope specific to the languange. Here is an example for Terraform HCL (excerpt from my `settings.json`). The links to howto customize the VSCode theme colors are given in the above Github issue.

```json
"editor.tokenColorCustomizations": {
  "textMateRules": [
    {
      // Terraform keywords
      "scope": [
        "entity.name.type.terraform",
      ],
      "settings": {
        "foreground": "#8040FF", "fontStyle": ""
      }
    },
    {
      // name of nested block within a resource block
      "scope": [
        "entity.name.label.terraform",
      ],
      "settings": {
        "foreground": "#007CCC", "fontStyle": ""
      }
    },
  ]
}
``` 

Base on the example above, may be you can add some additional scopes to correspond to Terraform's resource-type and resource-id?

I don't think it is necessary to worry about the custom themes. For themes which are not aware about Terraform specific scopes, they would just theme the resource-type and resource-id as string. For those which can or for users who are willing to edit their `settings.json` manually, they would see the proper syntax color.

radeksimko:
> Although bold is better than nothing. In my opinion, it would be even better if the user can define the font attribute directly.

The bold was just an example. As you can see from the attached code in the PR - it does not set any particular style, just "meaning". The main point here is exactly to allow users to define any styling they want, as shown in the snippet of config above - which just used bold.

I (mostly blindly) assumed that custom themes can simply assign their own styles to token types and modifiers but regardless we can certainly map these semantic tokens to some TextMate grammar scopes to make that possible or easier.

We will keep `entity.name.type.terraform` or `entity.name.type.terraform` in mind when introducing the mapping in the extension. Generally the two _would_ make sense from the semantics perspective but we also need to ensure the config is reasonably highlighted in the default themes - which was a challenge in the past. Either way - that sounds like a problem to discuss deal with and discuss on the extension side.

This PR here merely enables the mapping, nothing more (yet).

> Base on the example above, may be you can add some additional scopes to correspond to Terraform's resource-type and resource-id?

I'm afraid we are at the mercy of LSP here as we can only work with these language-agnostic types and modifiers here https://github.com/hashicorp/terraform-ls/blob/bff4e90811f3375c029cf099142adbf869c56f5c/internal/lsp/token_types.go#L64-L98

The protocol doesn't seem to allow just arbitrary custom token types at this point and we struggled to find the right mapping to HCL constructs.

However I'm assuming that you could work with `terraform` in the scope name & a scope we eventually pick for dependent labels.

Did you want to style `resource` labels differently from `data` labels? If so, then that I'm afraid is not a use case we can meet through semantic-token based highlighting due to the above limitations. We _could_ in theory introduce some custom scopes to a dedicated static Terraform grammar however.

radeksimko:
Actually there is this https://code.visualstudio.com/api/language-extensions/semantic-highlight-guide#custom-token-types-and-modifiers which suggests that custom token types are allowed, but it's unclear to me what implications does that have on the server. Unlike with TM scopes there's only one token type that we can assign to the same range, so this would suggest that any client that doesn't support that custom token type is just out of luck? 🤔

radeksimko:
@tringuyen-yw I believe we have a plan forward with regards to supporting custom themes via custom token types, see https://github.com/hashicorp/vscode-terraform/pull/958 with more details.

In the meantime though I'd still like to go through with this PR. The only difference after #958 is implemented fully (on LS side) is that `defaultLibrary` becomes a fallback for e.g. `hcl-dependent` modifier depending on what client capabilities are received.

cc @hashicorp/tf-editor-experience for review

dbanck (approved):
LGTM 👍

PR #818, opened 2022-03-03
internal/state: avoid deadlock caused by mid-flight job change

radeksimko:
There is a relatively rare case of when we may end up processing the same job twice here:

https://github.com/hashicorp/terraform-ls/blob/bff4e90811f3375c029cf099142adbf869c56f5c/internal/state/jobs.go#L188-L211

1. a job e.g. ID 42 is found by e.g. "closed scheduler" (`*`)
2. in the split second between finding this job and marking it as running (which itself _is_ correctly guarded by an internal memdb mutex), the involved directory is open.
3. the same job ID 42 is then also found by "open scheduler" (`**`)
4. whichever scheduler obtains the memdb mutex first marks that job as running
5. the other scheduler then waits for the mutex and attempts to do the same
6. because that fails, we return error, making the scheduler to effectively shut down, practically causing a deadlock in places which wait for jobs to finish - as they'd never finish

(`*`) - the scheduler responsible for processing jobs involving directories which are _not_ open
(`**`) - the scheduler responsible for processing jobs involving directories which _are_ open

While this may be rare, I ran into that bug myself this morning, as can be seen from this log snippet:
https://gist.github.com/radeksimko/7ce4e9a7d0c8bbc65bd65e277dc3bd92

### DIFF
--- a/internal/scheduler/scheduler_test.go
+++ b/internal/scheduler/scheduler_test.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"sync"
 	"sync/atomic"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


dbanck (approved):
Nice catch and excellent test case!

PR #820, opened 2022-03-04
github: bump dessant/lock-threads to v3

radeksimko:
This is in reaction to v2 apparently not being available anymore:

https://github.com/hashicorp/terraform-ls/runs/5419688427?check_suite_focus=true

follows the upgrade guide in https://github.com/dessant/lock-threads/commit/08e671be8ac8944d0e132aa71d0ae8ccfb347675

### DIFF
--- a/.github/workflows/lock.yml
+++ b/.github/workflows/lock.yml
@@ -1,5 +1,9 @@
 name: 'Lock Threads'
 
+permissions:
+  issues: write
+  pull-requests: write
+
 on:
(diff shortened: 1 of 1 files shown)


radeksimko:
Actually the error just indicates permissions problem as we recently changed default permissions for all Actions in this repo, not availability of the Action.

I'll deal with this slightly differently - let dependabot also manage GHA updates and add the relevant explicit permissions for the token.

PR #821, opened 2022-03-04
github: declare permissions for all GitHub Actions

radeksimko:
This allows us to be a little bit more defensive by not giving any workflows more permissions than they really need and use this safer default settings for the repo

Screenshot 2022-03-04 at 09 16 36

I also removed the explicit token for lock-threads as this is now default since 2.1.0 https://github.com/dessant/lock-threads/blob/master/CHANGELOG.md#210-2021-07-09 and did the same for the stale action: https://github.com/actions/stale#repo-token

More about GitHub token permissions: https://docs.github.com/en/actions/security-guides/automatic-token-authentication

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,9 +8,6 @@ on:
     branches:
       - main
 
-permissions:
-  contents: read
-
 env:
   GOPROXY: https://proxy.golang.org/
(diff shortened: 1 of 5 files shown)


dbanck (approved):
Good job on further locking down permissions

PR #827, opened 2022-03-04
internal/lsp: recognize type declarations as semantic tokens

### LINKED ISSUE #816: Semantic token highlighting for variable type declarations
### Use-cases

Improve accuracy of highlighting via semantic tokens for type declarations via semantic tokens.

```hcl
variable "test" {
  type = object({
    test_object = map(
      object({
        test_map = string
      })
    )
  })
}
```

### Proposal

Type declarations already seem to be reported by `hcl-lang` as `TokenTypeCapsule` and `TokenTypePrimitive`.

The goal is to choose appropriate LSP token types to translate these into.

https://github.com/hashicorp/terraform-ls/blob/10853c52f8728520095f9dfaa57abf3b85e2be65/internal/lsp/token_encoder.go#L33-L57

radeksimko:
Closes #816

## Before

Screenshot 2022-03-04 at 15 35 58 Screenshot 2022-03-04 at 15 36 10

## After

Screenshot 2022-03-04 at 15 36 31 Screenshot 2022-03-04 at 15 36 44

--- 

## Related Thoughts

It is possible that we may want to replace the two tokens with something even more accurate, like `TokenTypeString`, `TokenTypeNumber`, `TokenTypeSet`, `TokenTypeList` etc. but it is currently unclear whether anyone would actually benefit from that. We can always revisit supported tokens later.

We can also see that the VSCode grammar highlights types as variables. I would say that keyword scope is more appropriate, as that leaves variable scope for HCL traversals and allows us to visually distinguish the two. i.e. We can adjust the highlighting of primitive types like `string`, `number`, `null` or `bool` in the grammar as well.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/hashicorp/go-uuid v1.0.2
 	github.com/hashicorp/go-version v1.4.0
 	github.com/hashicorp/hc-install v0.3.1
-	github.com/hashicorp/hcl-lang v0.0.0-20211123142056-191cd51dec5b
+	github.com/hashicorp/hcl-lang v0.0.0-20220304112739-b2b314ea7c44
 	github.com/hashicorp/hcl/v2 v2.11.1
 	github.com/hashicorp/terraform-exec v0.16.0
 	github.com/hashicorp/terraform-json v0.13.0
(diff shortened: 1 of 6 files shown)


dbanck (approved):
LGTM 👍

Issue #646, opened 2021-09-22
Missing references for blocks in a separate config file


radeksimko:
### Server Version

```
0.22.0
```

### Terraform Configuration Files

**`variables.tf`**
```hcl
variable "aaa" {}
variable "bbb" {}
variable "ccc" {}
```

**`main.tf`**
```hcl
output "name" {
  value = var.
}
```

### Expected Behavior

References to all three variables (aaa, bbb, ccc) to be suggested on completion in `main.tf`.

### Actual Behavior

Missing `var.bbb` reference.
![2021-09-22 10 02 22](https://user-images.githubusercontent.com/287584/134314902-92410650-c369-4fff-a848-b412c8b60d40.gif)

### Steps to Reproduce

Failing E2E test:

https://github.com/hashicorp/terraform-ls/commit/6679f0e07d826213625cbfaa71e1df7c83137f23

radeksimko:
This is still reproducible in the latest `main` revision, same (rebased) failing test: https://github.com/hashicorp/terraform-ls/commit/845f16f

PR #829, opened 2022-03-14
fix: missing references for some blocks in a separate config file

radeksimko:
This brings in https://github.com/hashicorp/hcl-lang/pull/105 which also describes the root cause of the problem. I also attached a test to ensure this doesn't reoccur in the future, although the fix is already tested upstream.

Fixes https://github.com/hashicorp/terraform-ls/issues/646


dbanck (approved):
👍 Thanks for including the test case here as well

PR #837, opened 2022-03-21
docs: Document architecture of the LS

radeksimko:


jpogran (approved):
🚀 Thank you for the excellent documentation! 🚀 This helps a lot in understanding how things are working behind the scenes and I appreciate all the diagrams and the work it took to make that easy to read

danishprakash:
Just dropping by to say the same as above, the documentation is great. Helps a lot in explaining some of the flows I personally had trouble understanding all this while. Thanks!

PR #840, opened 2022-03-23
Job scheduler benchmarks

### LINKED ISSUE #834: Chore: Publish modules benchmarking results from job scheduler refactor
Think about how we can automate the monitoring of a team SLA for performance.

CodeLens command good first test case.

Performance
- Line in the sand
	- Where are we now?
	- Where would we like to be?
- What is a normal config?
- What is a top end config?
- Create a public affirmation of performance guidelines SLA
- We have existing benchmarks via test suite.
- Cold vs Hot load
- Existing framework tests against stable, in development how do we test against latest.

## Proposal

 - assemble a hand-crafted list of modules from the Registry
   - `hashicorp/terraform-aws-consul`
   - `hashicorp/terraform-aws-terraform-enterprise`
   - https://registry.terraform.io/modules/terraform-aws-modules/vpc/aws/latest
 - run `initialize` / `initialized` to trigger the directory hierarchy walk & parsing & discovery as per https://gist.github.com/radeksimko/aae53d1bb93e96dd524793113718da3f
 - report total memory allocated (e.g. in bytes)
 - report total CPU time spent (in seconds)
 - set a benchmark for both memory usage and CPU time and document it in `/docs` publicly and/or link from `Readme.md`
 - run nightly in GHA (input can be branch name)
 - report violations of benchmark for either CPU or memory nightly into Slack
 - (nice to have) render & store graphs of CPU / memory usage

radeksimko:
Fixes #834 

--- 

## New Benchmarks

This adds 12 new benchmarks into the `handler` package, to measure the performance of the hot path - i.e. walking/indexing for various different setups.

I created a few local modules with varying levels of complexity and different providers + picked some popular modules from the Registry using a similar pattern. All of them are described in the docs along with estimated average CPU time and memory consumption.

## Evaluating Benchmarks

Unfortunately there is no easy way of processing the benchmark outputs as per https://github.com/golang/go/issues/40588

I found this GitHub Action on the Marketplace: https://github.com/marketplace/actions/continuous-benchmarking-for-go but ended up only using the underlying [gobenchdata](https://github.com/bobheadxi/gobenchdata) to parse & evaluate the benchmark data, not the graphing part - primarily because it would require us giving the Action push access to the repository (which AFAIK cannot be limited to a particular branch even).

The charts would be lovely to have but not at these costs IMO.

## Slack Notifications

The [official Slack GitHub Action](https://github.com/marketplace/actions/slack-send) provides a few different integrations and I just picked the easiest one since creating a new App is far more involved. This results in a very brief message on failure with a link to the logs, such as
Screenshot 2022-04-07 at 10 07 45

We could potentially explore better integration which would also allow us to post details about thresholds breached etc. straight into Slack, but I thought this is good enough to start with.


jpogran (approved):
Shiny ✨

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,90 @@
...
+          RESULTS_PATH="${{ runner.temp }}/benchdata-results.json"
+          CHECKS_CONFIG_PATH="${{ github.workspace }}/.github/gobenchdata-checks.yml"
+
+          cat $BENCH_PATH | $BENCHDATA --json ${DATA_PATH} -v "${GITHUB_SHA}" -t "ref=${GITHUB_REF}"
+
+          $BENCHDATA checks eval \
+            ${DATA_PATH} \
+            ${DATA_PATH} \
+            --checks.config ${CHECKS_CONFIG_PATH} \
+            2> $RESULTS_PATH # eval returns result JSON to stderr for some reason
```
bobheadxi, line 62: Ah this command is intended to be used with the `--json [path]` flag, and probably should not print JSON unless that flag is provided - I'm fixing this with https://github.com/bobheadxi/gobenchdata/pull/53 ! (though that would be a breaking change here)

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,90 @@
...
+
+          cat $RESULTS_PATH | jq -e '.Status == "pass"'
+      -
+        name: Report failures
+        id: report
+        if: ${{ failure() && !cancelled() }}
+        run: |
+          RESULTS_PATH="${{ runner.temp }}/benchdata-results.json"
+          FAILURES=`cat ${RESULTS_PATH} | jq '.Checks[] | select(.Status == "fail")'`
+          echo $FAILURES
```
bobheadxi, line 72: There is a report command for CI, `gobenchdata checks report $RESULTS_PATH`, which will print a markdown table of your benchmark results and exit with a non-zero status

bobheadxi (commented):
Hey, hope it's alright if I swing by to leave some comments on the `gobenchdata` usage - super excited to see it in use! 😁 

> not the graphing part - primarily because it would require us giving the Action push access to the repository (which AFAIK cannot be limited to a particular branch even).

This can be done separately from the action by generating the visualization app separately: https://github.com/bobheadxi/gobenchdata#visualisation

tl;dr in a new repository:

```
gobenchdata web generate --web.config-only .
```

Then, in your GitHub action here, you can basically re-implement the script used in the published action:

```sh

# fetch published data
echo
echo "📚 Checking out ${INPUT_PUBLISH_REPO}@${INPUT_PUBLISH_BRANCH}..."
cd /tmp/build
git clone https://${GITHUB_ACTOR}:${GITHUB_TOKEN}@github.com/${INPUT_PUBLISH_REPO}.git .
git checkout ${INPUT_PUBLISH_BRANCH}
echo

  # merge results with published
  echo '☝️ Updating results...'
  if [[ -f "${INPUT_BENCHMARKS_OUT}" ]]; then
    echo '📈 Existing report found - merging...'
    gobenchdata merge "${RUN_OUTPUT}" "${INPUT_BENCHMARKS_OUT}" \
      --prune "${INPUT_PRUNE_COUNT}" \
      --json "${INPUT_BENCHMARKS_OUT}" \
      --flat
  else
    cp "${RUN_OUTPUT}" "${INPUT_BENCHMARKS_OUT}"
  fi

  # publish results
  echo
  echo '📷 Committing and pushing new benchmark data...'
  git add .
  git commit -m "${INPUT_GIT_COMMIT_MESSAGE}"
  git push -f origin ${INPUT_PUBLISH_BRANCH}
```

Would be happy to help you get the most out of the tool!

.github/workflows/benchmarks.yml
radeksimko: Ah, that's nice - I'll take a look into it! Thanks for the note.

radeksimko:
> This can be done separately from the action by generating the visualization app separately: https://github.com/bobheadxi/gobenchdata#visualisation and setting up a separate repository dedicated to hosting benchmark data (which should be low-risk enough for this use case hopefully!)

I recall having considered this option too, and it would certainly be better than pushing to the same repo, assuming we also leverage [deploy keys](https://docs.github.com/en/developers/overview/managing-deploy-keys#deploy-keys) and probably somehow connect the two repos so that it rebuilds on every run. Having it self-contained in one repo safely is however still more appealing.

To be clear the main problem here is GitHub's own inflexible permission model which doesn't allow tokens to be scoped to any particular repositories or branches, I don't see how gobenchdata can do any better here. 🤔  😄 

We can revisit this later as we have plans to publish docs to a website at some point too, so we'll have to build some kind of solution/infrastructure for that anyway.

This PR intended to just put something relatively quick & simple ready.

PR #842, opened 2022-03-28
Onboard to RelAPI

jeanneryan:
Onboard project to the Releases API.

***Note: RelEng  will merge this PR when ready***

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -52,19 +52,10 @@ jobs:
         name: Install signore
         uses: hashicorp/setup-signore-package@v1
       -
-        uses: hashicorp/setup-hc-releases@v1
+        name: Setup hc-releases
+        uses: hashicorp/actions-setup-hc-releases@v1
         with:
           github-token: ${{ secrets.CODESIGN_GITHUB_TOKEN }}
-      -
-        name: Configure AWS Credentials
-        uses: aws-actions/configure-aws-credentials@v1
-        with:
-          aws-access-key-id: ${{ secrets.TERRAFORM_PROD_AWS_ACCESS_KEY_ID }}
-          aws-secret-access-key: ${{ secrets.TERRAFORM_PROD_AWS_SECRET_ACCESS_KEY }}
-          aws-region: us-east-1
-          role-to-assume: ${{ secrets.TERRAFORM_PROD_AWS_ROLE_TO_ASSUME }}
-          role-skip-session-tagging: true
-          role-duration-seconds: 3600
       -
         name: Release
         uses: goreleaser/goreleaser-action@v2
@@ -72,6 +63,8 @@ jobs:
           version: latest
           args: release
         env:
+          HC_RELEASES_HOST_STAGING: ${{ secrets.HC_RELEASES_HOST_STAGING }}
+          HC_RELEASES_KEY_STAGING: ${{ secrets.HC_RELEASES_KEY_STAGING }}
           CODESIGN_IMAGE: ${{ steps.codesign.outputs.image }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           HOMEBREW_TAP_TOKEN: ${{ secrets.HOMEBREW_TAP_GITHUB_TOKEN }}
@@ -82,7 +75,20 @@ jobs:
           SIGNORE_CLIENT_SECRET: ${{ secrets.SIGNORE_CLIENT_SECRET }}
           SIGNORE_SIGNER: ${{ secrets.SIGNORE_SIGNER }}
       -
-        name: Publish released artifacts
-        run: hc-releases publish -product=terraform-ls
-        env:
-          FASTLY_API_TOKEN: ${{ secrets.RELEASE_FASTLY_API_TOKEN }}
+        name: Create Release metadata
+        uses: hashicorp/actions-hc-releases-create-metadata@main
+        with:
+          private-tools-token: ${{ secrets.CODESIGN_GITHUB_TOKEN }}
+          product-name: "terraform-ls"
+          version: ${{ github.ref_name }}
+          hc-releases-host: ${{ secrets.HC_RELEASES_HOST_STAGING}}
+          hc-releases-key: ${{ secrets.HC_RELEASES_KEY_STAGING }}
+      -
+        name: Promote
+        uses: hashicorp/actions-hc-releases-promote@main
+        with:
+          product-name: "terraform-ls"
+          version: ${{ github.ref_name }}
+          hc-releases-host: ${{ secrets.HC_RELEASES_HOST_PROD }}
+          hc-releases-key: ${{ secrets.HC_RELEASES_KEY_PROD }}
+          hc-releases-source_env_key: ${{ secrets.HC_RELEASES_KEY_STAGING }}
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -52,26 +52,19 @@ jobs:
         name: Install signore
         uses: hashicorp/setup-signore-package@v1
       -
-        uses: hashicorp/setup-hc-releases@v1
+        name: Setup hc-releases
+        uses: hashicorp/actions-setup-hc-releases@v1
```
mdeggies, line 56: PR is looking great!! Awesome job. Just one note- This will need to point to the `v2` tag, which will grab our launch version (0.1.0) of the hc-releases CLI -- being published tmr!

.goreleaser.yml
```diff
@@ -110,20 +110,18 @@ brews:
...
-    cmd: hc-releases upload-file {{ abs .ArtifactPath }}
-
 milestones:
   - name_template: "{{ .Tag }}"
     close: true
+      - HC_RELEASES_HOST={{ .Env.HC_RELEASES_HOST_STAGING }}
+      - HC_RELEASES_KEY={{ .Env.HC_RELEASES_KEY_STAGING }}
+    dir: "{{ dir .ArtifactPath }}"
+    cmd: |
+      hc-releases upload -file={{ .ArtifactName }} -product={{ .ProjectName }} -version={{ .Version }}
```
dbanck, line 124: Before this change, we passed a directory containing all release artifacts to `hc-releases`, `hc-releases` uploaded all files from this directory to the S3 bucket.

After the change, we're passing only a single file as an argument to `hc-releases`. Will this work for a release with multiple files?

dbanck (changes requested):
Thank you for taking the time to update the release process and raising this PR!

Currently, the go-releaser CI action is failing because of a syntax error. It would be great if you could look into that.

.github/workflows/release.yml
```diff
@@ -52,26 +52,19 @@ jobs:
...
-          role-duration-seconds: 3600
       -
         name: Release
         uses: goreleaser/goreleaser-action@v2
         with:
           version: latest
           args: release
         env:
+          HC_RELEASES_HOST_STAGING: ${{ secrets.HC_RELEASES_HOST_STAGING }}
+          HC_RELEASES_KEY_STAGING: ${{ secrets.HC_RELEASES_KEY_STAGING }}
```
shore, line 67: The env vars should use their real names so that goreleaser configs are not hardcoded to specific environments.
```suggestion
          HC_RELEASES_HOST: ${{ secrets.HC_RELEASES_HOST_STAGING }}
          HC_RELEASES_KEY: ${{ secrets.HC_RELEASES_KEY_STAGING }}
```

.goreleaser.yml
```diff
@@ -110,20 +110,18 @@ brews:
...
-      - AWS_ACCESS_KEY_ID={{ .Env.AWS_ACCESS_KEY_ID }}
-      - AWS_SECRET_ACCESS_KEY={{ .Env.AWS_SECRET_ACCESS_KEY }}
-      - AWS_SESSION_TOKEN={{ .Env.AWS_SESSION_TOKEN }}
-    cmd: hc-releases upload-file {{ abs .ArtifactPath }}
-
 milestones:
   - name_template: "{{ .Tag }}"
     close: true
+      - HC_RELEASES_HOST={{ .Env.HC_RELEASES_HOST_STAGING }}
+      - HC_RELEASES_KEY={{ .Env.HC_RELEASES_KEY_STAGING }}
```
shore, line 121: The env vars should use their real names so that goreleaser configs are not hardcoded to specific environments.
```suggestion
      - HC_RELEASES_HOST={{ .Env.HC_RELEASES_HOST }}
      - HC_RELEASES_KEY={{ .Env.HC_RELEASES_KEY }}
```

.goreleaser.yml
shore: The new `hc-releases` has a `-file` parameter that can be specified multiple times.  So in general, it can be called or multiple times to upload, whatever best fits the workflow.  In this particular case, I think goreleaser is running the hook once per artifact, no?

claire-labry: @dbanck after some testing, we've changed it from`-file={{ .ArtifactName }} ` to `-file={{ .ArtifactPath }}` to allow multiple files.

.goreleaser.yml
```diff
@@ -110,20 +110,18 @@ brews:
...
-      - AWS_REGION={{ .Env.AWS_REGION }}
-      - AWS_ACCESS_KEY_ID={{ .Env.AWS_ACCESS_KEY_ID }}
-      - AWS_SECRET_ACCESS_KEY={{ .Env.AWS_SECRET_ACCESS_KEY }}
-      - AWS_SESSION_TOKEN={{ .Env.AWS_SESSION_TOKEN }}
-    cmd: hc-releases upload-file {{ abs .ArtifactPath }}
-
 milestones:
   - name_template: "{{ .Tag }}"
     close: true
+      - HC_RELEASES_HOST={{ .Env.HC_RELEASES_HOST_STAGING }}
```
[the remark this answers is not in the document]
dbanck: Thanks! Build is green again :)

.goreleaser.yml
```diff
@@ -110,16 +110,14 @@ brews:
       system "#{bin}/terraform-ls --version"
 
 publishers:
-  - name: "hc-releases"
+  - name: upload
```
radeksimko, line 113:
```suggestion
  - name: "hc-releases"
```
I think in the context of GoRelaser config, the original name is still relevant?

.github/workflows/release.yml
```diff
@@ -82,7 +75,20 @@ jobs:
...
+          hc-releases-key: ${{ secrets.HC_RELEASES_KEY_STAGING }}
+      -
+        name: Promote
+        uses: hashicorp/actions-hc-releases-promote@v1
+        with:
+          product-name: "terraform-ls"
+          version: ${{ github.ref_name }}
+          hc-releases-host: ${{ secrets.HC_RELEASES_HOST_PROD }}
+          hc-releases-key: ${{ secrets.HC_RELEASES_KEY_PROD }}
+          hc-releases-source_env_key: ${{ secrets.HC_RELEASES_KEY_STAGING }}
```
radeksimko, line 94: Do we need to replace all these `*_STAGING` references with `*_PROD`? 🤔

.release/release-metadata.hcl
```diff
@@ -0,0 +1,2 @@
+url_source_repository      = "https://github/hashicorp/terraform-ls"
```
radeksimko, line 1:
```suggestion
url_source_repository      = "https://github.com/hashicorp/terraform-ls"
```

radeksimko (changes requested):

.github/workflows/release.yml
sarahethompson: `HC_RELEASES_KEY_STAGING` is correct in this context as the `hc-releases promote` command (which we are running in the action) requires the API key of the source environment in order to promote.

sarahethompson:
@radeksimko - I think we are good to go now 🤞

radeksimko (approved):
Thank you everyone involved!

I have also removed all the secrets we no longer need and will merge this now.

PR #843, opened 2022-03-31
Enable opening a single Terraform file

jpogran:
This PR enables single file support by skipping configuring the file watcher with information about what folders to watch. This leaves the watcher with little environment information, but still provides enough functionality to edit a single file. This suits the common transitory workflows for single files. For example, opening a file just to review contents or to make a few line changes without needing the context from an entire Terraform project.

A warning is shown to the user when a single file is opened, explaining that some capabilities may be reduced when editing a single file, and to open a directory to regain full editor functionality. This is to warn the user that while we enable this workflow, we don't encourage it because of the tradeoffs we currently have to do. This warning can be disabled by setting `ignoreSingleFileWarning` to `true`.

### DIFF
--- a/internal/langserver/handlers/initialize.go
+++ b/internal/langserver/handlers/initialize.go
@@ -271,19 +271,6 @@ func (svc *service) setupWalker(ctx context.Context, params lsp.InitializeParams
 	svc.openDirWalker.SetIgnoreDirectoryNames(options.IgnoreDirectoryNames)
 	svc.openDirWalker.SetExcludeModulePaths(excludeModulePaths)
 
-	// Walkers run asynchronously so we're intentionally *not*
-	// passing the request context here
-	// Static user-provided paths take precedence over dynamic discovery
-	// walkerCtx := context.Background()
-	// err = svc.closedDirWalker.StartWalking(walkerCtx)
-	// if err != nil {
-	// 	return fmt.Errorf("failed to start closedDirWalker: %w", err)
-	// }
-	// err = svc.openDirWalker.StartWalking(walkerCtx)
-	// if err != nil {
-	// 	return fmt.Errorf("failed to start openDirWalker: %w", err)
-	// }
-
 	if len(options.ModulePaths) > 0 {
 		svc.logger.Printf("Attempting to add %d static module paths", len(options.ModulePaths))
 		for _, rawPath := range options.ModulePaths {
--- a/internal/langserver/handlers/initialize.go
+++ b/internal/langserver/handlers/initialize.go
@@ -112,7 +112,7 @@ func (svc *service) Initialize(ctx context.Context, params lsp.InitializeParams)
 		if properties["options.ignoreSingleFileWarning"] == false {
 			jrpc2.ServerFromContext(ctx).Notify(ctx, "window/showMessage", &lsp.ShowMessageParams{
 				Type:    lsp.Warning,
-				Message: "Some capabilities may be reduced when editing a single file, but you can still open text files and edit them. We recommend opening a directory for full functionality.",
+				Message: "Some capabilities may be reduced when editing a single file. We recommend opening a directory for full functionality. Use 'ignoreSingleFileWarning' to suppress this warning.",
 			})
 		}
 	} else {
--- a/internal/langserver/handlers/initialize.go
+++ b/internal/langserver/handlers/initialize.go
@@ -17,7 +17,7 @@ import (
 )
 
 func (svc *service) Initialize(ctx context.Context, params lsp.InitializeParams) (lsp.InitializeResult, error) {
-	serverCaps := buildInitializeResult(ctx)
+	serverCaps := initializeResult(ctx)
 
 	out, err := settings.DecodeOptions(params.InitializationOptions)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


internal/langserver/handlers/initialize.go
```diff
@@ -188,8 +105,150 @@ func (svc *service) Initialize(ctx context.Context, params lsp.InitializeParams)
...
 	}
 
+	if params.RootURI == "" {
+		svc.singleFileMode = true
+		properties["root_uri"] = "file"
+		// return serverCaps, fmt.Errorf("Editing a single file is not yet supported." +
+		// 	" Please open a directory.")
+		jrpc2.ServerFromContext(ctx).Notify(ctx, "window/showMessage", &lsp.ShowMessageParams{
+			Type:    lsp.Warning,
+			Message: "Some of VS Code's capabilities are reduced when editing a single file, but you can still open text files and edit them. Please open a directory for full functionality.",
```
radeksimko, line 115:
```suggestion
			Message: "Some capabilities may be reduced when editing a single file. We recommend opening a directory for full functionality.",
```
I don't think we can imply any particular client here

internal/langserver/handlers/initialize.go
```diff
@@ -218,18 +277,18 @@ func (svc *service) Initialize(ctx context.Context, params lsp.InitializeParams)
...
-	if err != nil {
-		return serverCaps, err
-	}
+	// // Walker runs asynchronously so we're intentionally *not*
+	// // passing the request context here
+	// walkerCtx := context.Background()
+	// err = svc.walker.StartWalking(walkerCtx)
+	// if err != nil {
+	// 	return err
+	// }
```
radeksimko, line 286: I suppose we can keep starting of the walker(s) within the handler function, so this can be removed?

internal/langserver/handlers/initialize.go
```diff
@@ -188,8 +105,150 @@ func (svc *service) Initialize(ctx context.Context, params lsp.InitializeParams)
...
+		err := svc.setupTelemetry(tv, svc.server)
+		if err != nil {
+			svc.logger.Printf("failed to setup telemetry: %s", err)
+		}
+		svc.logger.Printf("telemetry enabled (version: %d)", tv)
+	}
+	defer svc.telemetry.SendEvent(ctx, "initialize", properties)
+}
+
+func getProperties(out *settings.DecodedOptions) map[string]interface{} {
```
radeksimko, line 151:
```suggestion
func getTelemetryProperties(out *settings.DecodedOptions) map[string]interface{} {
```

internal/langserver/handlers/initialize.go
```diff
@@ -188,8 +105,150 @@ func (svc *service) Initialize(ctx context.Context, params lsp.InitializeParams)
...
+	absPath, err := filepath.Abs(path)
+	return toLowerVolumePath(absPath), err
+}
+
+func toLowerVolumePath(path string) string {
+	volume := filepath.VolumeName(path)
+	return strings.ToLower(volume) + path[len(volume):]
+}
+
+func buildInitializeResult(ctx context.Context) lsp.InitializeResult {
```
radeksimko, line 201:
```suggestion
func initializeResult(ctx context.Context) lsp.InitializeResult {
```
It's not particularly _strong_ convention, but AFAIK it is a convention to prefix function names with the verbs like this in Go. e.g. getters and setters typically aren't prefixed with `get` or `set` in Go, unlike in other languages.

internal/langserver/handlers/initialize.go
```diff
@@ -188,8 +105,150 @@ func (svc *service) Initialize(ctx context.Context, params lsp.InitializeParams)
...
+	serverCaps.ServerInfo.Name = "terraform-ls"
+	version, ok := lsctx.LanguageServerVersion(ctx)
+	if ok {
+		serverCaps.ServerInfo.Version = version
+	}
+
+	return serverCaps
+}
+
+func (svc *service) setWalker(ctx context.Context, params lsp.InitializeParams, options settings.Options) error {
```
radeksimko, line 242:
```suggestion
func (svc *service) setupWalker(ctx context.Context, params lsp.InitializeParams, options settings.Options) error {
```

radeksimko (commented):
I gave it a try and it works pretty well! 👍🏻 

I left some naming suggestions in-line and there are now some conflicts to resolve after merging https://github.com/hashicorp/terraform-ls/pull/839

Enqueueing a new directory _was_

```go
walker.EnqueuePath(modPath)
```
and _is now_
```go
stateStore.WalkerPaths.EnqueueDir(dirUri)
```
i.e. the enqueueing happens on the state store (memdb) as opposed to the walker itself.

Secondly there are now two walkers (svc.closedDirWalker & svc.openDirWalker) so some setup methods previously called on `svc.walker` now need to be called on both instead.

I am happy to assist with the rebasing and resolution of conflicts, if you like - feel free to ping me and I can jump on Zoom with you!

internal/langserver/handlers/initialize.go
jpogran: Applied edit

internal/langserver/handlers/initialize.go
jpogran: Applied edit

internal/langserver/handlers/initialize.go
jpogran: I understand removing the `get` part, but do not understand the `telemetry` part. This builds up the struct for the settings sent by the client, not the telemetry. Though I admit I don't know what other name to name it....

radeksimko (commented):
One remaining error handling + commented out code, otherwise LGTM

radeksimko (approved):
Just proposed a small change to the warning message, but LGTM otherwise 🚀 

Tested in combination with https://github.com/hashicorp/vscode-terraform/pull/1031 and it works great!
magic

PR #846, opened 2022-04-01
langserver/handlers/command: remove `rootmodules` command

### LINKED ISSUE #515: Deprecate `rootmodules` custom command
### Current Version
```
0.16.2
```

### Use-cases

As described in https://github.com/hashicorp/terraform-ls/issues/507 this command was succeeded by `module.callers`. The new command has much narrower scope and more appropriate name.

### Proposal

It is proposed to deprecate the existing `rootmodules` command in favour of `module.callers`. Ideally clients (including the VS Code extension) should ask the server whether it supports certain commands before attempting to execute them, but I'm unsure whether this LSP convention is being followed.

Given that this command was historically used to "just" improve the UI, I'm thinking that returning error on execution should only mean that users are forced to type out module paths more often, but it shouldn't block anyone.

radeksimko:
Closes #515

Unblocks https://github.com/hashicorp/terraform-ls/pull/839


dbanck (approved):
👍

PR #851, opened 2022-04-06
fix: Ensure loading of parent directory after lower level directories works

radeksimko:
## Background

Previously if a nested directory/module was opened, kept open, and a parent directory was then opened, that parent directory would fail to load, resulting in `file not found` errors in the log and causing all IntelliSense for the parent to not be available.

The bug was introduced as part of https://github.com/hashicorp/terraform-ls/pull/771 (i.e. affects 0.26.0+)


dbanck (approved):
LGTM! I'm not able to reproduce the issue with this fix

PR #852, opened 2022-04-06
cmd: Remove `completion` CLI command

radeksimko:
## Background

This command was added in the early versions of LS primarily for debugging purposes of completion. We have since gotten more confident in the completion logic and also in the client-side, either reducing the need to debug or using the VS Code extension w/ log for debugging instead.

Finally @dbanck published https://github.com/dbanck/go-lsp-test which can further aid debugging with any LSP feature and act as a language client.

Keeping the command around also makes refactoring often more difficult than necessary.

--- 

Just for clarity for anyone less familiar with the codebase landing here - this is removing a `completion` _CLI command_, not completion functionality, which of course remains available via LSP (as exposed via `serve` CLI command).


dbanck (approved):
👍

PR #874, opened 2022-04-13
Add version to module calls

dbanck:
Depends on 
* https://github.com/hashicorp/hcl-lang/pull/115
* https://github.com/hashicorp/terraform-schema/pull/103

This PR populates the version field for `ModuleCall` and pulls in the latest `terraform-schema` and `hcl-lang` changes.

All this enables linking the source attribute for remote modules to the documentation.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,12 +14,12 @@ require (
 	github.com/hashicorp/go-uuid v1.0.3
 	github.com/hashicorp/go-version v1.4.0
 	github.com/hashicorp/hc-install v0.3.1
-	github.com/hashicorp/hcl-lang v0.0.0-20220406121211-c20527a75592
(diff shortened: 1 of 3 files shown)


radeksimko (commented):
The code LGTM, but I think we should create an issue upstream in https://github.com/microsoft/vscode to clarify the behaviour when both link & go-to-definition is associated with the same range.

I can test & re-review once the upstream PRs are merged and this is rebased.

radeksimko (approved):
🚀

Screenshot 2022-04-25 at 16 52 58

--- 

FYI I also filed https://github.com/hashicorp/terraform-ls/issues/892 just so we don't forget, but it will be handled by already planned https://github.com/hashicorp/terraform-ls/issues/882

PR #891, opened 2022-04-25
Use terraform-registry-address for parsing module sources

dbanck:
This PR uses the [latest additions to terraform-registry-address](https://github.com/hashicorp/terraform-registry-address/pull/7) to detect the source of a module.

Related: https://github.com/hashicorp/terraform-ls/pull/872


radeksimko (approved):
LGTM 🚢

PR #902, opened 2022-04-29
Subscribe to updates of module.providers data

jpogran:
Adds support for a client to register a command that will refresh clide side data based on changes the server has detected in the workspace.

This specifically adds a new command to refresh the module providers shown client side. The existing Module hook system is used to detect changes in the workspace and notify the client that data has changed.


radeksimko (approved):
LGTM :shipit: 

I found what I believe is _unrelated_ bug while testing this, filed under https://github.com/hashicorp/terraform-ls/issues/906 - we can take a look in a separate PR later.

PR #903, opened 2022-04-30
benchmarks: use --json instead of piping to results

bobheadxi:
It was not intended that this command outputs to stderr by default, and the behaviour will be removed in the future: https://github.com/bobheadxi/gobenchdata/pull/53

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -59,7 +59,7 @@ jobs:
             ${DATA_PATH} \
             ${DATA_PATH} \
             --checks.config ${CHECKS_CONFIG_PATH} \
-            2> $RESULTS_PATH # eval returns result JSON to stderr for some reason
+            --json ${RESULTS_PATH}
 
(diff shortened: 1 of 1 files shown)


radeksimko (approved):
Thanks for the PR - and of course - thanks for the project!

I think the main reason I probably overlooked the `json` flag was because `gobenchdata` generally works with a few _different_ JSON files (as both _inputs_ and _outputs_) and it wasn't clear to me at first what purpose would a flag called `json` have.

Personally I'd consider deprecating it in favour of something like `--json-to-file` which can avoid such a confusion.

I'm fine with the solution in the context of our project, but I imagine there may be valid use case for sending machine-readable (that is _not_ pretty-printed) output to stdout. It reduces the need for write permissions in the filesystem and direct piping might have some positive performance implications as there's less disk I/O. Just my 2c. 😉

bobheadxi:
Thank you for the feedback @radeksimko , I appreciate it! I hadn't considered things like write permissions and performance back when I first designed the API - if I ever get around to more significant refactors to the CLI I will keep these ideas in mind 😄

Issue #725, opened 2021-11-24
Use configuration as the primary source for `module.calls`


radeksimko:
Depends on https://github.com/hashicorp/terraform-ls/issues/722

--- 

### Background

We have recently introduced a custom command `module.calls` which currently _exclusively_ uses the data from the `.terraform` folder. There is currently no other easy way of getting the important metadata, such as (1) module hierarchy, (2) installed location on disk, (3) installed version.

As described in https://github.com/hashicorp/terraform/issues/30005 the existing behaviour of Terraform CLI is that old (removed) modules are kept in the filesystem "cache" and so using it as the only source leads to suboptimal behaviour of showing stale providers which are never removed unless the user explicitly removes them manually.

<img src="https://user-images.githubusercontent.com/287584/142994619-df3c38f0-1b58-47fd-9615-5f93c5043e39.gif">

### Use-cases

Provide more accurate representation of "module calls" for a given module URI in `module.calls` response.

### Attempted Solutions

Stale modules returned from disk cache.

### Proposal

 - [x] https://github.com/hashicorp/terraform-ls/issues/882
 - [x] Update [`module.calls` custom command](https://github.com/hashicorp/terraform-ls/blob/main/internal/langserver/handlers/command/module_calls.go) to use this metadata as a primary source and module manifest as secondary (manifest can provide hierarchy which we can't infer from the config)
 - [x] Update [relevant documentation](https://github.com/hashicorp/terraform-ls/blob/main/docs/commands.md#modulecalls)

PR #909, opened 2022-05-04
Subscribe to updates of module.calls data

jpogran:
Adds support for a client to register a command that will refresh clide side data based on changes the server has detected in the workspace.

This specifically adds a new command to refresh the module calls shown client side. The existing Module hook system is used to detect changes in the workspace and notify the client that data has changed.


internal/langserver/handlers/service.go
```diff
@@ -447,6 +447,11 @@ func (svc *service) configureSessionDependencies(ctx context.Context, cfgOpts *s
 				refreshModuleProviders(svc.sessCtx, svc.server, svc.logger, commandId))
 		}
 
+		if commandId, ok := lsp.ExperimentalClientCapabilities(cc.Experimental).RefreshModuleCallsCommandId(); ok {
+			svc.stateStore.Modules.ChangeHooks = append(svc.stateStore.Modules.ChangeHooks,
+				refreshModuleProviders(svc.sessCtx, svc.server, svc.logger, commandId))
```
radeksimko, line 452: I think we should either rename/repurpose this function as a generic `callClientCommandId()` or add a new one similar to `refreshModuleProviders`. Probably the former.

internal/langserver/handlers/handlers_test.go
```diff
@@ -71,6 +71,11 @@ func initializeResponse(t *testing.T, commandPrefix string) string {
 						"supported": true,
 						"changeNotifications": "workspace/didChangeWorkspaceFolders"
 					}
+				},
+				"experimental": {
+					"referenceCountCodeLens": false,
+					"refreshModuleProviders": false,
+					"refreshModuleCalls": false
```
radeksimko, line 78: Was there a particular motivation behind adding these? In Go fields of primitive types generally default to the "zero" - such as `false`.

radeksimko (changes requested):
Just one suggestion and two questions in-line.

internal/langserver/handlers/handlers_test.go
radeksimko: Oh right! Now I understand this relates to the bug you fixed in `initialize.go` as we now _always_ send all the experimental capabilities with their default values. That makes more sense now! Thanks for that context.

internal/langserver/handlers/hooks_module.go
```diff
@@ -153,7 +153,7 @@ func updateDiagnostics(ctx context.Context, notifier *diagnostics.Notifier) stat
 	}
 }
 
-func refreshModuleProviders(ctx context.Context, clientRequester session.ClientCaller, logger *log.Logger, commandId string) state.ModuleChangeHook {
+func callClientCommand(ctx context.Context, clientRequester session.ClientCaller, logger *log.Logger, commandId string) state.ModuleChangeHook {
 	return func(oldMod, newMod *state.Module) {
 		_, err := clientRequester.Callback(ctx, commandId, nil)
 		if err != nil {
```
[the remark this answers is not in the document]
jpogran: So the method should return early is either `oldMod` or `newMode` is nil?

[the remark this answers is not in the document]
radeksimko: I think we still want to reflect both removal or introduction in the UI, so we should still refresh, but we shouldn't crash upon logging just because it's removal.

Maybe something like this would help?

```go
var modPath string
if oldMod != nil {
  modPath = oldMod.Path
} else {
  modPath = newMod.Path
}

```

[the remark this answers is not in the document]
jpogran: Ah, now I understand. Modified!

docs/commands.md
```diff
@@ -184,3 +184,53 @@ installed version.
   }
 }
 ```
+
+# Client Side Commands
```
radeksimko, line 188: I didn't realize when we were discussing this on Slack that you meant implementation notes for LS maintainers - apologies! You were right that this isn't the best place as this is primarily for client-side maintainers. 😄 

We don't have any content for server maintainers yet, aside from the [architecture doc](https://github.com/hashicorp/terraform-ls/blob/main/docs/architecture.md). Would you mind copying this into a new one then? You can probably link it from the architecture doc.

We could call it `language-server.md`, to align the name with https://github.com/hashicorp/terraform-ls/blob/main/docs/language-clients.md

radeksimko: That said I think that some brief docs for client maintainers would be still helpful here, perhaps 1 section to describe the refresh mechanism in general and/or mention of the relevant capability under each relevant command.

radeksimko (approved):
Works great!

I did notice the effect of https://github.com/hashicorp/terraform/issues/30005 while testing this, but adding a new module & `init` worked great. We can avoid that upstream issue probably entirely as part of https://github.com/hashicorp/terraform-ls/issues/725

Issue #925, opened 2022-05-23
Reduce noise & improve performance of module hooks


radeksimko:
## Background

We use memdb hooks for a few use cases:

 - update telemetry data about modules
 - update diagnostics
 - refresh semantic tokens
 - update reference count code lens data
 - refresh `module.calls` or `module.providers` data

https://github.com/hashicorp/terraform-ls/blob/e6125557a60154e6bbc6d928c4aaf8b897e860cd/internal/langserver/handlers/service.go#L423-L459

When indexing any module, there is a number of concurrent operations which can index various details about a module. This means that in practice we may end up firing any of these hooks many times repeatedly in a short time period. This affects our performance at scale - e.g. if we fire 9 RPC events for each hook, that's `6*9 = 54` events for just a single module.

Clients may deduplicate some of these requests (esp. semantic token refresh) but it represents some wasted CPU to send these requests in the first place. This may negatively impact performance especially when there's many modules being indexed.

Performance will be especially important in the context of https://github.com/hashicorp/terraform-ls/issues/724

## Proposal

TODO
 - queue up change events in a separate memdb table - e.g. `module_events`; treat module ID as unique key to deduplicate events
 - set a reasonable delay per subscriber? e.g. telemetry can be <5secs whereas UI-affecting subscribers may need <1sec
 - maintain number of subscribers VS number of notified subscribers in memdb to ensure everyone receives the update
 - leverage memdb watchers on the `module_events` table to subscribers having to poll

PR #931, opened 2022-05-26
Batch module change events

radeksimko:
Closes https://github.com/hashicorp/terraform-ls/issues/925

## Notes

I have also left some TODOs in a few places as I believe we can further reduce the amount of events sent, but I didn't want to do it all in one PR.

This already reduces notifications for a simple root module w/ 1 submodule from 54 (18 `workspace/semanticTokens/refresh` + 18 `client.refreshModuleCalls` + 18 `client.refreshModuleProviders`) down to 3, assuming all operations finish within 1 second or earlier.

How? A new `notifier` package along with a `module_changes` memdb table is responsible for **batching module changes**. The "barrier" for releasing a batch is either elapsed 1 second from the first change or no queued/running jobs involving the module (whatever comes first).

## Data Flow

For simplicity, diagrams only represent semantic token refreshing, but in reality there's 4 more event types which would each potentially result in similar amount of refresh notifications sent.

### Before

Module Hooks Problem

### After

Module Hooks Problem (1)

## Log Before

https://gist.github.com/radeksimko/6d0901d6a65f79386e87a7450bfcad13#file-before-log

## Log After

https://gist.github.com/radeksimko/6d0901d6a65f79386e87a7450bfcad13#file-after-log


jpogran (approved):
😍

PR #932, opened 2022-05-29
Update docs for Emacs editor

psibi:
I have been recently improving support for terraform client for lsp-mode. The current documentation isn't quite right as there is no
need to manually register the client.

I'm also linking to the newly written official docs for terraform client for lsp-mode which will be helpful.

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -12,19 +12,20 @@ how you may configure the settings.
 
 ## Emacs
 
- - Install [lsp-mode](https://github.com/emacs-lsp/lsp-mode)
- - Add the following to your `.emacs`:
+If you are using `use-package`, you can use something like this:
 
-```
-(lsp-register-client
- (make-lsp-client :new-connection (lsp-stdio-connection '("/path/to/terraform-ls/terraform-ls" "serve"))
-                  :major-modes '(terraform-mode)
-                  :server-id 'terraform-ls))
-
-(add-hook 'terraform-mode-hook #'lsp)
+``` emacs-lisp
+(use-package lsp-mode
+  :ensure t
+  :hook ((terraform-mode . lsp-deferred)))
 ```
 
-The last line can instead be `(add-hook 'terraform-mode-hook #'lsp-deferred)` if you prefer the server to lazy load.
+The above code installs `lsp-mode` and also enables it for the
+terraform-mode.
+
+The terraform client for lsp-mode supports various features like
+semantic tokens, code lens for references etc. There is more detailed
+documentation [here](https://emacs-lsp.github.io/lsp-mode/page/lsp-terraform-ls/).
 
 ## IntelliJ IDE
 


docs/USAGE.md
```diff
@@ -12,19 +12,20 @@ how you may configure the settings.
...
 ```
-(lsp-register-client
- (make-lsp-client :new-connection (lsp-stdio-connection '("/path/to/terraform-ls/terraform-ls" "serve"))
-                  :major-modes '(terraform-mode)
-                  :server-id 'terraform-ls))
 
-(add-hook 'terraform-mode-hook #'lsp)
-```
+The above code installs `lsp-mode` and also enables it for the
+terraform-mode.
```
radeksimko, line 24:
```suggestion
```

docs/USAGE.md
```diff
@@ -12,19 +12,20 @@ how you may configure the settings.
...
 
-The last line can instead be `(add-hook 'terraform-mode-hook #'lsp-deferred)` if you prefer the server to lazy load.
+The terraform client for lsp-mode supports various features like
+semantic tokens, code lens for references etc. There is more detailed
+documentation [here](https://emacs-lsp.github.io/lsp-mode/page/lsp-terraform-ls/).
```
radeksimko, line 28:
```suggestion
The `lsp-mode` language client for Terraform supports various features, such as
semantic tokens, code lens for references etc. There is more detailed
documentation [here](https://emacs-lsp.github.io/lsp-mode/page/lsp-terraform-ls/).
```
As far as I can tell emacs has a single language client implementation with different configurations, so I think this wording may better reflect that?

psibi: There are two implementations available at Emacs side for language server:

- lsp-mode
- elgot

I believe the current wording is more clear, but I can change to the suggested one if you think that's more clear.

docs/USAGE.md
```diff
@@ -12,19 +12,20 @@ how you may configure the settings.
 
 ## Emacs
 
- - Install [lsp-mode](https://github.com/emacs-lsp/lsp-mode)
- - Add the following to your `.emacs`:
+If you are using `use-package`, you can use something like this:
```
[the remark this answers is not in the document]
psibi: > We should assume that even users who have not upgraded yet may land here

IIUC, MELPA's builds is not immutable. And it keeps replacing the older builds. So I don't the the older documentation is exactly right. I will push a fix where I will instead link the official documentation installation section.

> Would you mind linking to some docs about use-package?

I'm linking the official docs which has more details about use-package: https://emacs-lsp.github.io/lsp-mode/page/installation/#use-package

>  Can you clarify where does the config go exactly and/or link to some relevant docs?

It goes in the `init.el` file which is documented in the official doc. I can add it here, if you think it will help.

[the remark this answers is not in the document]
radeksimko: Thanks - this is helpful.

I think you accidentally removed the example config too though?

> It goes in the init.el file which is documented in the official doc. I can add it here, if you think it will help.

The only mention of `init.el` I can find is https://emacs-lsp.github.io/lsp-mode/page/installation/#doom-emacs which I don't think is relevant in this particular context - so yes - I think it would be helpful to document that too, and/or link to some relevant docs upstream.

docs/USAGE.md
psibi: > We can omit lsp-mod if the statement is true for both lsp-mode and elgot 

I'm not sure if the statement is true for `elgot`. From a quick look at the docs, it seems nobody has tested it for terraform-ls: https://github.com/joaotavora/eglot#connecting-to-a-server

That being said, even if it works - it will have very minimal features compared to `lsp-mode` as I made sure to support additional features for it - code lens, provides & module calls widget support etc specifically in `lsp-mode`.

>  but what I primarily meant above is that we shouldn't imply there's a "Terraform client" 

Ah, okay. I have made the suggested changes.

docs/USAGE.md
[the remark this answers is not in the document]
psibi: I have updated it, can you give it a look ? Thanks!

radeksimko (approved):
Amazing, thank you for all the edits! 👍🏻

PR #943, opened 2022-06-08
docs: revamp Troubleshooting section

radeksimko:
Based on some recent experience I decided to make the Troubleshooting section more detailed to help guide people better, especially when they're reporting logs. Namely a few people in the past provided us with the CLI execution logs instead of server logs, so the difference between the two should now be clearer.


jpogran (approved):
😍

Issue #944, opened 2022-06-08
Unable to format file. Error: "Malformated version:"


Sergey-Kirgizov:
### Server Version
installed by hombrew and reported by hombrew as v. 0.28.0
```
0.0.0-dev
platform: darwin/arm64
go: go1.18.3
compiler: gc
```

### Terraform Version
```
Terraform v1.2.2
on darwin_arm64
```

### Client Version
```
VSCode 
Version: 1.67.2 (Universal)
Commit: c3511e6c69bb39013c4a4b7b9566ec1ca73fc4d5
Date: 2022-05-17T18:20:57.384Z (3 wks ago)
Electron: 17.4.1
Chromium: 98.0.4758.141
Node.js: 16.13.0
V8: 9.8.177.13-electron.0
OS: Darwin arm64 21.5.0
```

### Terraform Configuration Files

```
terraform {
  backend "s3" {
    encrypt            = true
    region         = "us-east-1"
    key            = "some/key"
    bucket         = "some-bucket-name"
    dynamodb_table = "some-table-name"
  }
}
```

### Log Output

```
2022/06/08 21:03:23 rpc_logger.go:29: Incoming request for "textDocument/formatting" (ID 31): {"textDocument":{"uri":"file:///**********/backend.tf"},"options":{"tabSize":2,"insertSpaces":true}}
2022/06/08 21:03:23 formatting.go:30: formatting document via "/opt/homebrew/bin/terraform-ls"
2022/06/08 21:03:24 rpc_logger.go:45: Error for "textDocument/formatting" (ID 31): [-32098] fmt was first introduced in Terraform 0.7.7: unable to parse version "": Malformed version: 
2022/06/08 21:03:24 opts.go:218: Completed 1 requests [286.330583ms elapsed]
[Error - 21:03:24] Request textDocument/formatting failed.
  Message: fmt was first introduced in Terraform 0.7.7: unable to parse version "": Malformed version: 
  Code: -32098 
2022/06/08 21:03:24 opts.go:218: Received request batch of size 1 (qlen=0)
2022/06/08 21:03:24 opts.go:218: Dequeued request batch of length 1 (qlen=0)
2022/06/08 21:03:24 rpc_logger.go:29: Incoming notification for "textDocument/didSave": {"textDocument":{"uri":"file://***************/backend.tf"}}
[...]

PR #945, opened 2022-06-09
main: allow version metadata to be set

radeksimko:
This allows us - or more importantly anyone other than us - to set build metadata when building LS.

In particular this allows us to update the community Homebrew formula to help more easily distinguish LS installed from the Homebrew community tap from our own builds - which can be useful when troubleshooting - see e.g. https://github.com/hashicorp/terraform-ls/issues/944#issuecomment-1151095285

https://github.com/Homebrew/homebrew-core/blob/dc4bba37bc14ca0a7796656638d6f4df86cc3e0d/Formula/oras.rb#L21-L25

## How to test

```sh
go build -ldflags="-s -w -X main.version=0.1.0 -X main.versionMeta=homebrew"
```
```sh
./terraform-ls version
0.1.0-dev+homebrew
platform: darwin/arm64
go: go1.18.2
compiler: gc
```
```
./terraform-ls version -json
```
```json
{
  "version": "0.1.0-dev+homebrew",
  "go": "go1.18.2",
  "os": "darwin",
  "arch": "arm64",
  "compiler": "gc"
}
```

### DIFF
--- a/version.go
+++ b/version.go
@@ -1,8 +1,6 @@
 package main
 
 import (
-	"fmt"
-
 	goversion "github.com/hashicorp/go-version"
 )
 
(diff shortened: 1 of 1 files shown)


jpogran (approved):
🚢

PR #953, opened 2022-06-14
Replace internal watcher with LSP's watching mechanism

### LINKED ISSUE #867: Replace internal watcher with LSP's `DidChangeWatchedFilesRegistrationOptions`
### Background

Currently the LS implements its own watcher to catch any changes in what are typically non-user-editable, but important metadata (under `.terraform/`). Knowing when this metadata changes is important so that the server can re-load installed modules or provider schemas from there.

This became less "critical" feature in 0.16.0 after https://github.com/hashicorp/terraform-ls/pull/454 where we started pre-building schemas of ~200 official+partner providers. However, there is still some amount of users who rely on schemas sourced locally.

The LSP says the [following](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeWatchedFiles) about server-side file watching:

> Servers are allowed to run their own file system watching mechanism and not rely on clients to provide file system events. However this is not recommended due to the following reasons:

> - to our experience getting file system watching on disk right is challenging, especially if it needs to be supported across multiple OSes.
> - file system watching is not for free especially if the implementation uses some sort of polling and keeps a file system tree in memory to compare time stamps (as for example some node modules do)
> - a client usually starts more than one server. If every server runs its own file system watching it can become a CPU or memory problem.
> - in general there are more server than client implementations. So this problem is better solved on the client side.

Previously LSP however didn't provide any way for _server_ to tell the client what files it wants to receive changes for. This generally makes sense for _editable_ files (such as `*.tf` or `*.tfvars`), but it doesn't work for the files described above. i.e. we shouldn't expect clients to have list of file patterns additional to `*.tf` and `*.tfvars`.

Since LSP `3.15` there's a way to do this:

 - client sends `dynamicRegistration: true` in [`DidChangeWatchedFilesClientCapabilities`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#didChangeWatchedFilesClientCapabilities)  as part of `initialize` request
 - server responds on `initialize` request
 - server dynamically registers [`DidChangeWatchedFilesRegistrationOptions`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#didChangeWatchedFilesRegistrationOptions) by sending [`client/registerCapability`](https://microsoft.github.io/language-server-protocol/specifications/specification-3-16/#client_registerCapability) with appropriate `globPattern` per https://github.com/hashicorp/terraform-ls/blob/f49cd995e98d242914d7765bd0e0911605ee8405/internal/terraform/datadir/paths.go#L10-L21 to the client
 - server processes `workspace/didChangeWatchedFiles` notifications

## Proposal

 - [x] https://github.com/hashicorp/terraform-ls/pull/790 / https://github.com/hashicorp/terraform-ls/issues/15
 - [ ] Implement dynamic capability registration either for all capabilities or just for watched file capability
 - [ ] Remove builtin watcher
 - [ ] Update [architecture docs](https://github.com/hashicorp/terraform-ls/blob/main/docs/architecture.md#watcher)

## Related

https://github.com/hashicorp/vscode-terraform/issues/405

radeksimko:
Fixes #867 

--- 

## How to test this

I added two E2E tests confirming both provider schema & module manifest refresh is working. The previous watcher didn't have any tests (since mocking `fsnotify` events was not trivial).

One can also check the module & provider views in the VS Code extension as they update automatically after `terraform init`:

https://user-images.githubusercontent.com/287584/174062323-a4a6dce5-9c0d-44c2-8ad4-e5a4ed676a6d.mov


jpogran (approved):
This seems a lot faster in my local testing!

Issue #959, opened 2022-06-19
Language server crashes when using azure storage account blob container in vscode workspace


TeamDman:
### Server Version

```
0.28.1
```

### Terraform Version

```
Terraform v1.2.3
on windows_amd64
```

### Client Version

VSCode

![image](https://user-images.githubusercontent.com/9356891/174462226-a026c14d-6a73-4165-9c2e-4719f9fd360f.png)

Azure Storage extension
v0.14.1

### Terraform Configuration Files

an empty "a.tf" file can reproduce

### Log Output

https://gist.github.com/TeamDman/3d8684494d05b2a3f098410da395ee6c

### Expected Behavior

no crash

### Actual Behavior

crash

### Steps to Reproduce

1. Add azure storage extension
2. use azure tab in vscode to find a storage account blob container and add it to your workspace so you have a local folder open with an empty .tf file and you also have the storage account blob container open
3. Terraform language server stops working

radeksimko:
We have fixed the crash as part of https://github.com/hashicorp/terraform-ls/pull/969 - so the next release should only raise warnings, but more importantly the next release of the VS Code extension (v2.24.0) and latest preview version should not be sending URIs like these to the language server at all. This was addressed as part of https://github.com/hashicorp/vscode-terraform/pull/1163

You can either wait for these next releases, or try out the preview in the meantime:
https://marketplace.visualstudio.com/items?itemName=HashiCorp.terraform-preview

PR #969, opened 2022-06-27
fix: avoid crash on invalid URIs

radeksimko:
The language server isn't able to do anything with invalid/unsupported URIs, such as the ones listed in https://github.com/hashicorp/vscode-terraform/issues/1159

There's a broader question of whether these URIs should even be send to LS. Given that the known majority of such URIs come from VS Code, I created a separate ticket https://github.com/hashicorp/vscode-terraform/issues/1159 to discuss the problem there.

Prior to this patch it would crash when encountered such a URI during `initialize`, `textDocument/didOpen` or `workspace/didChangeWorkspaceFolders`, as reported in https://github.com/hashicorp/terraform-ls/issues/959

We already handle this gracefully in custom commands:

https://github.com/hashicorp/terraform-ls/blob/000cb538ca2325e8e7f7c8d1da86b3ea411942ba/internal/langserver/handlers/command/module_calls.go#L42-L44

https://github.com/hashicorp/terraform-ls/blob/000cb538ca2325e8e7f7c8d1da86b3ea411942ba/internal/langserver/handlers/command/module_callers.go#L30-L32

https://github.com/hashicorp/terraform-ls/blob/000cb538ca2325e8e7f7c8d1da86b3ea411942ba/internal/langserver/handlers/command/module_providers.go#L39-L41

https://github.com/hashicorp/terraform-ls/blob/000cb538ca2325e8e7f7c8d1da86b3ea411942ba/internal/langserver/handlers/command/init.go#L23-L25

https://github.com/hashicorp/terraform-ls/blob/000cb538ca2325e8e7f7c8d1da86b3ea411942ba/internal/langserver/handlers/command/validate.go#L25-L27

### DIFF
--- a/internal/langserver/handlers/did_open.go
+++ b/internal/langserver/handlers/did_open.go
@@ -2,19 +2,35 @@ package handlers
 
 import (
 	"context"
+	"fmt"
 
+	"github.com/creachadair/jrpc2"
 	"github.com/hashicorp/terraform-ls/internal/document"
 	"github.com/hashicorp/terraform-ls/internal/job"
-	ilsp "github.com/hashicorp/terraform-ls/internal/lsp"
 	lsp "github.com/hashicorp/terraform-ls/internal/protocol"
 	"github.com/hashicorp/terraform-ls/internal/state"
 	"github.com/hashicorp/terraform-ls/internal/terraform/exec"
 	"github.com/hashicorp/terraform-ls/internal/terraform/module"
 	op "github.com/hashicorp/terraform-ls/internal/terraform/module/operation"
+	"github.com/hashicorp/terraform-ls/internal/uri"
 )
 
 func (svc *service) TextDocumentDidOpen(ctx context.Context, params lsp.DidOpenTextDocumentParams) error {
-	dh := ilsp.HandleFromDocumentURI(params.TextDocument.URI)
+	docURI := string(params.TextDocument.URI)
+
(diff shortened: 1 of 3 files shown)


internal/langserver/handlers/did_open.go
```diff
@@ -2,19 +2,35 @@ package handlers
...
+	// URIs are always checked during initialize request, but
+	// we still allow single-file mode, therefore invalid URIs
+	// can still land here, so we check for those.
+	if !uri.IsURIValid(docURI) {
+		jrpc2.ServerFromContext(ctx).Notify(ctx, "window/showMessage", &lsp.ShowMessageParams{
+			Type: lsp.Warning,
+			Message: fmt.Sprintf("Ignoring workspace folder (unsupport or invalid URI) %s."+
+				" This is most likely bug, please report it.", docURI),
+		})
+		return fmt.Errorf("invalid URI: %s", docURI)
```
jpogran, line 30: Won't this open two message windows (one for warning and one for error)?

jpogran (approved):
Until we figure out why we're getting so many non file URIs I think this is a great stop-gap for improving UX

internal/langserver/handlers/did_open.go
radeksimko: Hmm, not sure actually - good question! It depends on how we deal with error from `textDocument/didOpen`. I'll test this scenario before merging.

Issue #970, opened 2022-06-27
Publish diagnostics for hidden `.*.tf` files & non-autoloaded `*.tfvars`


radeksimko:
## Background

The language server will _avoid_ publishing any diagnostics for files which aren't automatically loaded by Terraform.

 - `*.tfvars` which are _not_ `terraform.tfvars` nor `*.auto.tfvars`
 - `*.tf` which are prepended by `.` or `#`

This is because the only way we publish diagnostics today is through hooks, so there's no simple way of distinguishing between files which the user is actually editing and which ones are just loaded as a side-effect of us loading the whole directory while user is editing a different config file.

https://github.com/hashicorp/terraform-ls/blob/000cb538ca2325e8e7f7c8d1da86b3ea411942ba/internal/langserver/handlers/hooks_module.go#L128-L133

https://github.com/hashicorp/terraform-ls/blob/000cb538ca2325e8e7f7c8d1da86b3ea411942ba/internal/langserver/handlers/command/validate.go#L70-L74

## Proposal

Diagnostics for non-autoloaded files can be useful to users, so we should present it to them somehow, but we should not distract them with these diagnostics as they may not be relevant.

We could publish diagnostics as part of `textDocument/didOpen` and `textDocument/didChange`, assuming we can attach some metadata to the jobs which then end up firing the hooks that publish the diagnostics.

LSP spec states that open events may not necessarily reflect what is or isn't open, but that may be fine for this context. The spec generally doesn't cover this use case and assumes that _all_ files are loaded which doesn't apply well to Terraform here.

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_publishDiagnostics

PR #971, opened 2022-06-27
Provide basic IntelliSense (except for diagnostics) for hidden `*.tf` files

radeksimko:
Closes https://github.com/hashicorp/terraform-ls/issues/972

Depends on https://github.com/hashicorp/terraform-ls/pull/968

--- 

I label this as enhancement, since Terraform itself doesn't really load hidden files. We can however be more pragmatic about this and still provide some IntelliSense within. The only thing we do _not_ provide is diagnostics, see more on that in https://github.com/hashicorp/terraform-ls/issues/970

## Before

Screenshot 2022-06-27 at 12 43 18

## After

Screenshot 2022-06-27 at 12 43 55


dbanck (approved):
LGTM 👍

Issue #962, opened 2022-06-22
state: Allow setting explicit priority for any job


radeksimko:
## Background

Currently, jobs have implied priority based on whether files within the directory which the job affects
https://github.com/hashicorp/terraform-ls/blob/12cc996ef77526521b8b95b5eb10ea4664dff03e/internal/state/jobs.go#L193-L246

This is a reasonable design for most jobs, but https://github.com/hashicorp/terraform-ls/pull/924 introduced a new job, which is scheduled for open files, primarily _because_ it relies on up-to-date AST.

Running that job in the main indexer is rather undesirable side effect of the main indexer taking any jobs concerning open files. It's not desirable because unlike other jobs, this one has a higher probability of failing and/or taking longer to finish, potentially delaying other more important tasks.

## Proposal

Introduce more 1st class job priorities, such that the priority can be set during scheduling, instead of implied based on the file/directory.

That way we could still schedule `GetModuleDataFromRegistry` when the concerning file is open, but run it with lower priority on the background without blocking e.g. parsing or reference collections.

--- 

Example log where the request to the Registry times out: https://gist.github.com/radeksimko/20bbb39273caf8d1c557f56bf755f827

PR #977, opened 2022-06-28
job: Introduce explicit priority for jobs

radeksimko:
Fixes https://github.com/hashicorp/terraform-ls/issues/962

I also found one more bug where we'd schedule _new_ jobs as if they always involve closed directories - meaning that these would only get picked up by the lower-priority scheduler which is more likely to be busy indexing walked directories.

Therefore the fix (last commit) should further improve performance as we now correctly treat any new job for open directory as high priority.

## UX Impact

This simulates 10sec delay when obtaining data from the Module Registry.

### Before

https://user-images.githubusercontent.com/287584/176154111-712f69c2-8ac3-4612-9dac-9b1c534367b0.mov

### After

https://user-images.githubusercontent.com/287584/176154189-5e223274-451c-4741-8574-03af3e846fad.mov


dbanck (approved):
LGTM 👍 

Having high and low priority jobs makes things a lot clearer than differentiating between open and closed dir jobs!

PR #983, opened 2022-07-01
deps: bump terraform-schema to e05b3dd5db1d

radeksimko:
Changelog:
 - https://github.com/hashicorp/terraform-schema/pull/115
 - https://github.com/hashicorp/terraform-schema/pull/117

## UX
Screenshot 2022-07-01 at 10 44 33

Screenshot 2022-07-01 at 10 45 15

Screenshot 2022-07-01 at 10 45 54

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,7 @@ require (
 	github.com/hashicorp/terraform-exec v0.17.1
 	github.com/hashicorp/terraform-json v0.14.0
 	github.com/hashicorp/terraform-registry-address v0.0.0-20220623143253-7d51757b572c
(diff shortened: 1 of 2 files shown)


dbanck (approved):
👍 

Do you mind running `go mod tidy` to get rid of the old entries in the `go.sum` file? Or we can do it the next time we update a dependency.

radeksimko:
> Do you mind running go mod tidy to get rid of the old entries in the go.sum file?

Thank you for catching this! I wish there was an easy way of catching this in CI.

Looks like there might be 🔜  in Go 1.20 https://github.com/golang/go/issues/27005

PR #987, opened 2022-07-05
Use `moduleCalls.Declared` as source for `module.calls`

### LINKED ISSUE #725: Use configuration as the primary source for `module.calls`

dbanck:
WIP

Resolves #725


internal/terraform/datadir/module_types.go
```diff
@@ -16,44 +17,37 @@ const (
...
-	"../",
-	".\\",
-	"..\\",
-}
-
 // GetModuleType parses source addresses to determine what kind of source the Terraform module comes
 // from. It currently supports detecting Terraform Registry modules, GitHub modules, Git modules, and
 // local file paths
-func (r *ModuleRecord) GetModuleType() ModuleType {
+func GetModuleType(sourceAddr string) ModuleType {
```
radeksimko, line 23: Is it worth moving this and dependent logic out of the `datadir` package, given that it no longer deals with any data within `.terraform` (the data dir)? Given that `module.calls` command/handler is the only consumer it might make sense to move it to the same file, where we already have `getModuleDocumentationLink()`?

internal/langserver/handlers/command/module_calls.go
```diff
@@ -48,72 +49,41 @@ func (h *CmdHandler) ModuleCallsHandler(ctx context.Context, args cmd.CommandArg
...
-
+func (h *CmdHandler) parseModuleRecords(ctx context.Context, moduleCalls tfmod.ModuleCalls) []moduleCall {
 	modules := make(map[string]moduleCall)
-	for _, manifest := range records {
-		if manifest.IsRoot() {
-			// this is the current directory, which is technically a module
-			// skipping as it's not relevant in the activity bar (yet?)
+	for _, manifest := range moduleCalls.Declared {
+		if manifest.SourceAddr == nil {
+			// We skip all modules without a source address
```
radeksimko, line 66: Can you add a test for this case? On a related note I'm wondering if we should add the edge case of empty string to tfschema here https://github.com/hashicorp/terraform-schema/blob/43f5a813b1b10b271d39e7406c61b0eb807a5a9a/earlydecoder/load_module.go#L310-L318

internal/langserver/handlers/command/module_calls.go
```diff
@@ -48,72 +49,41 @@ func (h *CmdHandler) ModuleCallsHandler(ctx context.Context, args cmd.CommandArg
...
-		// is it better to look at SourceAddr and compare?
-		if strings.Contains(manifest.Key, ".") {
-			v := strings.Split(manifest.Key, ".")
-			moduleName = v[0]
-			subModuleName = v[1]
-		}
-
-		docsLink, err := getModuleDocumentationLink(ctx, manifest)
+		moduleName := manifest.LocalName
+		docsLink, err := getModuleDocumentationLink(ctx, manifest.SourceAddr.String(), manifest.Version.String())
```
radeksimko, line 71: Nitpick: Is it worth making use of the interface resulting from an already parsed address here, instead of rebuilding the string and parsing it again within `getModuleDocumentationLink()`?

docs/commands.md
```diff
@@ -127,7 +126,7 @@ The data is sourced from the local cache within `.terraform` and so it may not n
    - `version` - version constraint of the module call; applicable to modules hosted by the Terraform Registry (e.g. `~> 1.0`
    - `source_type` - source of the Terraform module, e.g. `github` or `tfregistry`
    - `docs_link` - a link to the module documentation; if available
-   - `dependent_modules` - array of dependent modules with the same structure as `module_calls`
+   - `dependent_modules` - *deprecated* - array of dependent modules with the same structure as `module_calls`
```
radeksimko, line 129:
```suggestion
   - `dependent_modules` - **DEPRECATED** (always empty in `v0.29+`) - array of dependent modules with the same structure as `module_calls`
```

We add more context about _why_ in the Changelog, I'd just call out the version here.

internal/langserver/handlers/command/module_calls.go
dbanck: Since we now have a default case with `UnknownSourceAddr`, we might now be able to remove this check entirely. Before I was running into all kind of null-pointer exceptions. Maybe we can discuss this over :zoom:

radeksimko (approved):
Just one minor suggestion about the formatting of the source address.

Looks so smooth now! 🌟 😍

internal/langserver/handlers/command/module_calls.go
```diff
@@ -48,72 +60,43 @@ func (h *CmdHandler) ModuleCallsHandler(ctx context.Context, args cmd.CommandArg
...
+		docsLink, err := getModuleDocumentationLink(ctx, module.SourceAddr)
 		if err != nil {
 			h.Logger.Printf("failed to get module docs link: %s", err)
 		}
 
 		// build what we know
 		moduleInfo := moduleCall{
 			Name:             moduleName,
-			SourceAddr:       manifest.SourceAddr,
+			SourceAddr:       module.SourceAddr.String(),
```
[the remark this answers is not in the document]
dbanck: Sounds good to me!

How would you rephrase the docs? To me, the current version already captures this
> - `source_addr` - the source address given for this module call (e.g. `terraform-aws-modules/eks/aws`)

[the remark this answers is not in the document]
radeksimko: I'd phrase it as

> - `source_addr` - human-readable version of the source address given for this module call (e.g. `terraform-aws-modules/eks/aws`)

Issue #993, opened 2022-07-06
decoder: Pick core schema only based on `required_version` constraint


radeksimko:
https://github.com/hashicorp/terraform-ls/issues/992 removes the other reason we run `terraform version`, so this issue may depend on it - at least we cannot avoid running it until we have other ways to get _provider_ versions.

## Background

Currently we pick core schema based on version that we determine by running `terraform version -json` via `GetTerraformVersion` job

https://github.com/hashicorp/terraform-ls/blob/81b49a174bd8f34a78ab959f7ab88325177b1daf/internal/terraform/module/module_ops.go#L54-L89

which we in turn run within every walked directory

https://github.com/hashicorp/terraform-ls/blob/81b49a174bd8f34a78ab959f7ab88325177b1daf/internal/indexer/walker.go#L61-L73

This makes walking through a deep hierarchy with many modules quite an expensive process.

The scheduler already de-duplicates jobs (including `GetTerraformVersion`), such that we never run the same job for the same folder more than once, if it's already in the queue. This is a good barrier, but we anticipate the following features to put more pressure on the scheduler and create more situations of duplicate work (same job for the same folder) scheduled more sparsely, which the scheduler itself has no chance of catching.

 - https://github.com/hashicorp/terraform-ls/issues/598
 - https://github.com/hashicorp/terraform-ls/issues/724

both of which increase the chance of us scheduling the same jobs for the same directory repeatedly.

## Proposal

 - [x] `terraform-schema`: Introduce a function similar to the existing [`CoreModuleSchemaForVersion`](https://pkg.go.dev/github.com/hashicorp/terraform-schema@v0.0.0-20220705091919-a4d3bf44d1d2/schema#CoreModuleSchemaForVersion) which takes version constraint instead of version and returns relevant root schema
   - Undeclared, empty constraint or constraint without upper bound can assume the latest TF version
 - [x] `decoder`: Update logic responsible for picking core schema https://github.com/hashicorp/terraform-ls/blob/81b49a174bd8f34a78ab959f7ab88325177b1daf/internal/decoder/module_schema.go#L14-L26

## Notes

While running `terraform version` for every walked directory is expensive, there's still value in knowing Terraform version.

 - more accurate IntelliSense (esp. relevant if users don't use version constraints at all)
   - e.g. (assuming versioned docs) we can link to docs relevant to that particular version, not just first or last version where a language construct was present
 - future UX improvements nudging to upgrade TF

[...]

Issue #992, opened 2022-07-06
terraform/module: Introduce new job to parse provider versions from lock file


radeksimko:
## Background

Currently we treat provider lock file (and installed provider binaries) as opaque and only use them in the watcher to know whether they have changed, to re-run `terraform version` and `terraform providers schema -json` to refresh provider versions and obtain new schemas.

The scheduler already de-duplicates jobs (including the two mentioned above), such that we never run the same job for the same folder more than once, if it's already in the queue. This is a good barrier, but we anticipate the following features to put more pressure on the scheduler and create more situations of duplicate work (same job for the same folder) scheduled more sparsely, which the scheduler itself has no chance of catching.

 - https://github.com/hashicorp/terraform-ls/issues/598
 - https://github.com/hashicorp/terraform-ls/issues/724

both of which increase the chance of us scheduling the same jobs for the same directory repeatedly.

The root problem is that we don't know if we need the schema before we run these commands and so we potentially waste CPU and IO on both jobs unless they end up deduplicated within the same job queue.

There are different ways that Terraform persists the installed provider versions between TF versions:
https://github.com/hashicorp/terraform-ls/blob/9fcc1c971fccae82146c0104e7f861c7b255e486/internal/terraform/datadir/paths.go#L12-L19

While relying on more of Terraform's internal state is not great, the cost/risk seems well out-weighted by our ability to prevent wasteful CLI commands if we know that we already have the same provider schema for the same version already available in memdb cache.

## Proposal

 - [ ] `datadir`: Introduce a function to parse provider versions from the data dir (`.terraform`)
    - we can attempt to parse different `.terraform` structures from the latest TF version to oldest and return the first successful parse. This will allow us to keep this logic entirely decoupled from parsing and understanding the core version or constraint
 - [ ] `terraform/module`: Introduce a new job `ParseProviderVersions` to https://github.com/hashicorp/terraform-ls/blob/66c8742/internal/terraform/module/module_ops.go
[...]

Issue #989, opened 2022-07-06
state: Track module directory ModTime


radeksimko:
## Background

Currently, whenever we schedule `ParseModuleConfiguration`, `ParseVariables`, or `ParseModuleManifest` from anywhere, we just parse these files without checking whether we have already parsed the exact same file (with the same content) previously.

The scheduler already de-duplicates jobs, such that we never run the same job for the same folder more than once, if it's already in the queue. This is a good barrier, but we anticipate the following features to put more pressure on the scheduler and create more situations of duplicate work (same job for the same folder) scheduled more sparsely, which the scheduler itself has no chance of catching.

 - https://github.com/hashicorp/terraform-ls/issues/598
 - https://github.com/hashicorp/terraform-ls/issues/724

To avoid wasting CPU on these unnecessary jobs we need some way of checking what files we have processed and whether they changed since.

The simplest way to do it would be reading, storing and comparing [`ModTime()`](https://pkg.go.dev/io/fs#FileInfo) (modification time) of the module folder.

## Proposal

 - [ ] Introduce `DirModTime time.Time` to [`state.Module`](https://github.com/hashicorp/terraform-ls/blob/81b49a174bd8f34a78ab959f7ab88325177b1daf/internal/state/module.go#L83-L85)
 - [ ] `state`: Add `lastModTime time.Time` as a 2nd argument to [`ModuleStore.Add(modPath string)`](https://github.com/hashicorp/terraform-ls/blob/81b49a174bd8f34a78ab959f7ab88325177b1daf/internal/state/module.go#L228)
 - [ ] `state`: Introduce `ModuleStore.HasChangedSince(modPath string, t time.Time) (bool, error)`
 - [ ] `walker`: Check `ModTime()` of every dir and pass it to `HasChangedSince()` before scheduling any jobs
 - [ ] `textDocument/didOpen` handler: Check `ModTime()` of every dir and pass it to `HasChangedSince()` before scheduling any jobs https://github.com/hashicorp/terraform-ls/blob/81b49a174bd8f34a78ab959f7ab88325177b1daf/internal/langserver/handlers/did_open.go#L59-L66
 - [ ] `workspace/didChangeWorkspaceFolders` handler: Check `ModTime()` of every dir and pass it to `HasChangedSince()` before scheduling any jobs

radeksimko:
@jpogran raised some valid concerns/questions yesterday:

[...]

PR #997, opened 2022-07-08
walker: Index uninitialized modules

radeksimko:
Fixes https://github.com/hashicorp/terraform-ls/issues/724

~Depends on https://github.com/hashicorp/terraform-schema/pull/126~

--- 

The PR changes the walker such that it indexes any directory it finds with >0 `*.tf` files in it. This has a few side effects I had to account for:

 - as a result of indexing many more modules, this uncovered a race condition where reference collection would run before metadata decoding, resulting in no references being collected.
   - Reference collection jobs were moved under `Defer` to provide that guarantee.
 - Some E2E tests assume that the existence of `.terraform` dir **without other files** would trigger indexing, but the new behaviour is more correct in not indexing such a directory. The reason we start with empty directory with just `.terraform` is because it means we can more easily provide the actual test configuration as part of the test via `didOpen` and we don't have to deal with copying/creating test configs in FS. 
   - The same function responsible for "bootstrapping" a directory to become indexable was updated to also create `empty.tf`
   - We could update all these E2E tests such that they don't rely on walker at all - i.e. we don't do any indexing on `initialize`, but that would also mean we have to find a different way of loading the schema, e.g. by inserting it into memdb as part of each test.
   - It is questionable whether these tests are in fact truly E2E when they mock out the Terraform CLI calls. I think the best solution might be having two versions of most tests, where (1) is truly E2E and installs and executes real Terraform CLI, and (2) mocks out enough memdb state such that tests can run offline
   - We so far used many E2E tests as a way to test the walker, so we'd have to also add some unit tests to the walker if we no longer run it as part of the E2E tests.

--- 

## Indexing Before

New Indexer (now vs proposed)@2x

## Indexing After

New Indexer (now vs proposed)@2x (3)

--- 

I filed a few follow-up issues to deal with the potential performance hit resulting from indexing more modules:
 - https://github.com/hashicorp/terraform-ls/issues/989
 - https://github.com/hashicorp/terraform-ls/issues/992
 - https://github.com/hashicorp/terraform-ls/issues/993


dbanck (approved):
Nice work! This is really fast & fun to use.

I tried different numbers of modules and couldn't see any performance impact

PR #999, opened 2022-07-09
Update usage doc supported filetypes and neovim docs

### LINKED ISSUE #660: docs: Reflect tfvars support in USAGE (client setup guides)
Currently we do not actively encourage clients to detect and send tfvars files to the language server here:

https://github.com/hashicorp/terraform-ls/blob/main/docs/USAGE.md

The only mention so far is in the dev-oriented docs:
https://github.com/hashicorp/terraform-ls/blob/main/docs/language-clients.md

teddylear:
Update usage doc supported filetypes and neovim docs with lua specific configuration. If there is anything I should add and/or change please let me know. 

Fixes #660

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -136,13 +136,21 @@ let g:LanguageClient_serverCommands = {
 ### Neovim v0.5.0+
 
  - Install the [nvim-lspconfig plugin](https://github.com/neovim/nvim-lspconfig)
- - Add the following to your `.vimrc` or `init.vim` or `init.lua`:
+ - Add the following to your `.vimrc` or `init.vim`:
 
 ```vim
 lua <<EOF
   require'lspconfig'.terraformls.setup{}
 EOF
 autocmd BufWritePre *.tf lua vim.lsp.buf.formatting_sync()
+```
+ - If you are using `init.lua`:
+```lua
+require'lspconfig'.terraformls.setup{}
+vim.api.nvim_create_autocmd({"BufWritePre"}, {
+  pattern = {"*.tf"},
+  callback = vim.lsp.buf.formatting_sync,
+})
 ```
 
 Make sure to read through to [server_configurations.md#terraformls](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#terraformls) if you need more detailed settings.
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -4,6 +4,16 @@ This guide assumes you have installed the server by following instructions
 in the [README.md](../README.md) if that is applicable to your client
 (i.e. if the client doesn't download the server itself).
 
+The following filetypes are supported by the Terraform Language Server:
+
+- `terraform` - standard `*.tf` config files
+- `terraform-vars` - variable files (`*.tfvars`)
+
+*NOTE* Clients specifically should **not** send `*.tf.json`, `*.tfvars.json` nor
+Packer HCL config nor any other HCL config files as the server is not
+equipped to handle these file types.
+
+
 Instructions for popular IDEs are below and pull requests
 for updates or addition of more IDEs are welcomed.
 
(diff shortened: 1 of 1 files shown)


docs/USAGE.md
```diff
@@ -126,13 +136,21 @@ let g:LanguageClient_serverCommands = {
...
-  require'lspconfig'.terraformls.setup{} 
+  require'lspconfig'.terraformls.setup{}
 EOF
 autocmd BufWritePre *.tf lua vim.lsp.buf.formatting_sync()
+```
+ - If you are using `init.lua`:
+```lua
+require'lspconfig'.terraformls.setup{}
+vim.api.nvim_create_autocmd({"BufWritePre"}, {
+  pattern = {"*.tf"},
```
radeksimko, line 151: A few questions here:

1. This appears to only recognize `*.tf`, not `*.tfvars`, is there a way we can recognise the vars files too?
2. Do you know where/how does NeoVim map these patterns to language IDs as documented? i.e. how does it know to map `*.tf` to `terraform` (and not any other arbitrary language identifier)? I can see [this setting](https://github.com/neovim/nvim-lspconfig/blob/41a826947eefbc7b59cc80cef405cce9ea6d82b8/lua/lspconfig/server_configurations/terraformls.lua#L6) which seems relevant but it's not clear to me how to configure more than 1 language per server.
3. Since these file naming conventions are pretty much hard-coded into Terraform, I don't see any way users can choose arbitrary names (as these would be ignored by Terraform and Terraform does _not_ make this configurable). So is there any way we could just embed this into the default NeoVim configuration so users don't have to configure themselves?

radeksimko (commented):
Thank you for the PR @teddylear , I left some questions in-line.

docs/USAGE.md
teddylear: @radeksimko Thanks for the quick review! Addressing your questions: 
1. I updated PR with adding `*.tfvars`
2. Looking into neovim source I see this for settings filetypes by extension: 
- https://github.com/neovim/neovim/blob/master/runtime/filetype.vim#L1984
- https://github.com/neovim/neovim/blob/master/runtime/filetype.vim#L2006
3. Unfortunately I don't think we can hard code these. These are an optional setting for neovim for doing a format on write of a buffer, which does have a performance hit. So users have to opt in for something like this. If nvim-lsp starts adding a config for making this easier, then it could be hard coded. However for now I don't think that is the case. 

If you have any other concerns/questions please let me know. Once again, thanks for such a quick review turnaround

radeksimko: (3) Fair enough. I mostly meant hard-coding somewhere here https://github.com/neovim/nvim-lspconfig/blob/41a826947eefbc7b59cc80cef405cce9ea6d82b8/lua/lspconfig/server_configurations/terraformls.lua#L4-L8 but I now realize that this is only about formatting, like you said. Do you know why does formatting require this additional handling **based on filenames**, when LS (as demonstrated by the linked Lua config) provides all the other IntelliSense based on **language IDs**? 🤔 I'm assuming that NeoVim would still find the relevant language ID from the mappings you linked to, find an active LS and send the formatting request there? Which is fine, it just feels a little awkward.

(1) (2) Thank you for the update. Unfortunately, based on [the code you linked to](https://github.com/neovim/neovim/blob/540d1af90e2276a8bd3b5edb1a63c8225a1aa423/runtime/filetype.vim#L1983-L1984), it appears we cannot safely enable support for variable files, unless/until `*.tfvars` maps to `terraform-vars` (not to `terraform`), otherwise the language server would be treating `*.tfvars` as `*.tf` and likely report lots of irrelevant diagnostics, completion etc.

Would you mind raising that PR upstream to NeoVim? Then I guess we could explain here that `*.tfvars` support is from NeoVim version X.Y.Z or something.
On a related note, I'm guessing that we'll also need to add `terraform-vars` [here](https://github.com/neovim/nvim-lspconfig/blob/41a826947eefbc7b59cc80cef405cce9ea6d82b8/lua/lspconfig/server_configurations/terraformls.lua#L6)?

radeksimko (commented):
Thanks for the updates

docs/USAGE.md
radeksimko: I have raised https://github.com/neovim/neovim/pull/19417

radeksimko: The proposed fixes with `terraform-vars` were both merged:

 - https://github.com/vim/vim/pull/10755
 - https://github.com/neovim/neovim/pull/19526

NeoVim has not cut a release yet with this change, but I expect they will soon.

I will just make a minor adjustment of the wording and merge it - thanks!

radeksimko (approved):

Issue #704, opened 2021-11-08
Repurpose `rootModulePaths` and `excludeRootModulePaths`


radeksimko:
The following config options should not be needed and if they are then we can direct people to the native LSP feature workspace folders, which we treat pretty much the same way already. If a client doesn't support that feature then this may help encourage the client to implement it. 🤷🏻 

https://github.com/hashicorp/terraform-ls/blob/91ba27b71f309c64fd0bff399bfe52824d71a192/internal/settings/settings.go#L19-L21

https://github.com/hashicorp/terraform-ls/blob/91ba27b71f309c64fd0bff399bfe52824d71a192/internal/langserver/handlers/initialize.go#L212-L226

radeksimko:
https://github.com/hashicorp/terraform-ls/issues/724 may need to be implemented first to ensure that adding a workspace with uninitialized module also gets it indexed and therefore would be comparable to `rootModulePaths`

jpogran:
We can add `deprecated` to the client settings in VS Code so users know this is coming while we work on the above.

PR #1003, opened 2022-07-18
settings: repurpose indexer-related options

radeksimko:
Closes https://github.com/hashicorp/terraform-ls/issues/704

### DIFF
--- a/internal/langserver/handlers/initialize.go
+++ b/internal/langserver/handlers/initialize.go
@@ -183,9 +183,8 @@ func getTelemetryProperties(out *settings.DecodedOptions) map[string]interface{}
 		"lsVersion":                                       "",
 	}
 
-	properties["options.rootModulePaths"] = len(out.Options.ModulePaths) > 0
-	properties["options.rootModulePaths"] = len(out.Options.ModulePaths) > 0
-	properties["options.excludeModulePaths"] = len(out.Options.ExcludeModulePaths) > 0
+	properties["options.rootModulePaths"] = len(out.Options.XLegacyModulePaths) > 0
+	properties["options.excludeModulePaths"] = len(out.Options.XLegacyExcludeModulePaths) > 0
 	properties["options.commandPrefix"] = len(out.Options.CommandPrefix) > 0
 	properties["options.ignoreDirectoryNames"] = len(out.Options.IgnoreDirectoryNames) > 0
 	properties["options.experimentalFeatures.prefillRequiredFields"] = out.Options.ExperimentalFeatures.PrefillRequiredFields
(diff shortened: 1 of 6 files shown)


radeksimko:
Tagging @jpogran as you dealt with settings in the VS Code extension recently:

We could potentially move both `ignoreDirectoryNames` and `ignorePaths` under `indexing` or similar, to indicate these options are related to indexing. This can of course be also done just on the client side, but I thought I'd flag it up here as we're touching the settings.

Also I can prepare a PR for the extension which reflects these changes and decisions.

internal/langserver/handlers/initialize.go
```diff
@@ -248,14 +247,33 @@ func (svc *service) setupWalker(ctx context.Context, params lsp.InitializeParams
...
+				options.XLegacyExcludeModulePaths),
+		})
+	}
+
+	var ignoredPaths []string
+	for _, rawPath := range options.IgnorePaths {
 		modPath, err := resolvePath(root.Path(), rawPath)
 		if err != nil {
-			svc.logger.Printf("Ignoring excluded module path %s: %s", rawPath, err)
+			jrpc2.ServerFromContext(ctx).Notify(ctx, "window/showMessage", &lsp.ShowMessageParams{
```
dbanck, line 269: Do we want to show a message to the user in this case? For example, if a user switches between many projects, the paths might only apply to some. VS Code has workspace-specific settings for this, but users might not use them or use an entirely different editor.

dbanck (approved):
LGTM! 

Does this qualify as a breaking change and requires extra communication?

radeksimko:
> Does this qualify as a breaking change and requires extra communication?

Yes, I expect we'd call it out as such in the Changelog.

jpogran (approved):
😍

internal/langserver/handlers/initialize.go
jpogran: 👍🏼 on the no message from server in this case.

jpogran:
> Tagging @jpogran as you dealt with settings in the VS Code extension recently:
> 
> We could potentially move both `ignoreDirectoryNames` and `ignorePaths` under `indexing` or similar, to indicate these options are related to indexing. This can of course be also done just on the client side, but I thought I'd flag it up here as we're touching the settings.
> 
> Also I can prepare a PR for the extension which reflects these changes and decisions.

👍🏼 on moving to indexing. 

I think we've already marked these as deprecated, so we're halfway there

internal/langserver/handlers/initialize.go
radeksimko: I filed https://github.com/hashicorp/terraform-ls/issues/1008 but after re-reading the code here I realized the warning would not be sent just because a path doesn't exist - only if the URI is invalid - so that seems like a good reason to continue raising warning to me?

Issue #996, opened 2022-07-07
Account for module calls pointing outside of the workspace which cannot be indexed via the walker ^ and for any outdated installed modules


dbanck:

radeksimko:
After internal discussion we decided to put this on hold for now.

One concern was that we would not receive watch events for these modules outside the workspace and we aren't able to watch them until the (gopls) LSP structs bring the following: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#relativePattern

PR #1004, opened 2022-07-18
docs: Clarify usage of workspaces

radeksimko:
Tracking of paths outside the workspace can be re-evaluated as part of https://github.com/hashicorp/terraform-ls/issues/996

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -10,6 +10,21 @@ for updates or addition of more IDEs are welcomed.
 See also [settings](./SETTINGS.md) to understand
 how you may configure the settings.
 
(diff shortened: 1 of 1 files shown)


dbanck (approved):

esn89:
@radeksimko 

How do I specify the workspace root?  I have a lot of workspaces

radeksimko:
@esn89 That is established by the client/editor (e.g. VS Code) and typically it's the root of the first workspace that was added.

FYI We also have https://discuss.hashicorp.com/c/terraform-core/terraform-editor-integrations/46 which we prefer as a medium for Q/A.

PR #1006, opened 2022-07-19
indexer/walker: Avoid running jobs where not needed

### LINKED ISSUE #1002: terraform/module: Avoid running a job if nothing has changed
## Background

This is a different approach to addressing a performance bottleneck when we have duplicate jobs, as originally described in https://github.com/hashicorp/terraform-ls/issues/989

## Proposal

Remove deduplication logic from the scheduler:

https://github.com/hashicorp/terraform-ls/blob/ca335f5ec3f320ab5a517592ae63ac90b04f127f/internal/state/jobs.go#L61-L75

Introduce `IgnoreExistingState bool` to `job.Job{}`, to indicate whether to ignore existing data/cache.

 - `initialize` - `IgnoreExistingState: false`
 - `textDocument/didOpen` - `IgnoreExistingState: true` - this is only necessary to account for an edge case where the user opens editor with unsaved files from previous session, we could potentially avoid it if we tracked `ModTime` or any kind of indicator allowing us to compare the file being opened with the one already indexed, effectively implementing [this TODO](https://github.com/hashicorp/terraform-ls/blob/ca335f5ec3f320ab5a517592ae63ac90b04f127f/internal/langserver/handlers/did_open.go#L59-L66)
 - `textDocument/didChange` - `IgnoreExistingState: true`
 - `workspace/didChangeWorkspaceFolders` - `IgnoreExistingState: true`
 - `workspace/didChangeWatchedFiles` - `IgnoreExistingState: true`

The job could return return an error, such as `NotChangedError{Handle: document.DirHandle}`, such that downstream jobs can decide based on that, whether to run.

 - [ ] `GetTerraformVersion` - https://github.com/hashicorp/terraform-ls/issues/993
 - [ ] `ObtainSchema` - https://github.com/hashicorp/terraform-ls/issues/992
 - [ ] `ParseModuleConfiguration` - check `ModuleParsingState`
 - [ ] `ParseVariables` - check `VarsParsingState`
 - [ ] `ParseModuleManifest` - check `ModManifestState`
 - [ ] `LoadModuleMetadata` - check `MetaState` field + `NotChangedError`
 - [ ] `DecodeReferenceTargets` - check `NotChangedError`
 - [ ] `DecodeReferenceOrigins` - check `NotChangedError`
 - [ ] `DecodeVarsReferences` - check `NotChangedError`
 - [ ] `GetModuleDataFromRegistry` - check `NotChangedError`

## Implementation Notes

Regarding `NotChangedError` - some downstream jobs run after multiple different jobs, so we need to make sure that the upstream job actually returns `NotChangedError` if and _only if_ all jobs detect no changes.

radeksimko:
Closes #1002

Depends on #1027

--- 

## Background

The removal of de-duplication upon job enqueuing is necessary to avoid various bugs and race conditions where we'd deduplicate jobs which appear to be same, but have different `Defer` or `DependsOn`, i.e. we'd be overly aggressive in the deduplication efforts.

It is worth clarifying here that this PR doesn't affect jobs which run as part of text synchronization - i.e. `textDocument/didOpen`, `textDocument/didChange` or `workspace/didChangeWatchedFiles`. It only aims to reduce the duplicated work as part of the walker (which is triggered by `initialize`), which would occur if the user has deep workspace which takes a while to index via walker, and opens a few (yet unindexed) modules, which get indexed as a priority.

For `textDocument/didChange` and `workspace/didChangeWatchedFiles` it is expected that the jobs _do_ need to run since we _know_ something has changed. We could attempt to do some de-duplication there as well, but it's probably going to have low impact.

For `textDocument/didOpen` - we treat any opened document as a change currently, because we have no way to tell whether it matches the contents on disk. This is IMO a common case, where user would start without any open files, walker indexes everything, and then they open any module and we re-index the whole module again, and we do it again for **every single file** they open. Reducing duplicated work here is just little more involved, so I filed a separate ticket for that: https://github.com/hashicorp/terraform-ls/issues/1031

### Benchmarks

I am (finally) updating the benchmarks as part of the PR, which may suggest that this PR has negative performance impact, but it's little more complicated as we're really reflecting a few PRs, each contributing to those numbers in slightly different way:

 - https://github.com/hashicorp/terraform-ls/pull/997 increased the total number of jobs in most cases
 - https://github.com/hashicorp/terraform-ls/pull/1014 reduced time as we no longer run `terraform providers schema -json` when it's not necessary. Previously we'd run it on _every_ indexed & initialized module.
   - This impacts cases of deep workspace hierarchies with initialized modules which use the same provider versions, which also makes more effective use of the embedded schemas now.
 - https://github.com/hashicorp/terraform-ls/pull/1027 removed 1 job (`terraform version`) entirely from the walker indexing. It now runs only on the text synchronization methods.

### DIFF
--- a/internal/indexer/walker.go
+++ b/internal/indexer/walker.go
@@ -97,7 +97,7 @@ func (idx *Indexer) WalkedModule(ctx context.Context, modHandle document.DirHand
 			},
 			Type: op.OpTypeParseModuleManifest.String(),
 			Defer: func(ctx context.Context, jobErr error) (job.IDs, error) {
-				return idx.decodeInstalledModuleCalls(modHandle)
+				return idx.decodeInstalledModuleCalls(modHandle, false)
 			},
 		})
 		if err != nil {
@@ -114,7 +114,7 @@ func (idx *Indexer) WalkedModule(ctx context.Context, modHandle document.DirHand
 		pSchemaVerId, err := idx.jobStore.EnqueueJob(job.Job{
 			Dir: modHandle,
 			Func: func(ctx context.Context) error {
-				return module.ParseProviderVersions(idx.fs, idx.modStore, modHandle.Path())
+				return module.ParseProviderVersions(ctx, idx.fs, idx.modStore, modHandle.Path())
 			},
 			Type:      op.OpTypeParseProviderVersions.String(),
 			DependsOn: providerVersionDeps,
@@ -144,7 +144,7 @@ func (idx *Indexer) WalkedModule(ctx context.Context, modHandle document.DirHand
 	}
 
 	if parseId != "" {
-		rIds, err := idx.collectReferences(modHandle, refCollectionDeps)
+		rIds, err := idx.collectReferences(modHandle, refCollectionDeps, false)
 		if err != nil {
 			errs = multierror.Append(errs, err)
 		} else {
--- a/internal/indexer/walker.go
+++ b/internal/indexer/walker.go
@@ -22,7 +22,7 @@ func (idx *Indexer) WalkedModule(ctx context.Context, modHandle document.DirHand
 	parseId, err := idx.jobStore.EnqueueJob(job.Job{
 		Dir: modHandle,
 		Func: func(ctx context.Context) error {
-			return module.ParseModuleConfiguration(idx.fs, idx.modStore, modHandle.Path())
+			return module.ParseModuleConfiguration(ctx, idx.fs, idx.modStore, modHandle.Path())
 		},
 		Type: op.OpTypeParseModuleConfiguration.String(),
 	})
@@ -40,7 +40,7 @@ func (idx *Indexer) WalkedModule(ctx context.Context, modHandle document.DirHand
 			Dir:  modHandle,
 			Type: op.OpTypeLoadModuleMetadata.String(),
 			Func: func(ctx context.Context) error {
-				return module.LoadModuleMetadata(idx.modStore, modHandle.Path())
+				return module.LoadModuleMetadata(ctx, idx.modStore, modHandle.Path())
 			},
 			DependsOn: job.IDs{parseId},
 		})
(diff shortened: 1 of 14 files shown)


internal/indexer/walker.go
```diff
@@ -111,47 +111,29 @@ func (idx *Indexer) WalkedModule(ctx context.Context, modHandle document.DirHand
...
 
-				return ids, nil
+		pSchemaId, err := idx.jobStore.EnqueueJob(job.Job{
+			Dir: modHandle,
+			Func: func(ctx context.Context) error {
+				ctx = exec.WithExecutorFactory(ctx, idx.tfExecFactory)
+				return module.ObtainSchema(ctx, idx.modStore, idx.schemaStore, modHandle.Path())
 			},
+			Type:      op.OpTypeObtainSchema.String(),
+			DependsOn: job.IDs{pSchemaVerId},
```
dbanck, line 136: Should we check that `pSchemaVerId` isn't `""` here?

internal/indexer/watcher.go
```diff
@@ -33,55 +35,37 @@ func (idx *Indexer) ModuleManifestChanged(ctx context.Context, modHandle documen
...
+	pSchemaId, err := idx.jobStore.EnqueueJob(job.Job{
+		Dir: modHandle,
+		Func: func(ctx context.Context) error {
+			ctx = exec.WithExecutorFactory(ctx, idx.tfExecFactory)
+			return module.ObtainSchema(ctx, idx.modStore, idx.schemaStore, modHandle.Path())
 		},
-		Type: op.OpTypeParseProviderVersions.String(),
+		IgnoreState: true,
+		Type:        op.OpTypeObtainSchema.String(),
+		DependsOn:   job.IDs{pSchemaVerId},
```
dbanck, line 62: Should we check that `pSchemaVerId` isn't `""` here?

dbanck (changes requested):
Nice work and thanks for the detailed PR description! 👍 

I've found just two minor things

radeksimko:
@dbanck I found a slightly different solution, although the difference is really minor given how rarely (if ever) scheduling of the job would fail - this would have something to do with memdb or the underlying indexes. So if/when that happens, it's more likely we'd be unable to schedule any/all jobs, so most of this conditional logic will likely never come into play.

PTAL

dbanck:
(not meant as part of this PR)

Maybe we could avoid all the checks when creating jobs and `continue` on empty job IDs here:

https://github.com/hashicorp/terraform-ls/blob/a2fa3a048446ec24bbb3086e79553c247cc39f3b/internal/state/jobs.go#L66-L75

dbanck (approved):

radeksimko:
> Maybe we could avoid all the checks when creating jobs and continue on empty job IDs here

Hmm, I like that idea! 🤔

PR #1010, opened 2022-07-22
settings: break down indexing options

radeksimko:
As per our internal conversation yesterday, this allows us to get best of both worlds where VS Code can present nice settings UI and LS isn't limited to a flat structure.

I will raise a **separate PR** to update all three `terraform*` options to follow the same principles as documented and further align it with https://github.com/hashicorp/vscode-terraform/blob/main/docs/settings-migration.md , i.e.

instead of

```json
{
  "terraformExecPath": "",
  "terraformExecTimeout": "",
  "terraformExecLogFilePath": ""
}
```
we will accept
```json
{
  "terraform": {
    "path": ""
    "timeout": ""
    "logFilePath": ""
  }
}
```

and any future (stable, non-experimental) settings can follow these principles.

### DIFF
--- a/docs/SETTINGS.md
+++ b/docs/SETTINGS.md
@@ -1,5 +1,11 @@
 # Settings
 
+## Data Structures
+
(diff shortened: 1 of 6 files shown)


dbanck (approved):
LGTM 👍 I like the approach of handling settings this way

PR #1011, opened 2022-07-22
settings: break down terraform CLI related settings

radeksimko:
~Depends on https://github.com/hashicorp/terraform-ls/pull/1010~

This aligns Terraform CLI related settings with the recently documented principles https://github.com/hashicorp/terraform-ls/blob/main/docs/language-clients.md#configuration

### DIFF
--- a/docs/SETTINGS.md
+++ b/docs/SETTINGS.md
@@ -10,23 +10,39 @@ Clients which expose these config options to the end-user are advised to match t
 
 The language server supports the following configuration options:
 
-## `terraformLogFilePath` (`string`)
+## `terraform` (object `{}`)
+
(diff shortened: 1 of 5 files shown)


dbanck (approved):
LGTM!

I was expecting changes to `settings_test.go`, but there are no test for this.

radeksimko:
> I was expecting changes to settings_test.go, but there are no test for this.

I can't think of a way of testing the settings in a useful and non-fragile way, i.e. that we don't just end up testing the underlying `mapstructure` decoding logic, but I'm open to ideas.

dbanck:
Mhm, I guess the only worth testing might be the validation of `TerraformExecPath`?

jpogran (approved):

PR #1012, opened 2022-07-22
cmd/serve: Remove deprecated `-tf-exec-*` CLI flags

radeksimko:
Depends on:

 - https://github.com/hashicorp/terraform-ls/pull/1010
 - https://github.com/hashicorp/terraform-ls/pull/1011

--- 

These CLI flags were deprecated almost a year ago as part of https://github.com/hashicorp/terraform-ls/blob/main/CHANGELOG.md#0210-23-august-2021 so it's about time to just remove them.

### DIFF
--- a/internal/cmd/serve_command.go
+++ b/internal/cmd/serve_command.go
@@ -6,12 +6,10 @@ import (
 	"fmt"
 	"log"
 	"os"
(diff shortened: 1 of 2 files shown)


dbanck (approved):
👍

PR #1016, opened 2022-07-26
Add terraform version command

jpogran:
This commit adds a new `module.terraform` command which returns the required and discovered version of Terraform in the currrent workspace.

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -196,7 +196,7 @@ Provides information about the terraform binary version for the current module.
 
  - `v` - describes version of the format; Will be used in the future to communicate format changes.
  - `required_version` - Version constraint specified in configuration
- - `discovered_version` - Version discovered from `terraform version --json` using the `uri` given
+ - `discovered_version` - Version discovered from `terraform version --json` in the directory specified in `uri`
 
 ```json
 {
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -183,3 +183,25 @@ installed version.
   }
 }
 ```
+
+### `module.terraform`
+
+Provides information about the terraform binary version for the current module.
+
+**Arguments:**
+
+ - `uri` - URI of the directory of the module in question, e.g. `file:///path/to/network`
+
+**Outputs:**
+
+ - `v` - describes version of the format; Will be used in the future to communicate format changes.
+ - `required_version` - Version constraint specified in configuration
+ - `discovered_version` - Version discovered from `terraform version --json` using the `uri` given
+
+```json
+{
+  "v": 0,
+  "required_version": "~> 0.15",
+  "discovered_version": "1.1.0"
+}
+```
--- a/internal/langserver/handlers/command/terraform.go
+++ b/internal/langserver/handlers/command/terraform.go
@@ -48,17 +48,15 @@ func (h *CmdHandler) TerraformVersionRequestHandler(ctx context.Context, args cm
 		return response, nil
 	}
 
-	progress.Report(ctx, "Module info found ...")
-	if mod.TerraformVersion == nil {
-		return response, nil
+	progress.Report(ctx, "Recording terraform version info ...")
+	if mod.TerraformVersion != nil {
+		response.DiscoveredVersion = mod.TerraformVersion.String()
 	}
-	if mod.Meta.CoreRequirements == nil {
-		return response, nil
+	if mod.Meta.CoreRequirements != nil {
+		response.RequiredVersion = mod.Meta.CoreRequirements.String()
 	}
 
-	progress.Report(ctx, "Recording terraform version info ...")
-	response.DiscoveredVersion = mod.TerraformVersion.String()
-	response.RequiredVersion = mod.Meta.CoreRequirements.String()
+	progress.Report(ctx, "Sending response ...")
 
 	return response, nil
 }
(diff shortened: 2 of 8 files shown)


internal/langserver/handlers/command/terraform.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/hashicorp/terraform-ls/internal/uri"
+)
+
+const terraformVersionRequestVersion = 0
+
+type terraformInfoResponse struct {
+	FormatVersion     int    `json:"v"`
+	RequiredVersion   string `json:"required_version,omitempty"`
+	DiscoveredVersion string `json:"discovered_version,omitempty"`
+	DiscoveredPath    string `json:"discovered_path,omitempty"`
```
radeksimko, line 18:
```suggestion
```
Shall we omit this for now, until we figure out how to persist it and keep it up to date?

internal/langserver/handlers/command/terraform.go
```diff
@@ -0,0 +1,56 @@
...
+	if mod == nil {
+		return response, nil
+	}
+
+	if mod.TerraformVersion == nil {
+		return response, nil
+	}
+	if mod.Meta.CoreRequirements == nil {
+		return response, nil
+	}
```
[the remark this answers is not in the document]
radeksimko: I would say quite the opposite. I think the LS should respond even when it doesn't know either of these (yet). It may take time to obtain the version or decode the requirements.

What's more important however is that we notify the client when this changes - see how we already do this for `module.providers` and `module.calls`.

radeksimko (changes requested):
Aside from my in-line comments, do you mind adding a test for this new command?

Feel free to look into https://github.com/hashicorp/terraform-ls/blob/main/internal/langserver/handlers/execute_command_module_providers_test.go for some inspiration.

jpogran:
> What's more important however is that we notify the client when this changes - see how we already do this for `module.providers` and `module.calls`.

This ticket is targeting https://github.com/hashicorp/vscode-terraform/issues/697, which points to https://github.com/hashicorp/terraform-ls/issues/69 for detecting and tracking when the version changes. Should we tackle both work items in this PR, or just the original ticket?

radeksimko:
Good question!

So I think for the Terraform version in particular, until #69 is addressed, it won't change, but the version _requirements_ which are parsed from the configuration may change at any time - so for those I think we should send the updates.

internal/langserver/handlers/command/terraform.go
```diff
@@ -0,0 +1,62 @@
...
+
+	progress.Report(ctx, "Recording terraform version info ...")
+	if mod.TerraformVersion != nil {
+		response.DiscoveredVersion = mod.TerraformVersion.String()
+	}
+	if mod.Meta.CoreRequirements != nil {
+		response.RequiredVersion = mod.Meta.CoreRequirements.String()
+	}
+
+	progress.Report(ctx, "Sending response ...")
```
radeksimko, line 59: Since this is getting the data from memory, it should be pretty quick (definitely sub-second, low milliseconds in most cases), so I'm not sure there's much value in the progress reporting for this particular command?

radeksimko:
Thank you for adding the test - that looks good - we could possibly deal with the updates in a separate PR, if you wish.

One thing we should still do however is to document this new command here https://github.com/hashicorp/terraform-ls/blob/main/docs/commands.md

jpogran:
Version change notification added and docs updated

internal/langserver/handlers/command/terraform.go
jpogran: The method is a no-op if a token isn't provided in the request, so this technically doesn't do anything yet.

My thinking along this lines is that we enable progress information everywhere as we go, instead of having to add when there's a problem. Then we can dial back if it's too much.

docs/commands.md
```diff
@@ -183,3 +183,25 @@ installed version.
...
+
+**Arguments:**
+
+ - `uri` - URI of the directory of the module in question, e.g. `file:///path/to/network`
+
+**Outputs:**
+
+ - `v` - describes version of the format; Will be used in the future to communicate format changes.
+ - `required_version` - Version constraint specified in configuration
+ - `discovered_version` - Version discovered from `terraform version --json` using the `uri` given
```
radeksimko, line 199:
```suggestion
 - `discovered_version` - Version discovered from `terraform version --json` in the directory specified in `uri`
```

internal/langserver/handlers/service.go
```diff
@@ -463,6 +463,10 @@ func (svc *service) configureSessionDependencies(ctx context.Context, cfgOpts *s
 			moduleHooks = append(moduleHooks, callRefreshClientCommand(svc.server, commandId))
 		}
 
+		if commandId, ok := lsp.ExperimentalClientCapabilities(cc.Experimental).RefreshTerraformVersionCommandId(); ok {
+			moduleHooks = append(moduleHooks, callRefreshClientCommand(svc.server, commandId))
+		}
```
radeksimko, line 468: I think in the long run we'll probably want to create a separate hook which only fires when the data exposed through each command changes (not just any data), but this is 🆗  for now, since the other two refresh commands above have the exact same problem.

radeksimko (approved):
I'm half wondering if the refresh command should be called `refreshModuleTerraform`, to indicate that it's related to the `module.terraform`, which in turn exposes more than just a Terraform version, however I'm also happy to keep it as is while we figure out what exact data and in what exact shape do we want to expose. 😃 

👍🏻

Issue #868, opened 2022-04-11
Introduce hooks for dynamic completion


radeksimko:
## Background

We currently provide completion based entirely on schemas, which are sourced either from providers (as JSON), declared within `terraform-schema`, or augmented from the configuration (such as module inputs/outputs based on variables and outputs).

### Use Cases

There are use cases for a more dynamic completion however

 - https://github.com/hashicorp/vscode-terraform/issues/672
 - https://github.com/hashicorp/vscode-terraform/issues/676

which cannot be served using a static nor purely config-informed schema.

### Future Related Use Cases

The solution may not necessarily be ready to solve any of these but these look like the same problems that _could_ be solved by the same solution:

 - Terraform: Completion of AWS regions or AZs in `provider "aws"` blocks
 - Nomad: completion of Docker `image` from Docker Hub, Quay and/or other image sources https://www.nomadproject.io/docs/drivers/docker
 - Packer: completion of `source_image` from GCP https://www.packer.io/plugins/post-processors/googlecompute/googlecompute-export#basic-example

### `hashicorp/hcl-lang`

 - [x] Introduce`CompletionHooks []lang.CompletionHook` to `schema.AttributeSchema`
 - [x] Introduce `CompletionHooks CompletionFuncMap` to `decoder.DecoderContext`
 - [x] Introduce `CompletionResolveHooks CompletionResolveFuncMap` to `decoder.DecoderContext`
 - [x] Introduce `Meta interface{}` field to `decoder.PathContext` to allow LS to pass extra context (such as which provider/module to pull versions for) to completion hooks
 - [x] Introduce `MetaFromContext(ctx)` to `decoder` to help obtain the metadata from completion hooks in LS
 - [x] Update `CandidatesAtPos()` in `PathDecoder`
   - [x] Add `ctx context.Context` argument (for cancellation)
   - [x] Ensure completion hooks are called if defined within `AttributeSchema`
      - Pass `hcl.Position` via ctx to each hook
      - Pass `filename` via ctx to each hook
      - Pass `lang.Path` via ctx to each hook
 - [x] Introduce `ResolveCandidate()` to `decoder.PathDecoder`

```go
type AttributeSchema struct {
// ...
	// CompletionHooks represents any hooks which provide
	// additional completion candidates for the attribute dynamically
	CompletionHooks []lang.CompletionHook
}
```
```go
type Decoder struct {
// ...
	CompletionHooks        CompletionFuncMap
[...]

PR #1017, opened 2022-07-27
Add hooks for dynamic completion

dbanck:
This PR introduces a new `hooks` package, which we can use to implement hooks for dynamic completion. Hooks will be implemented as part of https://github.com/hashicorp/vscode-terraform/issues/672.

We also introduce a new `completionItem/resolve` handler, which editors can use to update a completion item with more detailed data. The data is obtained via the execution of a resolve hook.

---

* Closes https://github.com/hashicorp/terraform-ls/issues/868
* Requires https://github.com/hashicorp/hcl-lang/pull/123

### DIFF
--- a/internal/decoder/decoder.go
+++ b/internal/decoder/decoder.go
@@ -75,9 +75,11 @@ func varsPathContext(mod *state.Module) (*decoder.PathContext, error) {
 
 func DecoderContext(ctx context.Context) decoder.DecoderContext {
 	dCtx := decoder.DecoderContext{
-		UtmSource:     utm.UtmSource,
-		UtmMedium:     utm.UtmMedium(ctx),
-		UseUtmContent: true,
+		UtmSource:              utm.UtmSource,
+		UtmMedium:              utm.UtmMedium(ctx),
+		UseUtmContent:          true,
+		CompletionHooks:        make(decoder.CompletionFuncMap),
+		CompletionResolveHooks: make(decoder.CompletionResolveFuncMap),
 	}
 
 	cc, err := ilsp.ClientCapabilities(ctx)
--- a/internal/decoder/decoder.go
+++ b/internal/decoder/decoder.go
@@ -74,13 +74,10 @@ func varsPathContext(mod *state.Module) (*decoder.PathContext, error) {
 }
 
 func DecoderContext(ctx context.Context) decoder.DecoderContext {
-	dCtx := decoder.DecoderContext{
-		UtmSource:              utm.UtmSource,
-		UtmMedium:              utm.UtmMedium(ctx),
-		UseUtmContent:          true,
-		CompletionHooks:        make(decoder.CompletionFuncMap),
-		CompletionResolveHooks: make(decoder.CompletionResolveFuncMap),
-	}
+	dCtx := decoder.NewDecoderContext()
+	dCtx.UtmSource = utm.UtmSource
+	dCtx.UtmMedium = utm.UtmMedium(ctx)
+	dCtx.UseUtmContent = true
 
 	cc, err := ilsp.ClientCapabilities(ctx)
 	if err == nil {
--- a/internal/decoder/decoder.go
+++ b/internal/decoder/decoder.go
@@ -15,12 +15,6 @@ import (
 	tfschema "github.com/hashicorp/terraform-schema/schema"
 )
 
-func NewDecoder(ctx context.Context, pathReader decoder.PathReader) *decoder.Decoder {
-	d := decoder.NewDecoder(pathReader)
-	d.SetContext(decoderContext(ctx))
-	return d
-}
-
 func modulePathContext(mod *state.Module, schemaReader state.SchemaReader, modReader ModuleReader) (*decoder.PathContext, error) {
 	schema, err := schemaForModule(mod, schemaReader, modReader)
 	if err != nil {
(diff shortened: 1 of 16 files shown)


internal/decoder/decoder.go
```diff
@@ -79,11 +73,13 @@ func varsPathContext(mod *state.Module) (*decoder.PathContext, error) {
...
+func DecoderContext(ctx context.Context) decoder.DecoderContext {
 	dCtx := decoder.DecoderContext{
-		UtmSource:     utm.UtmSource,
-		UtmMedium:     utm.UtmMedium(ctx),
-		UseUtmContent: true,
+		UtmSource:              utm.UtmSource,
+		UtmMedium:              utm.UtmMedium(ctx),
+		UseUtmContent:          true,
+		CompletionHooks:        make(decoder.CompletionFuncMap),
+		CompletionResolveHooks: make(decoder.CompletionResolveFuncMap),
```
radeksimko, line 82: As mentioned in the other PR, ideally this shouldn't be necessary.

internal/hooks/hooks.go
```diff
@@ -0,0 +1,20 @@
+// Package hooks enables the implementation of hooks for dynamic
+// autocompletion. Hooks should be added to this package and
+// registered in AppendCompletionHooks in completion_hooks.go.
```
radeksimko, line 3:
```suggestion
// registered via AppendCompletionHooks in completion_hooks.go.
```

internal/hooks/hooks.go
```diff
@@ -0,0 +1,20 @@
...
+//
+// A hook must have the following signature:
+//  func (h *Hooks) Name(ctx context.Context, value cty.Value) ([]decoder.Candidate, error)
+// It receives the current value of the attribute and must return
+// a list of completion candidates.
```
radeksimko, line 8: I would argue this is something that a compile-time error would explain and it would likely be even more accurate. As much as I appreciate the commenting efforts, the risk here is that we end up changing the interface upstream but won't remember to update this comment. Compiler will however always bring it up if types/interfaces don't match.

We could just refer to `decoder.CompletionFunc` here and document that one upstream instead.

internal/hooks/hooks.go
```diff
@@ -0,0 +1,20 @@
...
+//
+// All hooks have access to path, filename and pos via context:
+//  path, ok := decoder.PathFromContext(ctx)
+//  filename, ok := decoder.FilenameFromContext(ctx)
+//  pos, ok := decoder.PosFromContext(ctx)
```
radeksimko, line 13: This is very useful context, but I think that should be part of docs for `CompletionFunc` upstream, where we can more easily keep it up to date.

internal/hooks/module_source_local.go
```diff
@@ -0,0 +1,18 @@
+package hooks
+
+import (
+	"context"
+
+	"github.com/hashicorp/hcl-lang/decoder"
+	"github.com/zclconf/go-cty/cty"
+)
+
+func (h *Hooks) LocalModuleSources(ctx context.Context, value cty.Value) ([]decoder.Candidate, error) {
```
radeksimko, line 10: I'm assuming this would be eventually part of a separate PR? 👀

internal/langserver/handlers/completion_resolve.go
```diff
@@ -0,0 +1,55 @@
...
+	"context"
+
+	"github.com/hashicorp/hcl-lang/decoder"
+	"github.com/hashicorp/hcl-lang/lang"
+	ilsp "github.com/hashicorp/terraform-ls/internal/lsp"
+	"github.com/hashicorp/terraform-ls/internal/mdplain"
+	lsp "github.com/hashicorp/terraform-ls/internal/protocol"
+)
+
+type CompletionItem struct {
```
radeksimko, line 13: Nitpick: We could move this under the `protocol` package as we already have some extensions of the original structs there for telemetry or experimental functions:
https://github.com/hashicorp/terraform-ls/tree/main/internal/protocol

internal/lsp/completion.go
```diff
@@ -70,7 +70,12 @@ func toCompletionItem(candidate lang.Candidate, caps lsp.CompletionClientCapabil
 		Documentation:       doc,
 		TextEdit:            textEdit(candidate.TextEdit, snippetSupport),
 		Command:             cmd,
-		AdditionalTextEdits: textEdits(candidate.AdditionalTextEdits, snippetSupport),
+		AdditionalTextEdits: TextEdits(candidate.AdditionalTextEdits, snippetSupport),
+		// TODO set deprecated via `tags` if supported.
```
radeksimko, line 74: Thanks for adding the TODO, is it worth moving it below, where we use the legacy `DeprecatedSupport` bool field?

radeksimko (approved):
LGTM, aside from a few in-line comments, although you may want to look at the upstream `hcl-lang` PR first and then rebase here once merged.

internal/hooks/module_source_local.go
dbanck: I added this as a placeholder to showcase the registration of hooks in `AppendCompletionHooks`. Without any hooks we wouldn't be able to create `hooks.Hooks{}`.

But I can remove it if you want? And defer everything to the next PR

internal/lsp/completion.go
dbanck: I missed below. We're already using tags

internal/protocol/completion.go
```diff
@@ -0,0 +1,9 @@
+package protocol
+
+import "github.com/hashicorp/hcl-lang/lang"
+
+type CompletionItemR struct {
```
radeksimko, line 5:
```suggestion
type CompletionItemWithResolveHook struct {
```
Just for some extra clarity

PR #1021, opened 2022-08-01
state: Introduce `DependsOn` for N-to-1 job dependencies

radeksimko:
## Background

Prior to this PR, the only way we could chain jobs was via `Defer`. This had one unfortunate limit, which is that a job can only depend on exactly one other job, i.e. there was just 1-to-1 dependency.

As part of introducing a new job type for parsing provider versions in https://github.com/hashicorp/terraform-ls/pull/1014 it became clear that the new job will need to rely on 2+ other jobs:

 - parsed files
 - decoded metadata (via earlydecoder) - so that we have `required_providers`
 - parsed module manifest (+ any/all modules decoded) - so that we have `required_providers` of any submodules

I originally attempted to solve it in a hacky way by introducing `WaitForJobs()` inside `Defer`, but that created a few other problems - need to run at least two go routines, so that one can be blocked (waiting) and the other one can be dispatching + I ran into some race conditions, so I binned that hacky approach.

A nice side-effect demonstrated in the 2nd commit is that using `DependsOn` instead of `Defer` makes the code (IMHO) clearer and more readable, as there's less nesting and with the right variable names for job IDs it makes it much more obvious what the relationships are.

There is unfortunately still a few remaining (valid) use cases for `Defer` though - specifically because we cannot schedule any jobs for module paths parsed from a module manifest, until that manifest is parsed - so those few lines of code to do the scheduling still need to run as part of `Defer`.

The only downside of `DependsOn` compared to `Defer` is that the job which we depend on may be long gone/finished by the time the dependent one is dispatched, which means that we don't have direct access to its error/outcome. However I plan to address that problem more holistically as part of https://github.com/hashicorp/terraform-ls/pull/1006 - TL;DR we can still run the jobs and return early if we find out that the data which we were expecting the previous job to provide aren't there.

### DIFF
--- a/internal/indexer/module_calls.go
+++ b/internal/indexer/module_calls.go
@@ -38,6 +38,8 @@ func (idx *Indexer) decodeInstalledModuleCalls(modHandle document.DirHandle) (jo
 		// copy path for queued jobs below
 		mcPath := mc.Path
 
+		refCollectionDeps := make(job.IDs, 0)
+
 		parseId, err := idx.jobStore.EnqueueJob(job.Job{
 			Dir: mcHandle,
 			Func: func(ctx context.Context) error {
@@ -47,31 +49,36 @@ func (idx *Indexer) decodeInstalledModuleCalls(modHandle document.DirHandle) (jo
 		})
 		if err != nil {
 			multierror.Append(errs, err)
-			continue
-		}
-		jobIds = append(jobIds, parseId)
-
-		metaId, err := idx.jobStore.EnqueueJob(job.Job{
-			Dir:  mcHandle,
-			Type: op.OpTypeLoadModuleMetadata.String(),
-			Func: func(ctx context.Context) error {
-				return module.LoadModuleMetadata(idx.modStore, mcPath)
-			},
-			DependsOn: job.IDs{parseId},
-		})
-		if err != nil {
-			multierror.Append(errs, err)
-			continue
 		} else {
-			jobIds = append(jobIds, metaId)
+			jobIds = append(jobIds, parseId)
+			refCollectionDeps = append(refCollectionDeps, parseId)
 		}
 
-		ids, err := idx.collectReferences(mcHandle, job.IDs{parseId, metaId})
-		if err != nil {
-			multierror.Append(errs, err)
-			continue
-		} else {
-			jobIds = append(jobIds, ids...)
+		var metaId job.ID
+		if parseId != "" {
+			metaId, err = idx.jobStore.EnqueueJob(job.Job{
+				Dir:  mcHandle,
+				Type: op.OpTypeLoadModuleMetadata.String(),
+				Func: func(ctx context.Context) error {
+					return module.LoadModuleMetadata(idx.modStore, mcPath)
+				},
+				DependsOn: job.IDs{parseId},
+			})
+			if err != nil {
+				multierror.Append(errs, err)
+			} else {
+				jobIds = append(jobIds, metaId)
+				refCollectionDeps = append(refCollectionDeps, metaId)
+			}
+		}
+
+		if parseId != "" {
+			ids, err := idx.collectReferences(mcHandle, refCollectionDeps)
+			if err != nil {
+				multierror.Append(errs, err)
+			} else {
+				jobIds = append(jobIds, ids...)
+			}
 		}
 
 		varsParseId, err := idx.jobStore.EnqueueJob(job.Job{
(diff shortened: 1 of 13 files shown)


internal/indexer/walker.go
```diff
@@ -16,72 +16,60 @@ func (idx *Indexer) WalkedModule(ctx context.Context, modHandle document.DirHand
...
-				},
-			})
-			if err != nil {
-				return ids, err
-			} else {
-				ids = append(ids, id)
-			}
+	})
+	if err != nil {
+		errs = multierror.Append(errs, err)
```
dbanck, line 27: Are we able to continue in the case of an error here? We're using `parseId` in the next job's `DependsOn`, but `parseId` is likely to be `""`.

dbanck (changes requested):
I really like the reduced nesting when enqueuing jobs!

I _think_, I've found two things which need further inspection

internal/indexer/walker.go
radeksimko: Good catch - passing empty string to `DependsOn` is certainly unwanted here! 🙈

internal/state/jobs.go
```diff
@@ -396,6 +431,36 @@ func (js *JobStore) FinishJob(id job.ID, jobErr error, deferredJobIds ...job.ID)
...
+		idx, ok := idIsInSlice(sJob.DependsOn, id)
+		if ok {
+			jobCopy := sJob.Copy()
+
+			// remove found ID from DependsOn
+			jobCopy.DependsOn[idx] = jobCopy.DependsOn[len(jobCopy.DependsOn)-1]
+			jobCopy.DependsOn = jobCopy.DependsOn[:len(jobCopy.DependsOn)-1]
+
+			// re-insert updated data
+			_, err := txn.DeleteAll(js.tableName, "id", id)
```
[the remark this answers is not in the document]
radeksimko: Of course, I wonder how did the tests not catch this - thank you for spotting it!

internal/indexer/module_calls.go
```diff
@@ -11,104 +11,111 @@ import (
...
-				continue
+			} else {
+				jobIds = append(jobIds, metaId)
+				refCollectionDeps = append(refCollectionDeps, metaId)
 			}
-			jobIds = append(jobIds, id)
+		}
 
-			id, err = idx.jobStore.EnqueueJob(job.Job{
+		if parseId != "" {
```
dbanck, line 75: Do we need module metadata before collecting references? Should we check for `&& metaId != ""` here?

internal/indexer/module_calls.go
```diff
@@ -11,104 +11,111 @@ import (
...
-					ids = append(ids, id)
-					return ids, err
+					return module.DecodeVarsReferences(ctx, idx.modStore, idx.schemaStore, mcPath)
 				},
+				Type:      op.OpTypeDecodeVarsReferences.String(),
+				DependsOn: job.IDs{varsParseId},
 			})
 			if err != nil {
 				multierror.Append(errs, err)
 				continue
```
dbanck, line 108:
```suggestion
```

With the introduction of the `else` branch, there is nothing to skip.

internal/indexer/walker.go
```diff
@@ -111,44 +110,39 @@ func (idx *Indexer) WalkedModule(ctx context.Context, modHandle document.DirHand
...
-	if err != nil {
-		return ids, err
-	}
-
-	rIds, err := idx.collectReferences(ctx, modHandle)
-	if err != nil {
-		errs = multierror.Append(errs, err)
-	} else {
-		ids = append(ids, rIds...)
+	if parseId != "" {
```
dbanck, line 139: Do we need module metadata before collecting references? Should we check for `&& metaId != ""` here?

internal/indexer/module_calls.go
```diff
@@ -11,104 +11,111 @@ import (
...
+			} else {
+				jobIds = append(jobIds, metaId)
+				refCollectionDeps = append(refCollectionDeps, metaId)
 			}
-			jobIds = append(jobIds, id)
+		}
 
-			id, err = idx.jobStore.EnqueueJob(job.Job{
+		if parseId != "" {
+			ids, err := idx.collectReferences(mcHandle, refCollectionDeps)
```
dbanck, line 76: I didn't notice this before: why are we calling `collectReferences` here, before parsing variables? While in the walker, we're calling it much later?

dbanck (commented):
Thank you for fixing/addressing my concerns!

I'm still a bit unsure about the order/dependencies of some jobs.

internal/indexer/module_calls.go
radeksimko: It can certainly enrich the references (e.g. because it's used to inform module inputs/outputs). The parsing is however essential and we'd always produce zero references without it. Bear in mind the meaning of this check isn't to draw the dependency, but to allow the reference collection to run. If it does run and the metadata job was successfully scheduled, then it will depend on it too.

internal/indexer/walker.go
radeksimko: Same answer as above - strictly speaking we don't, but it's useful if we have them. This logic basically applies progressive enhancement principle in that we still collect references, even if scheduling of metadata has failed, there will just likely be less references collected as a result.

internal/indexer/module_calls.go
radeksimko: The ordering of _scheduling_ is not significant - the `*.tfvars` parsing and `*.tf` references should be independent of each other so it doesn't matter in what order do they run - we just happen to schedule them in different order in the two places, `DependsOn` and `Defer` (if any) decide in what order they actually get executed.

dbanck: Okay! Thanks for the explanation!

dbanck (approved):

internal/indexer/module_calls.go
radeksimko: We could align the ordering of scheduling, but there are also other differences - e.g. that when we parse submodules here, we assume that the modules are not initialised by themselves (given the context) and don't attempt to parse module manifest or obtain schema - both of which would otherwise be two more jobs which reference collection should depend on.

PR #1023, opened 2022-08-02
Improve TF registry client

dbanck:
This PR fixes the sorting of registry module versions. Before we would return the oldest instead of the newest version.

And passes context to the registry client for cancellation purposes.

### DIFF
--- a/internal/registry/registry.go
+++ b/internal/registry/registry.go
@@ -1,9 +1,11 @@
 package registry
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"sort"
 	"time"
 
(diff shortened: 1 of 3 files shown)


radeksimko (approved):
It would be nice to test the context patch by passing a context with short timeout (e.g. 50 ms) and adding a `time.Sleep` into the test server which exceeds the timeout.

Similarly you could test the sorting with a test server which returns versions in (wrong) reverse order.

Either way thank you for finding those bugs and patching both!

dbanck:
> It would be nice to test the context patch by passing a context with short timeout (e.g. 50 ms) and adding a `time.Sleep` into the test server which exceeds the timeout.

Good idea! I'll add a test for that.

> Similarly you could test the sorting with a test server which returns versions in (wrong) reverse order.

`moduleVersionsMockResponse` contains an unsorted list of versions and [the change to the test ](https://github.com/hashicorp/terraform-ls/pull/1023/files#diff-d6eaf470eeca03f039f802dc831d89061a557dbe660e1993e7cce8d4cc8b61cbR130) should test that behavior. Before the update to `sort` we would get `0.0.7` for those constraints. Now we're getting `0.0.8` as expected.

radeksimko:
> moduleVersionsMockResponse contains an unsorted list of versions and [the change to the test ](https://github.com/hashicorp/terraform-ls/pull/1023/files#diff-d6eaf470eeca03f039f802dc831d89061a557dbe660e1993e7cce8d4cc8b61cbR130)should test that behavior. Before the update to sort we would get 0.0.7 for those constraints. Now we're getting 0.0.8 as expected. I found this to be the easiest way without introducing a lot of new mock data.

Aha, I didn't realise that was the motivation/effect of that particular change - makes sense! 😄

PR #1024, opened 2022-08-02
Complete module `source` and `version`

dbanck:
Closes https://github.com/hashicorp/vscode-terraform/issues/672

--- 

This PR adds three hooks for completing module `source` and `version`.

### `CompleteLocalModuleSources`

Uses the local indexed modules and offers relative path suggestions, when completing module sources.

### `CompleteRegistryModuleSources`

Searches the Terraform registry modules via Algolia and fetches the full source address and module description.

For this to work, we now need to pass `main.algoliaAppID` and `main.algoliaAPIKey` via `ldflags` on build.

### `CompleteRegistryModuleVersions`

Fetches the available module versions from the Terraform registry.

## Deferred work

Completion of required inputs was moved to #1029

### DIFF
--- a/internal/cmd/serve_command.go
+++ b/internal/cmd/serve_command.go
@@ -23,6 +23,9 @@ type ServeCommand struct {
 	Ui      cli.Ui
 	Version string
 
+	AlgoliaAppID  string
+	AlgoliaAPIKey string
+
 	// flags
 	port           int
 	logFilePath    string
@@ -97,6 +100,7 @@ func (c *ServeCommand) Run(args []string) int {
 	logger.Printf("Starting terraform-ls %s", c.Version)
 
 	ctx = lsctx.WithLanguageServerVersion(ctx, c.Version)
+	ctx = lsctx.WithAlgoliaCredentials(ctx, c.AlgoliaAppID, c.AlgoliaAPIKey)
 
 	srv := langserver.NewLangServer(ctx, handlers.NewSession)
 	srv.SetLogger(logger)
--- a/internal/cmd/serve_command.go
+++ b/internal/cmd/serve_command.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"syscall"
 
+	"github.com/hashicorp/terraform-ls/internal/algolia"
 	lsctx "github.com/hashicorp/terraform-ls/internal/context"
 	"github.com/hashicorp/terraform-ls/internal/langserver"
 	"github.com/hashicorp/terraform-ls/internal/langserver/handlers"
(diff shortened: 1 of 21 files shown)


internal/hooks/module_source_registry.go
```diff
@@ -0,0 +1,86 @@
...
+		return candidates, nil
+	}
+
+	if h.AlgoliaClient == nil {
+		return candidates, nil
+	}
+
+	modules, err := h.fetchModulesFromAlgolia(ctx, prefix)
+	if err != nil {
+		return candidates, err
```
radeksimko, line 57: Given that we ignore these errors from hooks upstream https://github.com/hashicorp/hcl-lang/blob/b66451713db4e6e3bcf6cde7d70bd8cb51cc2729/decoder/expression_candidates.go#L278

is it worth logging it here? e.g. by plumbing logger to `Hooks` and using it here.

radeksimko (changes requested):
It works great for me! I just left some implementation-related questions and suggestions in-line.

internal/cmd/serve_command.go
```diff
@@ -97,6 +100,7 @@ func (c *ServeCommand) Run(args []string) int {
 	logger.Printf("Starting terraform-ls %s", c.Version)
 
 	ctx = lsctx.WithLanguageServerVersion(ctx, c.Version)
+	ctx = lsctx.WithAlgoliaCredentials(ctx, c.AlgoliaAppID, c.AlgoliaAPIKey)
```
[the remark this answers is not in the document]
dbanck: Good catch! I'm going with `&&` instead of `||`, so both values have to be set

internal/hooks/module_version_test.go
```diff
@@ -0,0 +1,124 @@
...
+		},
+		{
+			Label:         `"0.0.1"`,
+			Kind:          lang.StringCandidateKind,
+			RawInsertText: `"0.0.1"`,
+			SortText:      "  2",
+		},
+	}
+
+	candidates, _ := h.RegistryModuleVersions(ctx, cty.StringVal(""))
```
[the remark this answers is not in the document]
dbanck: Exactly! We're always getting the full list of versions and are not using the prefix in the hook.

internal/langserver/handlers/completion_hooks.go
```diff
@@ -1,15 +1,25 @@
...
-		ModStore: s.modStore,
+		ModStore:       s.modStore,
+		RegistryClient: s.registryClient,
+		Logger:         s.logger,
 	}
 
-	ctx.CompletionHooks["CompleteLocalModuleSources"] = h.LocalModuleSources
+	credentials, ok := algolia.CredentialsFromContext(s.srvCtx)
+	if ok {
+		h.AlgoliaClient = search.NewClient(credentials.AppID, credentials.APIKey)
```
radeksimko, line 19: Nitpick: I suppose we could also avoid registering the completion hook if the credentials aren't available?

radeksimko (approved):
LGTM :shipit: 

Just one minor (non-blocking) note.

internal/langserver/handlers/completion_hooks.go
dbanck: Interesting idea. I think the code looks a bit cleaner if we always register the hook and have it exiting early. So it's up to the hook to decide if all requirements are met for a run.

PR #1027, opened 2022-08-04
Pick core schema only based on `required_version` constraint

radeksimko:
Depends on https://github.com/hashicorp/terraform-schema/pull/129

Closes https://github.com/hashicorp/terraform-ls/issues/993

## UX Impact

While the primary motivation behind the change was performance, it brings a minor UX impact.

### IntelliSense with too _old_ version (e.g. 0.10)

**Before**: No IntelliSense available, most schema-dependent jobs providing data for IntelliSense fail, as per [errors in log](https://gist.github.com/radeksimko/0fb6825951bb316deef9ed4b43b306b5).

Screenshot 2022-08-08 at 12 02 43

**After**: 0.12-based IntelliSense provided, no errors in log.

Screenshot 2022-08-08 at 12 18 26

--- 

The UX for too new version (e.g. 1.5.0) and for when the version is not known (yet) remains practically the same, it is just hopefully more obvious from the changed code combined with the upstream tfschema changes.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,7 @@ require (
 	github.com/hashicorp/terraform-exec v0.17.2
 	github.com/hashicorp/terraform-json v0.14.0
 	github.com/hashicorp/terraform-registry-address v0.0.0-20220623143253-7d51757b572c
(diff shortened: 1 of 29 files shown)


dbanck (approved):
Nice work! 👍

PR #1036, opened 2022-08-11
docs: Update benchmarks

radeksimko:
Not something I'm particularly happy about seeing, but at least it reflects the current state and makes it more obvious that we'll need to work on reducing the memory usage.


dbanck (approved):
Oh, wow! I assume continued development on each module also leads to increased memory usage.

radeksimko:
> I assume continued development on each module also leads to increased memory usage.

I would hope not too significantly, but frankly we don't measure that, so I don't know.

radeksimko:
I think the biggest notable differences are caused by the fact that we simply didn't index the whole module hierarchy before and some modules do have pretty deep one (like the GKE module), so instead of 1 module we now store data for many more.

PR #1044, opened 2022-08-19
docs: Add link to post explaining vim plugin installation

radeksimko:
Not every reader may be familiar with the vim plugin system and I _think_ the linked post explains it quite well, in fact better than the [official manual](https://github.com/vim/vim/blob/03c3bd9fd094c1aede2e8fe3ad8fd25b9f033053/runtime/doc/repeat.txt#L515), but I may be biased as a non-seasoned vim user. 🤷🏻‍♂️

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -123,7 +123,7 @@ inoremap <silent><expr> <c-space> coc#refresh()
 
 ### vim-lsp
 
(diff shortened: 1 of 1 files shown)


dbanck (approved):
There are more ways and different plugin managers for vim. But I think the link captures the most important information.

Issue #445, opened 2021-03-22
Address flaky exit handler test


radeksimko:
The following test failed on 5abae0c

```
--- FAIL: TestExit (0.00s)
    exit_test.go:20: Expected service stop function to be called
FAIL
```

Probably because receiving/processing the request takes a moment and is done in a separate goroutine of upstream library.

PR #1049, opened 2022-08-25
fix: flaky exit handler test

dbanck:
This PR fixes the flaky exit handler test. The sleep makes sure that we account for the asynchronous nature of LS notifications.

I think using `time.Sleep` isn't the best, but a very pragmatic solution here. 

### Before
```
$ go test -timeout 30s -run '^TestExit$' github.com/hashicorp/terraform-ls/internal/langserver/handlers -count=1000
--- FAIL: TestExit (0.00s)
    exit_test.go:20: Expected service stop function to be called
--- FAIL: TestExit (0.00s)
    exit_test.go:20: Expected service stop function to be called
FAIL
FAIL    github.com/hashicorp/terraform-ls/internal/langserver/handlers  1.807s
FAIL
```

### After
```
$ go test -timeout 30s -run '^TestExit$' github.com/hashicorp/terraform-ls/internal/langserver/handlers -count=1000
ok      github.com/hashicorp/terraform-ls/internal/langserver/handlers  13.512s
```

Closes https://github.com/hashicorp/terraform-ls/issues/445

### DIFF
--- a/internal/langserver/handlers/exit_test.go
+++ b/internal/langserver/handlers/exit_test.go
@@ -2,6 +2,7 @@ package handlers
 
 import (
 	"testing"
+	"time"
 
 	"github.com/hashicorp/terraform-ls/internal/langserver"
 )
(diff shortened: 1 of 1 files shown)


radeksimko (approved):
I agree this is pragmatic. I assume that we'd have to rework the E2E test framework built on top of jrpc2 for any hope of a better solution.

There is https://pkg.go.dev/github.com/creachadair/jrpc2/server#NewLocal which we could consider, but ironically the `exit` handler was one of the few reasons I avoided it in the past as it made it hard to access the internal state to verify that a SIGTERM signal was received and method was called.

Maybe we can either avoid testing this particular handler, or use different approach. 🤷🏻

Issue #74, opened 2020-04-29
Support UNC paths on Windows


paultyng:
This is an extension of the work in #73.

PR #1057, opened 2022-08-30
Handle Invalid WSL URIs

### LINKED ISSUE #656: Better guide WSL users opening UNC `\\wsl$\` paths
### Current Version

```
0.22.0
```

### Use-cases

Currently the language server doesn't support UNC paths as per https://github.com/hashicorp/terraform-ls/issues/74

However even when it does, users will continue to be unable to use WSL paths.

Generally speaking WSL users can (and should) use the Linux path (such as `/mnt/c/...`) directly, and most do.

However, the following attempt

```
code "\\wsl$\Ubuntu-20.04\home\example"
```
(assuming the Linux instance for WSL is still running somewhere on the background)

leads to an error message such as

> URI "file://wsl%24/Ubuntu-20.04/home/example" is not valid

### Proposal

The hostname in a WSL URI uses the dollar sign (`$`), which is passed escaped as `%24` by the language client. This trips up the Go URL parser, which is otherwise capable of unescaping the path, but clearly not the hostname.

See https://play.golang.org/p/ROjLYnDbDjR

We can attempt to unescape the whole URL somehow and just make WSL UNC paths work and/or return a more helpful error message to nudge users to use the native Linux path. Additionally on VS Code we could recommend installation of the WSL extension https://code.visualstudio.com/docs/remote/wsl

jpogran:
This adds another piece of logic to attempt to detect UNC WSL Uri as the workspace to open, and return a more informative error so the client can handle prompting the user.

The UNC WSL Uri being passed is in the form of `\\wsl$\Ubuntu\home\james\some\path`. If the user wants to edit files in the WSL instance, they should be opening the workspace in the Remote WSL extension.

When the VS Code sends the path to terraform-ls, it is urlencoded as `\\wsl%24\Ubuntu\home\james\some\path`. This fails `uri.IsURIValid`, which is incorrect because it is a valid path that needs to be decoded. Even if it is decoded, further on in the `uri` package we make assumptions about file paths that ignore the server portion of a UNC path, resulting in terraform-ls attempting to use `Ubuntu\home\james\some\path`. A future PR can address that.

Works with, but not required, https://github.com/hashicorp/vscode-terraform/pull/1219

Closes #656

### DIFF
--- a/internal/langserver/handlers/initialize.go
+++ b/internal/langserver/handlers/initialize.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/creachadair/jrpc2"
+	"github.com/creachadair/jrpc2/code"
 	lsctx "github.com/hashicorp/terraform-ls/internal/context"
 	"github.com/hashicorp/terraform-ls/internal/document"
 	ilsp "github.com/hashicorp/terraform-ls/internal/lsp"
@@ -132,22 +133,23 @@ func (svc *service) Initialize(ctx context.Context, params lsp.InitializeParams)
 		}
 	} else {
 		rootURI := string(params.RootURI)
+
+		invalidUriErr := jrpc2.Errorf(code.InvalidParams,
+			"Unsupported or invalid URI: %q "+
+				"This is most likely client bug, please report it.", rootURI)
+
+		if uri.IsWSLURI(rootURI) {
+			properties["root_uri"] = "invalid"
+			// For WSL URIs we return additional error data
+			// such that clients (e.g. VS Code) can provide better UX
+			// and nudge users to open in the WSL Remote Extension instead.
+			return serverCaps, invalidUriErr.WithData("INVALID_URI_WSL")
+		}
+
 		if !uri.IsURIValid(rootURI) {
 			properties["root_uri"] = "invalid"
 
-			if uri.IsURLEncodedPath(rootURI) {
-				uri, err := uri.UnEncodeURI(rootURI)
-				if err == nil {
-					if uri.Scheme == "file" && uri.Host == "wsl$" {
-						// panic("foo")
-						return serverCaps, fmt.Errorf("Unsupported WSLPATH: %q "+
-							"This is most likely bug, please report it.", rootURI)
-					}
-				}
-			}
-
-			return serverCaps, fmt.Errorf("Unsupported or invalid URI: %q "+
-				"This is most likely bug, please report it.", rootURI)
+			return serverCaps, invalidUriErr
 		}
 
 		err := svc.setupWalker(ctx, params, cfgOpts)
--- a/internal/langserver/handlers/initialize.go
+++ b/internal/langserver/handlers/initialize.go
@@ -134,6 +134,18 @@ func (svc *service) Initialize(ctx context.Context, params lsp.InitializeParams)
 		rootURI := string(params.RootURI)
 		if !uri.IsURIValid(rootURI) {
 			properties["root_uri"] = "invalid"
+
+			if uri.IsURLEncodedPath(rootURI) {
+				uri, err := uri.UnEncodeURI(rootURI)
+				if err == nil {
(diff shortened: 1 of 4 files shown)


internal/langserver/handlers/initialize.go
```diff
@@ -134,6 +134,18 @@ func (svc *service) Initialize(ctx context.Context, params lsp.InitializeParams)
...
+					if uri.Scheme == "file" && uri.Host == "wsl$" {
+						// panic("foo")
+						return serverCaps, fmt.Errorf("Unsupported WSLPATH: %q "+
+							"This is most likely bug, please report it.", rootURI)
+					}
+				}
+			}
+
 			return serverCaps, fmt.Errorf("Unsupported or invalid URI: %q "+
 				"This is most likely bug, please report it.", rootURI)
```
radeksimko, line 150:
```suggestion
			invalidUriErr := jrpc2.Errorf(code.InvalidParams,
				"Unsupported or invalid URI: %q "+
					"This is most likely client bug, please report it.", rootURI)

			if uri.IsWSLURI(rootURI) {
				// For WSL URIs we return additional error data
				// such that clients (e.g. VS Code) can provide better UX
				// and nudge users to open the Linux path instead.
				return serverCaps, invalidUriErr.WithData("INVALID_URI_WSL")
			}

			return serverCaps, invalidUriErr
```

The `Data` field can be anything per LSP - I assume that string is easy to deal with on client-side and also easy to build here, but feel free to change it to anything else if you feel it's more convenient.

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#responseError

I also proposed updating of the error for the generic case when the URI is not WSL but it's still invalid, as I think `InvalidParams` is more appropriate/relevant than the default `ServerError`.

internal/uri/uri.go
```diff
@@ -69,6 +69,32 @@ func IsURIValid(uri string) bool {
...
+	}
+
+	parsed, err := url.ParseRequestURI(unescapedPath)
+	if err != nil {
+		return nil, err
+	}
+
+	return parsed, nil
+}
+
```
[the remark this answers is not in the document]
radeksimko: Do you mind adding a test for this? I imagine 2-3 cases should suffice:
1. improperly escaped URI
2. WSL URI
3. non-WSL URI

radeksimko (changes requested):
I left suggestions in-line.

I reckon we'll also need to update this test https://github.com/hashicorp/terraform-ls/blob/2e881babf52296ec6fcf6b24ed2b3180a3b71edc/internal/langserver/handlers/initialize_test.go#L98-L117

We could also add a similar one with a WSL URI but it's probably not worth it since we're comparing error type only anyway without the extra Data field.

internal/uri/uri_test.go
```diff
@@ -178,3 +178,36 @@ func TestMustParseURI(t *testing.T) {
...
 	}
 }
+
+func TestIsWSLURI(t *testing.T) {
+	type args struct {
+		uri string
+	}
+	tests := []struct {
+		name string
+		args args
```
radeksimko, line 188:
```suggestion
		args string
```
Is the extra struct here necessary, since it's just one argument and a string?

radeksimko (approved):
Just one minor question/suggestion regarding the test, merge at will.

One thing I was aiming to achieve with the extra `Data` field was that we'd be able to compare it more easily with just `==` or `===` on the client side. Therefore I don't fully understand the motivation behind `startsWith` method [here](https://github.com/hashicorp/vscode-terraform/pull/1219/files#diff-bd14b59b887f26b70c6f4c93f439f7fca0ca4cdf688fee62a47a4f647a52cd47R183), but we can discuss it in that PR rather than here.

> Even if it is decoded, further on in the uri package we make assumptions about file paths that ignore the server portion of a UNC path, resulting in terraform-ls attempting to use Ubuntu\home\james\some\path. A future PR can address that.

Yep - we can address that as part of https://github.com/hashicorp/terraform-ls/issues/74

radeksimko (approved):

PR #1048, opened 2022-08-24
Fix panic on obtaining provider schemas

dbanck:
This PR fixes the LS panic reported in https://github.com/hashicorp/vscode-terraform/issues/1213.

<details><summary>Stack trace:</summary>

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x13c4898]

goroutine 38 [running]:
github.com/hashicorp/go-version.prereleaseCheck(0xc09a8067e0, 0x3)
github.com/hashicorp/go-version@v1.6.0/constraint.go:198 +0x18
github.com/hashicorp/go-version.constraintGreaterThanEqual(0xc09a8067e8, 0xc0aa07f858)
github.com/hashicorp/go-version@v1.6.0/constraint.go:250 +0x25
github.com/hashicorp/go-version.(*Constraint).Check(...)
github.com/hashicorp/go-version@v1.6.0/constraint.go:163
github.com/hashicorp/go-version.Constraints.Check({0xc0cb192370, 0x1, 0xc}, 0x170c841)
github.com/hashicorp/go-version@v1.6.0/constraint.go:93 +0x5e
github.com/hashicorp/terraform-ls/internal/state.(*ProviderSchemaStore).schemaExists(0xc0007ed6c0, {{0xc0c90cacfa, 0x6}, {0xc0c90cacf0, 0x9}, {0x1714198, 0x15}}, {0xc0cb192370, 0x1, 0x1})
github.com/hashicorp/terraform-ls/internal/state/provider_schema.go:227 +0x9c9
github.com/hashicorp/terraform-ls/internal/state.(*ProviderSchemaStore).AllSchemasExist(0xc0008924c0, 0xc0604fc2a7)
github.com/hashicorp/terraform-ls/internal/state/provider_schema.go:197 +0xef
github.com/hashicorp/terraform-ls/internal/terraform/module.ObtainSchema({0x1d64de0, 0xc053b1f980}, 0x163d6a0, 0xc0bbe71c60, {0xc0604fc2a7, 0x64})
github.com/hashicorp/terraform-ls/internal/terraform/module/module_ops.go:128 +0xea
github.com/hashicorp/terraform-ls/internal/indexer.(*Indexer).WalkedModule.func8({0x1d64de0, 0xc053b1f950})
github.com/hashicorp/terraform-ls/internal/indexer/walker.go:135 +0xc5
[...]

PR #1060, opened 2022-09-01
fix: Add missing provider schema versions

radeksimko:
This is a follow-up to a recent bug fix - https://github.com/hashicorp/terraform-ls/pull/1048

The linked PR addressed the fact that we may sometimes get provider schemas with `nil` versions. This PR additionally ensures that versions are in fact stored as they should be and clarifies some related edge cases in attached comments.

The temporary lack of versions for schemas is not as easily noticeable thanks to the loose/optimistic matching of schemas we perform, where even if we don't match the version, we always match based on the address, or even just provider name alone.

The only difference it makes is when there are two schemas for the same provider address of different versions - so that inaccuracy is being addressed.

https://github.com/hashicorp/terraform-ls/blob/0f745e191e738796628db49b1e962ec881eb6982/internal/state/provider_schema.go#L264-L333

Additionally - to make testing easier, I also added ranking of schemas based on version constraint match, further making schema matching more accurate.

### DIFF
--- a/internal/langserver/handlers/execute_command_terraform_version_test.go
+++ b/internal/langserver/handlers/execute_command_terraform_version_test.go
@@ -46,7 +46,7 @@ func TestLangServer_workspaceExecuteCommand_terraformVersion_basic(t *testing.T)
 		t.Fatal(err)
 	}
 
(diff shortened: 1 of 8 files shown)


dbanck (approved):
LGTM! Thanks for getting to the root cause

Issue #990, opened 2022-07-06
Lazy-load embedded provider schemas to improve performance


radeksimko:
## Background

As documented in https://github.com/hashicorp/terraform-ls/blob/main/docs/benchmarks.md#memory-usage we currently load 1 big blob of JSON with all ~200 providers into memory, which makes up the majority of the ~300MB of memory.

https://github.com/hashicorp/terraform-ls/blob/81b49a174bd8f34a78ab959f7ab88325177b1daf/internal/schemas/gen/gen.go#L163-L180

Aside from memory usage, it also impacts the `initialize` **response time** - i.e. time it takes for the server to become ready for other requests. It currently takes **90+% of the 2 seconds**.

https://github.com/hashicorp/terraform-ls/blob/4308af82997298fc051f7f9efaf4866aab60178e/internal/langserver/handlers/service.go#L510-L513

Related:
 - https://github.com/hashicorp/terraform-ls/issues/986 and similar issues people reported about memory usage before
 - https://github.com/hashicorp/terraform-ls/issues/517
 - https://github.com/hashicorp/terraform-ls/issues/506
 - https://github.com/hashicorp/terraform-ls/issues/645
 - https://github.com/hashicorp/terraform-ls/issues/828
 - https://github.com/hashicorp/terraform-ls/issues/193
 - https://github.com/hashicorp/terraform-ls/issues/150

## Proposal

There's no exact proposal yet, but there are early ideas we have discussed within the team:

1. breaking down the single JSON into 1 per provider
2. lazily-loading the schema into memdb, i.e. loading it only if/when we parse the relevant provider constraint
3. distributing JSON files alongside the LS binary, as opposed to embedding files into it

Each area comes with some different trade-offs which may need to be considered alongside the issues referenced above.

radeksimko:
Based on various sources and experimenting it appears that:

 - any files embedded via `embed` are only allocated into memory when actually used/needed
 - the memory allocated as a result of this _does_ get garbage collected

In my [experiment](https://gist.github.com/radeksimko/ab710002f1928854e487015e9b389be5), I tried to load ~70MiB JSON file with schemas into memdb and then delete it. This resulted in allocation of 198MiB, which dropped back to zero after garbage collection. The garbage collector acted automatically in ~2mins after deletion, but I was also able to force it via `runtime.GC()`.

[...]

PR #1071, opened 2022-09-13
schemas: Break down schema generation

radeksimko:
Closes https://github.com/hashicorp/terraform-ls/issues/990

This addresses a few different problems, as originally outlined in the linked issue.

## Runtime Memory Usage

As tested on macOS (M1 Pro)

**Previously** (v0.29.2) empty config: ~572 MB
**Previously** (v0.29.2) config with `hashicorp/random`: ~572 MB
**Previously** (v0.29.2) config with `hashicorp/aws`: ~572 MB

**After** (this PR) empty config: ~9.8 MB
**After** (this PR) config with `hashicorp/random`: ~12 MB
**After** (this PR) config with `hashicorp/aws`: ~70 MB

## Launch Time

**Previously** (v0.29.2) `initialize` request/response time: ~2 s
**After** (this PR) `initialize` request/response time: 1-3 ms

This is a result of lazily loading each provider only when it's necessary (when we encounter relevant provider requirement). The cost of "lazy-loading" a provider at runtime can be anywhere between <1ms (e.g. `hashicorp/random`) and ~130ms (e.g. `hashicorp/aws`) which should still be fast enough for the user to not notice a difference.

### Time to generate schemas

As measured in the GitHub Actions CI (Ubuntu latest)

**Previously** ([recent CI run](https://github.com/hashicorp/terraform-ls/actions/runs/3113650108/jobs/5048547547#step:6:1)) 11-12 minutes
**After** ([this PR](https://github.com/hashicorp/terraform-ls/actions/runs/3134759412/jobs/5089670113#step:6:1)) 3 minutes

This is mostly a result of running `init` for each provider individually and doing so in parallel. Terraform itself does not parallelise provider installation as part of `terraform init`.

### Platform-Specific Provider Ignorelist

The schema generation, which runs as part of CI for every release and PR and can also run locally, is subject to OS/arch requirements. Previously we maintained a long list of providers which were known to be unavailable for certain platforms, just to make generation work on these platforms. Such a list may easily get outdated the moment it is committed as the Registry is the only source of truth (i.e. providers may release compatible artifacts the next day after we put it on the ignorelist).

This PR addresses the problem by breaking down the generation, such that we don't run `init` for a single giant config with all providers, but we first poke the Registry API and filter out any providers which we know are not available for the platform where we're running.

### Broken Provider Installations

In order to obtain schema for each provider, we first have to install it. The process of installation can break at any time for a few reasons, of which the most commons ones are:

 - Provider maintainers manually yank or change artifacts or break release process in some other way
 - GitHub outage
 - Issues on the network path between GitHub Actions environment and GitHub Releases hosting the artifacts
 - Terraform Registry outage

This PR addresses most of these problems by retrying `init` for each provider individually (5 times, with 2 seconds backoff in between).

Additionally we also now treat even the final failures (after 5 unsuccessful attempts) as soft failures, which just cause schema for the provider to not be included, but the CI to succeed and schema embedding to still work. There is a downside to this: There could be a widespread temporary outage causing us to skip _all_ providers at release time and releasing LS which has no embedded schemas. This could be mitigated by careful observation and potential follow-up release when outage is resolved.

We could also add some checks to verify that e.g. 80-90% of providers were installed correctly, but I'd prefer to leave this for another PR.


jpogran (approved):
Works as advertised on my machine, with reduced mem usage and increased speed!

internal/schemas/gen/gen.go
```diff
@@ -51,273 +50,420 @@ func main() {
...
+// retryInit runs "terraform init" and attempts to retry
+// on known (typically network-related) transient errors
+func retryInit(ctx context.Context, tf *tfexec.Terraform, fullName string, retried int) (time.Duration, error) {
+	maxRetries := 5
+	backoffPeriod := 2 * time.Second
+
+	startTime := time.Now()
+	err := tf.Init(ctx, tfexec.Upgrade(true))
+	if err != nil {
+		if retried >= maxRetries {
```
dbanck, line 405: This is off by one and retries 6 times in total if we start with `retried = 0`

internal/schemas/gen/gen.go
```diff
@@ -51,273 +50,420 @@ func main() {
...
-		src := strings.ToLower(provider.Source())
-		if ok := ignore[src]; ok {
-			continue
+func retryProviderSchema(ctx context.Context, tf *tfexec.Terraform, fullName string, retried int) (*tfjson.ProviderSchemas, error) {
+	maxRetries := 5
+	backoffPeriod := 2 * time.Second
+
+	ps, err := tf.ProvidersSchema(ctx)
+	if err != nil {
+		if retried >= maxRetries {
```
dbanck, line 430: This is off by one and retries 6 times in total if we start with `retried = 0`

dbanck (approved):
Nice work! Just a minor thing in the retry logic.

Displaying some kind of summary after running `go generate ...` would be nice, but we can add something later. 
```
Tier     | Count |  Embedded | Errors |
---------|-------|-----------|--------|
official |    34 | 34 (100%) |      0 |
partner  |   230 | 200 (87%) |     30 |
---------|-------|-----------|--------|
```

Issue #857, opened 2022-04-07
Provide function signature help


radeksimko:
## Background

LSP provides [`textDocument/signatureHelp`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_signatureHelp) to provide the client with data about a function within the context of a function call.

Terraform configuration language has a number of [built-in functions](https://www.terraform.io/docs/configuration/functions.html), some of which come from HCL.

### Function Signature Help

After picking the function name from a completion popup, or typing the name and the "trigger character" `(` (opening brace).

![image3](https://user-images.githubusercontent.com/287584/162210577-cb37d8ad-ed9b-4268-969d-04ceae756986.gif)

## Implementation Notes

 - [x] https://github.com/hashicorp/terraform-ls/issues/37
 - [x] Introduce `SignatureAtPos(filename string, pos hcl.Pos) (*FuncSignature, error)` into `PathDecoder` in `hcl-lang/decoder`
 - [x] Update `hcl-lang` in LS

Issue #37, opened 2020-03-24
Support built-in functions


radeksimko:
## Background

Terraform configuration language has a number of [built-in functions](https://www.terraform.io/docs/configuration/functions.html), some of which come from HCL.

### Name Completion

When a user requests completion on the RHS (after "="), aside from other candidates (such as references), they're presented with function names compatible with the attribute type, i.e. functions which return `string` (or "any type") if the attribute is of type `string`.

<img width="863" alt="image5" src="https://user-images.githubusercontent.com/287584/162210695-0210281a-45ba-48af-b82f-253d93bc0ab6.png">

### Argument Completion

When the user confirms the function name from completion or types the opening brace, they are provided with any references of the argument type, i.e. if the argument being completed is of `string` type, then any references of `string` type are provided.

<img width="613" alt="image4" src="https://user-images.githubusercontent.com/287584/162210858-4e967146-abc5-466f-9af3-61463664d9af.png">

### Hover

When the user hovers over the function name, they are provided with the signature of the function, brief description and link to the documentation for that function.

<img width="527" alt="image2" src="https://user-images.githubusercontent.com/287584/162211023-e40e7733-30d4-4a33-8455-b92b73f1d941.png">

### Semantic Tokens

The name of a valid function, such as `format` will be highlighted semantically as function name, name of an invalid function (such as `frmt`) will not. That way users can tell the difference between typo-ed function name and correct one just from the colours.

## Implementation Notes

 - [x] `zclconf/go-cty`
    - [x] Introduce `Description` to the [`Spec`](https://pkg.go.dev/github.com/zclconf/go-cty@v1.10.0/cty/function#Spec) and [`Parameter`](https://pkg.go.dev/github.com/zclconf/go-cty@v1.10.0/cty/function#Parameter) types of `zclconf/go-cty/function`
    - [x] Backfill descriptions of functions & args in cty [`stdlib`](https://pkg.go.dev/github.com/zclconf/go-cty@v1.10.0/cty/function/stdlib) from [docs](https://www.terraform.io/language/functions)
 - [x] `hashicorp/terraform`
    - [x] Backfill descriptions of functions & args in [Terraform Core](https://github.com/hashicorp/terraform/blob/main/internal/lang/functions.go) from [docs](https://www.terraform.io/language/functions) to enable signatures to be exposed via CLI later
 - [x] `hashicorp/terraform-schema`
[...]

PR #1077, opened 2022-09-23
Add support for function signature help

dbanck:
This PR enables signature help for function signatures. 

To make this work, we first add the function signatures from terraform-schema (introduced in https://github.com/hashicorp/terraform-schema/pull/145) to the `PathContext`. This work is also required to enable completion, hover, etc. for functions later on.

We then add a new handler for the `textDocument/signatureHelp` LSP method.

## UX
Nested function calls with variadic parameters

Functions without any arguments

No signature help, if passing too many arguments

Multiline function calls

---

* Depends on https://github.com/hashicorp/terraform-schema/pull/145
* Depends on https://github.com/hashicorp/hcl-lang/pull/135
* Part of https://github.com/hashicorp/terraform-ls/issues/37
* Closes https://github.com/hashicorp/terraform-ls/issues/857


internal/langserver/handlers/service.go
```diff
@@ -272,6 +272,17 @@ func (svc *service) Assigner() (jrpc2.Assigner, error) {
...
 			return handle(ctx, req, svc.TextDocumentFormatting)
 		},
+		"textDocument/signatureHelp": func(ctx context.Context, req *jrpc2.Request) (interface{}, error) {
+			err := session.CheckInitializationIsConfirmed()
+			if err != nil {
+				return nil, err
+			}
+
+			ctx = ilsp.WithClientCapabilities(ctx, cc)
+			ctx = ilsp.ContextWithClientName(ctx, &clientName)
```
radeksimko, line 282: What do we need the client name for here? 🤔

internal/lsp/signature.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/hashicorp/terraform-ls/internal/mdplain"
+	lsp "github.com/hashicorp/terraform-ls/internal/protocol"
+)
+
+func ToSignatureHelp(signature *lang.FunctionSignature) *lsp.SignatureHelp {
+	if signature == nil {
+		return nil
+	}
+
+	parameters := make([]lsp.ParameterInformation, 0)
```
radeksimko, line 14:
```suggestion
	parameters := make([]lsp.ParameterInformation, 0, len(signature.Parameters))
```
Nitpick: We can allocate the exact capacity since we know it upfront.

internal/lsp/signature.go
```diff
@@ -0,0 +1,33 @@
...
+	for _, p := range signature.Parameters {
+		parameters = append(parameters, lsp.ParameterInformation{
+			Label:         p.Name,
+			Documentation: mdplain.Clean(p.Description.Value),
```
radeksimko, line 18:
```suggestion
			// TODO: Support markdown per https://github.com/hashicorp/terraform-ls/issues/1212
			Documentation: mdplain.Clean(p.Description.Value),
```

internal/lsp/signature.go
```diff
@@ -0,0 +1,33 @@
...
+		})
+	}
+
+	return &lsp.SignatureHelp{
+		Signatures: []lsp.SignatureInformation{
+			{
+				Label:           signature.Name,
+				Documentation:   mdplain.Clean(signature.Description.Value),
```
radeksimko, line 26:
```suggestion
				// TODO: Support markdown per https://github.com/hashicorp/terraform-ls/issues/1212
				Documentation:   mdplain.Clean(signature.Description.Value),
```

internal/lsp/signature.go
```diff
@@ -0,0 +1,33 @@
...
+	parameters := make([]lsp.ParameterInformation, 0)
+	for _, p := range signature.Parameters {
+		parameters = append(parameters, lsp.ParameterInformation{
+			Label:         p.Name,
+			Documentation: mdplain.Clean(p.Description.Value),
+		})
+	}
+
+	return &lsp.SignatureHelp{
+		Signatures: []lsp.SignatureInformation{
```
radeksimko, line 23: Thinking out loud - what is even the use case for multiple signatures in the spec? 🤔

internal/langserver/handlers/signature_help.go
```diff
@@ -0,0 +1,38 @@
...
+
+	ilsp "github.com/hashicorp/terraform-ls/internal/lsp"
+	lsp "github.com/hashicorp/terraform-ls/internal/protocol"
+)
+
+func (svc *service) SignatureHelp(ctx context.Context, params lsp.SignatureHelpParams) (*lsp.SignatureHelp, error) {
+	_, err := ilsp.ClientCapabilities(ctx)
+	if err != nil {
+		return nil, err
+	}
```
radeksimko, line 14: Just double checking: Do we not need to gate certain parts of the response based on the client capabilities?

 - `documentationFormat` - we cannot provide markdown even if the client asks for it - so I suppose there's nothing to do there until/unless https://github.com/hashicorp/terraform-ls/issues/1212 is addressed
 - `activeParameterSupport` - I guess we'd have to send literal `0` if client doesn't support it, because the struct field is `int` (i.e. not optional), so that's a problem similar to https://github.com/hashicorp/terraform-ls/issues/1212 is it? Although we don't track it anywhere explicitly yet.
 - `labelOffsetSupport` - this would be dependent on https://github.com/hashicorp/terraform-ls/issues/1211 right?

radeksimko (approved):
LGTM overall, works great 🥇 👍🏻 

I left only very minor questions and comments in-line but they're not really blocking.

internal/langserver/handlers/service.go
dbanck: Good catch, copy/waste from `hover`

internal/lsp/signature.go
dbanck: That feature might be required for [function overloads](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads).

radeksimko: I see, so not something we'd ever need in HCL then.

internal/langserver/handlers/signature_help.go
dbanck: `documentationFormat` and `labelOffsetSupport` shouldn't be a problem. But I keep thinking about `activeParameterSupport`. 

We currently set `activeParameter` on [SignatureInformation](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#signatureInformation) and this field was added in `3.16.0`. So I might just set this on [SignatureHelp](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#signatureHelp) to instead to be save?

radeksimko: > We currently set activeParameter on [SignatureInformation](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#signatureInformation) and this field was added in 3.16.0. So I might just set this on [SignatureHelp](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#signatureHelp) to instead to be save?

Aye, I didn't realise _that_ minor difference either.

I was more referring to the fact that `activeParameter` in both cases is optional and if we were strictly following the spec, we shouldn't be sending it to the client, unless it explicitly sent us `activeParameterSupport: true`. But again - this is in practice hard to achieve since the underlying type is now `uint32` and not `*uint32`, so we cannot omit that unless we reimplement the relevant struct(s).

PR #1081, opened 2022-10-05
schemas: Treat legacy providers as hashicorp or builtin

radeksimko:
This was discovered as part of recent failing benchmarks, although it isn't direct root cause of the fail:

https://github.com/hashicorp/terraform-ls/actions/runs/3187102652/jobs/5198346112


dbanck (approved):
Good catch!

PR #1082, opened 2022-10-06
Update `setup-go` to obtain version via `go-version-file`

dbanck:
This PR updates our workflows to make `setup-go` obtain the Go version directly from the `.go-version` file.


radeksimko (approved):
Thank you for the cleanup! 🧹 👍🏻

PR #1084, opened 2022-10-07
state: avoid infinite recursion in JobStore

### LINKED ISSUE #1065: runtime: goroutine stack exceeds 1000000000-byte limit
### Server Version

```
0.29.1
```

### Terraform Version

```
Terraform v1.2.9
on darwin_amd64
```

### Client Version

```
Visual Studio Code 1.71.0
```

```hcl

```

### Expected Behavior

terraform-ls processes files

### Actual Behavior

terraform-ls crashes after attempting to process the files

runtime: goroutine stack exceeds 1000000000-byte limit
runtime: sp=0xc127800340 stack=[0xc127800000, 0xc147800000]
fatal error: stack overflow

radeksimko:
Fixes #1065

I couldn't come up with a more light-weight test which would reproduce the bug. As mentioned in the comment, it (scheduling over 1 million jobs) takes about 3-4 minutes to finish on M1 Pro and consumes decent chunk of almost all cores + ~5-6GB of memory at peak.

The Windows GHA environment runs out of memory before the test can even finish. 🙊 

There are a few options to consider:

1. remove the test entirely (some risk that we re-introduce the bug)
2. gate the test on ENV variable, such as `LONG_TESTS=1`
3. gate the test on build tags, such as `-tags=longtests`

We could also consider not running the test on non-empty `CI` ENV variable (which most CI systems set, incl. GitHub Actions), but I'd argue that it's probably not as valuable to run that test locally for most of the time either.

### DIFF
--- a/.github/workflows/nightly-tests.yml
+++ b/.github/workflows/nightly-tests.yml
@@ -1,4 +1,4 @@
-name: ci
+name: Nightly Tests
 
 on:
   schedule:
--- a/.github/workflows/nightly-tests.yml
+++ b/.github/workflows/nightly-tests.yml
@@ -1,0 +1,34 @@
+name: ci
+
+on:
+  schedule:
+    - cron:  '0 3 * * *'
+  workflow_dispatch:
+
+env:
+  GOPROXY: https://proxy.golang.org/
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    strategy:
+      fail-fast: false
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v3
+      -
+        name: Unshallow
+        run: git fetch --prune --unshallow
+      -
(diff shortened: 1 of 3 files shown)


dbanck (commented):
We (especially you) did quite some work to reduce CI run time over the last couple of weeks. Therefore, I would instead not run this as part of our usual CI pipeline. 

What do you think about running this as a nightly job, similar to the benchmarks? That would still ensure that we catch this bug before a release.

radeksimko:
> What do you think about running this as a nightly job, similar to the benchmarks?

We could, but that would have to be limited to linux and/or macos as it's just impossible to run that in the Windows GHA environment.

radeksimko:
@dbanck I have gated the test on `longtest` build tag and added a nightly workflow, as suggested, PTAL

dbanck (approved):
LGTM! I would just update the name

PR #1099, opened 2022-10-25
deps: bump hcl-lang to `b20d6288c7f3`

radeksimko:
This is to bring https://github.com/hashicorp/hcl-lang/pull/144 in


jpogran (approved):

dbanck (changes requested):
Can we bump this to `b8c811` now with https://github.com/hashicorp/hcl-lang/pull/145 merged?

radeksimko:
@dbanck Agreed, but I will raise a separate PR to fix the bug with `DependentBody` merging in `hcl-lang`, so we can reflect that here too

dbanck (approved):

PR #1116, opened 2022-11-25
ci: Bump `windows-2019` to `windows-latest`

dbanck:
[The issue](https://github.com/actions/runner-images/issues/4856), why we pinned the windows version to 2019 should be resolve by now. Let's bump this to latest.


jpogran (approved):
🚀

PR #1126, opened 2022-12-08
docs: remove deprecated function & change recommended function

devoc09:
This PR replaces the nvim's deprecated functions with new ones.

You can check for deprecated functions with `:h  deprecated` command on nvim.
```
- vim.lsp.buf.formatting()              Use vim.lsp.buf.format() with
                                        {async = true} instead.
```
nvim docs [here](https://neovim.io/doc/user/lsp.html#vim.lsp.buf.format()).

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -160,15 +160,15 @@ let g:LanguageClient_serverCommands = {
 lua <<EOF
   require'lspconfig'.terraformls.setup{}
 EOF
-autocmd BufWritePre *.tfvars lua vim.lsp.buf.formatting_sync()
-autocmd BufWritePre *.tf lua vim.lsp.buf.formatting_sync()
+autocmd BufWritePre *.tfvars lua vim.lsp.buf.format()
+autocmd BufWritePre *.tf lua vim.lsp.buf.format()
 ```
  - If you are using `init.lua`:
 ```lua
 require'lspconfig'.terraformls.setup{}
 vim.api.nvim_create_autocmd({"BufWritePre"}, {
   pattern = {"*.tf", "*.tfvars"},
-  callback = vim.lsp.buf.formatting_sync,
+  callback = function() vim.lsp.buf.format() end,
 })
 ```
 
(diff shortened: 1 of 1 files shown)


radeksimko (approved):
Thanks for the catch @devoc09 

Just one minor question: I'm assuming this API change was introduced in a recent version of NeoVim, and the heading currently implies the instructions are relevant for 0.5.0+. Shall we update this to reflect the version which introduced the `format()` method?

devoc09:
Thanks for question!
I've added a more recent version of the heading while keeping the original heading.

docs/USAGE.md
```diff
@@ -168,7 +168,28 @@ autocmd BufWritePre *.tf lua vim.lsp.buf.formatting_sync()
 require'lspconfig'.terraformls.setup{}
 vim.api.nvim_create_autocmd({"BufWritePre"}, {
   pattern = {"*.tf", "*.tfvars"},
-  callback = vim.lsp.buf.formatting_sync,
+  callback = function() vim.lsp.buf.formatting_sync() end,
```
radeksimko, line 171:
```suggestion
  callback = vim.lsp.buf.formatting_sync(),
```

docs/USAGE.md
```diff
@@ -168,7 +168,28 @@ autocmd BufWritePre *.tf lua vim.lsp.buf.formatting_sync()
...
+EOF
+autocmd BufWritePre *.tfvars lua vim.lsp.buf.format()
+autocmd BufWritePre *.tf lua vim.lsp.buf.format()
+```
+ - If you are using `init.lua`:
+```lua
+require'lspconfig'.terraformls.setup{}
+vim.api.nvim_create_autocmd({"BufWritePre"}, {
+  pattern = {"*.tf", "*.tfvars"},
+  callback = function() vim.lsp.buf.format() end,
```
radeksimko, line 192:
```suggestion
  callback = vim.lsp.buf.format(),
```

radeksimko (approved):
Thanks for the updates.

I am no expert in Lua - in fact I don't know it at all 😄 , but I made one small change to the PR as I _believe_ the function declaration is not necessary in both cases.

If I'm wrong - please do let me know - happy to update it in a separate PR. I'll just merge this one for now.

PR #1127, opened 2022-12-13
github: Pin external GitHub Actions to hashes

radeksimko:
The intention here is to reduce the security risk posed by the supply chain - i.e. externally maintained GitHub Actions.

The expectation is that dependabot will continue to update these hashes as and when new versions become available.


dbanck (approved):
👍 

There was a new setup-go release (3.5.0) two hours ago. We can use this to see if the upgrades via dependabot work as expected.

PR #1140, opened 2022-12-30
Update doc: call NeoVim lsp format API with `async = true`

xiao-pp:
After following the setup guide for Neovim `v0.8.0+`, the auto formatting did
not work on `BufWritePre`.
Following the git history and the docs was updated in #1126. As it said in the
PR:
```
- vim.lsp.buf.formatting()              Use vim.lsp.buf.format() with
                                        {async = true} instead.
```
So I tried to add `async = true` and it works as expected.

The example config is incorrect anyway because the parameter `callback` takes
a function, not the invocation.

I know very little about NeoVim's API. Therefore I am not sure if this should
be the common setup. Also I only use/tested the pure lua config version.

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -181,15 +181,17 @@ vim.api.nvim_create_autocmd({"BufWritePre"}, {
 lua <<EOF
   require'lspconfig'.terraformls.setup{}
 EOF
-autocmd BufWritePre *.tfvars lua vim.lsp.buf.format()
-autocmd BufWritePre *.tf lua vim.lsp.buf.format()
+autocmd BufWritePre *.tfvars lua vim.lsp.buf.format({ async = true })
+autocmd BufWritePre *.tf lua vim.lsp.buf.format({ async = true })
 ```
  - If you are using `init.lua`:
 ```lua
 require'lspconfig'.terraformls.setup{}
 vim.api.nvim_create_autocmd({"BufWritePre"}, {
   pattern = {"*.tf", "*.tfvars"},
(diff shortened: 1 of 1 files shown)


radeksimko:
Hi @xiao-pp 
Thanks for the note.

Two points here:
1. Would you mind singing the CLA?
2. I am also not very familiar with NeoVim's API and admittedly I didn't test this. I mostly just trust @devoc09 - hence I'm pinging them here again - @devoc09 do you have any context to add about (a)sync mode when it comes to formatting? i.e. why should/shouldn't it be async?

radeksimko:
@devoc09 Fair enough - can you confirm that it _does_ work for you with `async = true` too?

devoc09:
I think that it can work with `{async = true}`.
If we add it args, we can ignore the timeout, so it seems to be affected.

ref: https://neovim.io/doc/user/lsp.html
> timeout_ms (integer|nil, default 1000): Time in milliseconds to block for formatting requests. No effect if async=true

radeksimko:
I'd be happy to merge this then if @xiao-pp can accept the CLA.

dimbleby:
neovim docs suggest that async is _not_ wanted for formatting on save, they give this explicitly as an example where you would not want to be asynchronous (and default to formatting not being asynchronous).

https://github.com/neovim/neovim/blob/443bbfd59e2818aeee72d2ed77af1cb24617e46f/runtime/doc/lsp.txt#L128-L136

which seems to make sense, you wouldn't want to leave a file before the formatting had completed.

Issue #1085, opened 2022-10-10
Failing `walker` tests when generated schemas are in place


radeksimko:
## Background

Due to recent changes made as part of https://github.com/hashicorp/terraform-ls/pull/1071, which includes the removal of `preloadschema` build tag, some `walker` tests now fail, if the embedded schemas contain the providers used by those tests:

```
--- FAIL: TestWalker_complexModules (10.28s)
    --- FAIL: TestWalker_complexModules/0-/Users/radeksimko/gopath/src/github.com/hashicorp/terraform-ls/internal/walker/testdata/single-root-ext-modules-only (3.86s)
        walker_test.go:453: expected only local sources, found: "preloaded"
    --- FAIL: TestWalker_complexModules/1-/Users/radeksimko/gopath/src/github.com/hashicorp/terraform-ls/internal/walker/testdata/single-root-local-and-ext-modules (4.18s)
        walker_test.go:453: expected only local sources, found: "preloaded"
    --- FAIL: TestWalker_complexModules/2-/Users/radeksimko/gopath/src/github.com/hashicorp/terraform-ls/internal/walker/testdata/single-root-local-modules-only (0.41s)
        walker_test.go:453: expected only local sources, found: "preloaded"
    --- FAIL: TestWalker_complexModules/3-/Users/radeksimko/gopath/src/github.com/hashicorp/terraform-ls/internal/walker/testdata/single-root-no-modules (0.00s)
        walker_test.go:453: expected only local sources, found: "preloaded"
    --- FAIL: TestWalker_complexModules/4-/Users/radeksimko/gopath/src/github.com/hashicorp/terraform-ls/internal/walker/testdata/nested-single-root-ext-modules-only (0.00s)
        walker_test.go:453: expected only local sources, found: "preloaded"
    --- FAIL: TestWalker_complexModules/5-/Users/radeksimko/gopath/src/github.com/hashicorp/terraform-ls/internal/walker/testdata/nested-single-root-local-modules-down (0.29s)
        walker_test.go:453: expected only local sources, found: "preloaded"
    --- FAIL: TestWalker_complexModules/6-/Users/radeksimko/gopath/src/github.com/hashicorp/terraform-ls/internal/walker/testdata/nested-single-root-local-modules-up (0.17s)
        walker_test.go:453: expected only local sources, found: "preloaded"
[...]

PR #1154, opened 2023-01-25
ci: Release binaries via CRT

dbanck:
This is a result of following [the CRT onboarding guide](https://hashicorp.atlassian.net/wiki/spaces/RELENG/pages/2300936220/CRT+Onboarding+Guide).

## TODO

 - [x] Have access to cut releases granted to maintainers
 - [x] Verify all staged artifacts were built correctly
 - [x] Remove `onboard-crt` as build branch from configs
 - [ ] https://github.com/hashicorp/linux-packaging/pull/165
 - [x] Add `release-engineering` team to repository


.github/workflows/build.yml
```diff
@@ -0,0 +1,130 @@
...
+          product_name: ${{ env.PKG_NAME }}
+          product_version: ${{ needs.set-product-version.outputs.product-version }}
+          go_version: ${{ needs.get-go-version.outputs.go-version }}
+          os: ${{ matrix.goos }}
+          arch: ${{ matrix.goarch }}
+          reproducible: report
+          instructions: |
+            go generate ./internal/schemas # TODO don't run for each entry
+            du -h -s ./internal/schemas/data
+            go build -ldflags "-s -w -X 'main.algoliaAppID=${{ secrets.ALGOLIA_APP_ID }}' -X 'main.algoliaAPIKey=${{ secrets.ALGOLIA_API_KEY }}'"
```
radeksimko, line 101:
```suggestion
            go build -trimpath -ldflags "-s -w -X 'main.algoliaAppID=${{ secrets.ALGOLIA_APP_ID }}' -X 'main.algoliaAPIKey=${{ secrets.ALGOLIA_API_KEY }}'" -o $BIN_PATH
```
It does not appear to be well documented, but it turns out the binary is expected to appear in a specific location (`$BIN_PATH`). I discovered this by trial & error here: https://github.com/hashicorp/hc-install/actions/runs/4014058525/jobs/6894127678#step:4:560

The `-trimpath` flag is what we had in GoReleaser config - it ensures reproducible builds.

version/VERSION
```diff
@@ -0,0 +1 @@
+0.31.0-dev
```
radeksimko, line 1: Is the expectation that we'll set the version back to `-dev` in the first commit after successful release?

version/version.go
```diff
@@ -0,0 +1,50 @@
...
+	VersionMetadata = ""
+
+	// The date/time of the build (actually the HEAD commit in git, to preserve stability)
+	// This isn't just informational, but is also used by the licensing system. Default is chosen to be flagantly wrong.
+	BuildDate string = "1970-01-01T00:00:01Z"
+)
+
+// GetHumanVersion composes the parts of the version in a way that's suitable
+// for displaying to humans.
+func GetHumanVersion() string {
```
radeksimko, line 35: Is there a reason we cannot continue calling this `VersionString()`? The `Get*` part in particular is typically considered anti-pattern in Go method names and AFAICT we also use this version as part of the `terraform-ls version -json` output.

version/version.go
```diff
@@ -0,0 +1,50 @@
+package version
```
radeksimko, line 1: Is there a reason this has to be a separate (now importable) package, as opposed to a file under `main` package as before?

version/version.go
```diff
@@ -0,0 +1,50 @@
...
+	// The next version number that will be released. This will be updated after every release
+	// Version must conform to the format expected by github.com/hashicorp/go-version
+	// for tests to work.
+	// A pre-release marker for the version can also be specified (e.g -dev). If this is omitted
+	// then it means that it is a final release. Otherwise, this is a pre-release
+	// such as "dev" (in development), "beta", "rc1", etc.
+	//go:embed VERSION
+	fullVersion string
+
+	Version, VersionPrerelease, _ = strings.Cut(fullVersion, "-")
```
radeksimko, line 23: Is there a reason to omit the `go-version` check?

version/version.go
```diff
@@ -0,0 +1,50 @@
...
+	if release != "" {
+		version += fmt.Sprintf("-%s", release)
+	}
+
+	if metadata != "" {
+		version += fmt.Sprintf("+%s", metadata)
+	}
+
+	// Strip off any single quotes added by the git information.
+	return strings.ReplaceAll(version, "'", "")
```
radeksimko, line 49: I don't entirely understand where/how we could end up with the quotes here? I thought it would either come from ldflags, or the plaintext `VERSION` file?

radeksimko (commented):
I understand this is still a draft - so not all comments may be relevant as you're still working on this and you may have already noticed the same thing, but I hope the early review is still helpful. 🤷🏻‍♂️ 😃

version/VERSION
dbanck: I've seen teams doing a [stage commit](https://github.com/hashicorp/consul/commit/bd257019c580a7f06b3ef99168698929b74f4bfc) to get everything ready for a release and then re-add the `-dev` prerelease string afterwards.

version/version.go
dbanck: I've reverted it to `VersionString`

version/version.go
dbanck: Moved back to `main` as discussed

version/version.go
dbanck: I re-added the version check. Do you know what calls the `init` function?

radeksimko: The `init` function is called automatically upon import of the package (see e.g. https://www.digitalocean.com/community/tutorials/understanding-init-in-go), so in our case at the earliest stage of any `terraform-ls` command. This ensures that if it panics, then it panics very loudly and early.

.release/ci.hcl
```diff
@@ -0,0 +1,79 @@
...
+  action "promote-production" {
+    organization = "hashicorp"
+    repository = "crt-workflows-common"
+    workflow = "promote-production"
+  }
+
+  notification {
+    on = "always"
+  }
+}
```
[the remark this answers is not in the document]
dbanck: Good catch! Added it.

.github/CODEOWNERS
```diff
@@ -1 +1,5 @@
 * @hashicorp/tf-editor-experience-engineers
+
+# release configuration
+/.release/                              @hashicorp/release-engineering
+/.github/workflows/build.yml            @hashicorp/release-engineering
```
radeksimko, line 5: We can double check this, but my understanding of the [relevant guide part](https://hashicorp.atlassian.net/wiki/spaces/RELENG/pages/2307588257/CODEOWNERS) is that this is only necessary where there's assumed ownership over these files by the RelEng team - typically a case of a bigger team which has more dedicated RelEng staff who actually manage the files, but I'm not entirely sure this applies to us here.

Relatedly, the last section of the guide page talks about ownership expectations, which in 95% of the common cases (where external contributors generally don't understand and change these files) implies we are the owners (i.e. responsible for review & merge) so I'm not sure there's benefit in putting the RelEng team down as owner here.

.github/workflows/build-pr.yml
```diff
@@ -0,0 +1,70 @@
...
+        uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # https://github.com/actions/checkout/releases/tag/v3.3.0
+      - name: Set up Go
+        uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568 # https://github.com/actions/setup-go/releases/tag/v3.5.0
+        with:
+          go-version-file: ".go-version"
+      - name: Download provider schemas
+        uses: actions/download-artifact@9bc31d5ccc31df68ecc42ccf4149144866c47d8a # https://github.com/actions/download-artifact/releases/tag/v3.0.2
+        with:
+          name: provider-schema-data
+          path: ./internal/schemas/data
```
radeksimko, line 64: Is there a strong benefit of waiting for building _with_ the schemas, as opposed to without - and running both jobs in parallel? I can see the reason for blocking tests on schema generation (once https://github.com/hashicorp/terraform-ls/issues/1085 is addressed) but I'm not sure the extra time here in context of a PR is justified.

I know that building could fail due to incorrect `embed` notation or something like that, but that's something tests will implicitly check for anyway, no?

.github/workflows/ci.yml
```diff
@@ -12,33 +12,6 @@ env:
   GOPROXY: https://proxy.golang.org/
```
radeksimko, line 13: Given the split now, shall we call this workflow (both file & `name` above) something like `ci-tests`?

.github/workflows/ci.yml
```diff
@@ -12,33 +12,6 @@ env:
...
-      -
-        name: Snapshot build (cross-platform)
-        uses: goreleaser/goreleaser-action@8f67e590f2d095516493f017008adc464e63adb1 # https://github.com/goreleaser/goreleaser-action/releases/tag/v4.1.0
-        with:
-          version: latest
-          args: build --snapshot --skip-post-hooks
-
   test:
     runs-on: ${{ matrix.os }}
     timeout-minutes: 10
```
radeksimko, line 17: I don't expect https://github.com/hashicorp/terraform-ls/issues/1085 to be addressed in this PR/context here, but just asking for our future selves: Can you confirm we'll still have some way of reading the artifacts (e.g. using the `actions/download-artifacts`) even when it's a separate workflow, rather than just a job?

.release/ci.hcl
```diff
@@ -0,0 +1,92 @@
+schema = "1"
+
+project "terraform-ls" {
+  team = "terraform-editor-experience"
```
radeksimko, line 4: This attribute is document as unused - is there a reason to include it at this point? If so - what does it refer to specifically?

version.go
```diff
@@ -1,19 +1,27 @@
...
-var version = "0.0.0"
+var (
+	// The next version number that will be released. This will be updated after every release
+	// Version must conform to the format expected by github.com/hashicorp/go-version
+	// for tests to work.
+	// A pre-release marker for the version can also be specified (e.g -dev). If this is omitted
+	// then it means that it is a final release. Otherwise, this is a pre-release
+	// such as "dev" (in development), "beta", "rc1", etc.
+	//go:embed version/VERSION
+	fullVersion string
```
radeksimko, line 18: Nitpick: Given that the version will be hand-crafted, is it worth calling this `rawVersion` or something like that, to indicate that the version was not checked for validity/correctness yet?

version.go
```diff
@@ -1,19 +1,27 @@
...
 
-// A pre-release marker for the version. If this is "" (empty string)
-// then it means that it is a final release. Otherwise, this is a pre-release
-// such as "dev" (in development), "beta", "rc1", etc.
-var versionPrerelease = "dev"
+	version, versionPrerelease, _ = strings.Cut(fullVersion, "-")
```
radeksimko, line 20: Nitpicky idea: Could we possibly call a custom function which leverages `go-version` and returns the different parts, instead of re-implementing small part of it via `strings.Cut` which may not cover all edge cases correctly? e.g. `1.2.3+meta-data`.

For example
```go
func parseRawVersion(rawVersion string) (version, prerelease, meta string) {
	v := goversion.Must(goversion.NewVersion(rawVersion))
	version = v.Core().String()
	prerelease = v.Prerelease()
	meta = v.Metadata()
}
```

We _might_ be able to then also omit the `init()` check, but I'm not sure how `panic`s are handled in this context - that would need double checking.

radeksimko (commented):
LGTM for the most part, I just left some minor in-line comments/questions. Thanks for all your work on this!

.github/workflows/ci.yml
dbanck: What artifacts are you referring? For example, if we want to run the tests with schemas present, we would also need to generate those in this workflow.

.github/workflows/ci.yml
dbanck: Yes! I would just go with `test` then

.github/workflows/ci.yml
radeksimko: In that case - and in relation to my other comment - would it make sense to only extract the actual building to `build-pr` workflow and keep the schema generation here? By artifacts I mean the provider schema JSON files.

version.go
dbanck: Manually calling `panic` inside `init` or using `goversion.Must` inside a `parseRawVersion` function produces similar results. Any `terraform-ls` command will panic:

```
$ terraform-ls serve -port 3333
panic: Malformed version: 1.0.0-beta+exp.sha.5114f85"

goroutine 1 [running]:
github.com/hashicorp/go-version.Must(...)
        /Users/dbanck/go/pkg/mod/github.com/hashicorp/go-version@v1.6.0/version.go:102
main.parseRawVersion({0x17733b6?, 0x1a0?})
        /Users/dbanck/Projects/terraform-ls/version.go:28 +0x8e
main.init()
        /Users/dbanck/Projects/terraform-ls/version.go:19 +0x39
```

Instead of storing version, prerelease and metadata in separate variables, we could store the `Version` instead. So we don't need to recreate the version string manually and can return `version.String()`. What do you think of this approach?

.release/ci.hcl
radeksimko: FWIW I omitted it for `hc-install` and the initial CI run was green, although I didn't get to staging/promotion due to other reasons yet.

.github/CODEOWNERS
dbanck: After consulting with RelEng, I also added our team as an owner for the release configuration.

radeksimko:
Is it worth removing the `dist` directory from gitignore now, that we won't build via GoReleaser there?

https://github.com/hashicorp/terraform-ls/blob/32d9b33ff956845af56bb29d898e36360c6c84ee/.gitignore#L1

dbanck:
I did the following checks to verify the (most) artifacts:
* `terraform-ls_0.31.0-dev_darwin_amd64.zip` — ✅  works and has provider schemas
* `terraform-ls_0.31.0-dev_windows_amd64.zip` — ✅ works and has provider schemas
* `terraform-ls_0.31.0-dev_linux_amd64.zip` — ✅ works and has provider schemas
* Staged artifacts match published on releases.hashicorp.com/terraform-ls ✅ 
* File size of rpm and debian packages matches the binary ✅

radeksimko (approved):
LGTM. 🚢 

Some random spot checks confirm the artifacts (as available in Artifactory currently) were built as expected:

```
gpg --verify terraform-ls_0.31.0-dev_SHA256SUMS.72D7468F.sig terraform-ls_0.31.0-dev_SHA256SUMS
gpg: Signature made Tue 31 Jan 16:07:24 2023 GMT
gpg:                using RSA key 374EC75B485913604A831CC7C820C6D5CD27AB87
gpg: Good signature from "HashiCorp Security (hashicorp.com/security) <security@hashicorp.com>" [ultimate]
```
```
shasum -a256 -c terraform-ls_0.31.0-dev_SHA256SUMS
terraform-ls_0.31.0-dev_darwin_amd64.zip: OK
terraform-ls_0.31.0-dev_darwin_arm64.zip: OK
terraform-ls_0.31.0-dev_freebsd_386.zip: OK
terraform-ls_0.31.0-dev_freebsd_amd64.zip: OK
terraform-ls_0.31.0-dev_freebsd_arm.zip: OK
terraform-ls_0.31.0-dev_linux_386.zip: OK
terraform-ls_0.31.0-dev_linux_amd64.zip: OK
terraform-ls_0.31.0-dev_linux_arm.zip: OK
terraform-ls_0.31.0-dev_linux_arm64.zip: OK
terraform-ls_0.31.0-dev_openbsd_386.zip: OK
terraform-ls_0.31.0-dev_openbsd_amd64.zip: OK
terraform-ls_0.31.0-dev_solaris_amd64.zip: OK
terraform-ls_0.31.0-dev_windows_386.zip: OK
terraform-ls_0.31.0-dev_windows_amd64.zip: OK
terraform-ls_0.31.0-dev_windows_arm64.zip: OK
```
```
codesign -dv --verbose=4 ./terraform-ls
Executable=/Users/radeksimko/Downloads/terraform-ls
Identifier=payload
Format=Mach-O thin (x86_64)
CodeDirectory v=20500 size=163539 flags=0x10000(runtime) hashes=5105+2 location=embedded
VersionPlatform=1
VersionMin=657664
VersionSDK=657664
Hash type=sha256 size=32
CandidateCDHash sha1=c91449e674478803a7dbb843ebb9157c77da769e
CandidateCDHashFull sha1=c91449e674478803a7dbb843ebb9157c77da769e
CandidateCDHash sha256=b8f70d10093725bb3c22e77f94c19f41e0714863
CandidateCDHashFull sha256=b8f70d10093725bb3c22e77f94c19f41e07148638c3b365e4439d3b3db9c3623
Hash choices=sha1,sha256
CMSDigest=12b41989d9fbb031684044c8847f832e48c3e5fad2bfc8beae62306d37ff2bc2
CMSDigestType=2
Executable Segment base=0
Executable Segment limit=19476480
Executable Segment flags=0x1
Page size=4096
CDHash=b8f70d10093725bb3c22e77f94c19f41e0714863
Signature size=9127
Authority=Developer ID Application: Hashicorp, Inc. (D38WU7D763)
Authority=Developer ID Certification Authority
Authority=Apple Root CA
Timestamp=31 Jan 2023 at 15:50:13
Info.plist=not bound
TeamIdentifier=D38WU7D763
Runtime Version=10.9.0
Sealed Resources=none
Internal requirements count=1 size=168
```

PR #1166, opened 2023-02-03
ci: Remove old release workflow

radeksimko:
This is now not needed after onboarding to CRT per https://github.com/hashicorp/terraform-ls/pull/1154


dbanck (approved):
Whoops, I missed that in #1154

PR #1167, opened 2023-02-03
ci: Automate releasing via bob

radeksimko:
## Context

Releasing currently requires a number of additional steps outside of GitHub. For a simple project with simple release pipeline such as this one, there is no strong benefit to do multi-staged releasing, as some other bigger projects do.

## Proposal

Simplify release process by moving promotion steps into a GHA workflow.

## TODO

 - [x] Create a dedicated GH bot account with appropriate access
 - [x] Grant access for `tf-editor-exp-ci-bot` to `bob` repo
 - [x] Generate new PAT within the GH bot account ^
 - [x] Set PAT as a new secret within this repo `BOB_GITHUB_TOKEN`
 - [x] Update release docs

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -208,4 +208,4 @@ Release process:
  1. Run the [Release workflow](https://github.com/hashicorp/terraform-ls/actions/workflows/release.yml) with the appropriate version (matching the one in `version/VERSION`) & SHA (long one).
  1. Wait for `staging` release [is finished](https://github.com/hashicorp/crt-workflows-common/actions/workflows/crt-promote-staging.yml).
  1. Wait for a message in the Slack channel saying that authorisation is needed to promote artifacts to production. Click on the link and approve.
- 1. Once notified that promotion is successful, go to https://github.com/hashicorp/crt-workflows-common/actions/workflows/promote-production-packaging.yml, locate the terraform-ls promote-production-packaging workflow, and approve.
+ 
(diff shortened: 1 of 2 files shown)


dbanck:
Can we run the staging and production back to back? Or do we have to wait until staging completes before running production?

radeksimko:
> Can we run the staging and production back to back? Or do we have to wait until staging completes before running production?

I have added the dependency on line 49 https://github.com/hashicorp/terraform-ls/pull/1167/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R49 but it's not strictly necessary. The real "awkward part" though about the whole process is that both `bob` commands are only triggers and do not wait for completion, so the dependency won't make that much of a difference in reality.

I discussed this problem more with the RelEng team and they are planning to remove the requirement of `staging` altogether some time in 2024, which should eventually resolve this awkwardness.

Strictly speaking - and to answer your question - it should be fine because the `production` workflow does not require `staging` to finish until it is actually approved.

dbanck:
Thanks for the additional context. Should we add a note to the README that we need to wait until the staging workflow has succeeded and the artifacts have landed in the staging bucket (also a good time to manually test the build) before approving the production workflow?

radeksimko:
@dbanck added a step for this in the instructions, PTAL

.github/CONTRIBUTING.md
```diff
@@ -205,31 +205,7 @@ Release process:
...
-  --slack-channel=C02AGQXCAF5 \
-  --product-version="$VERSION" \
-  --sha="$SHA" \
-  --branch=main \
-  production
- ```
+ 1. Run the [Release workflow](https://github.com/hashicorp/terraform-ls/actions/workflows/release.yml) with the appropriate version (matching the one in `version/VERSION`) & SHA (long one).
+ 1. Wait for `staging` release [is finished](https://github.com/hashicorp/crt-workflows-common/actions/workflows/crt-promote-staging.yml).
+ 1. Wait for a message in the Slack channel saying that authorisation is needed to promote artifacts to production. Click on the link and approve.
+ 1. Once notified that promotion is successful, go to https://github.com/hashicorp/crt-workflows-common/actions/workflows/promote-production-packaging.yml, locate the terraform-ls promote-production-packaging workflow, and approve.
```
dbanck, line 211: I think this isn't required anymore. You only have to approve the production workflow once.

dbanck (approved):
Nice!

Issue #1067, opened 2022-09-08
Mishandling of hidden files & emacs backup files


jleechpe:
### Server Version
0.29.0
0.29.1
0.29.2
```

```

### Terraform Version
```
1.2.8
```

### Client Version
```
ArchLinux
Emacs 28.1
- LSP-Mode (latest git)
- Eglot (Latest Git)
```

### Terraform Configuration Files
```hcl
<New File>
```

`[|]` is cursor
```hcl
resource "azu[|]" "" {

}
```
### Log Output
Shows successful completion right after initializing then error on subsequent (after file is modified):
https://gist.github.com/jleechpe/3c449eab5885fc9149c5fa9a28d95261

### Expected Behavior
Completions to populate

### Actual Behavior
Emacs LSP-mode responds with "cannot find file: main.tf"
Emacs Eglot responds with json-rpc error `"main.tf: file not found"`

### Steps to Reproduce
LSP-Mode
1. Open a terraform file in Emacs
2. Connect with lsp-mode (default settings)
3. Try to complete -> First attempt succeeds
4. Edit the file in any way
5. Try to complete -> LSP errors with cannot find file

Eglot
1. Open a terraform file in Emacs
2. Connect with eglot (`terraform-ls serve -log-file ~/tmp/tflog/log`)
3. Trigger Completion -> Succeeds
4. Edit the file
5. Trigger Completion -> JSON RPC error `"main.tf: file not found"`
6. Undo edits to file (reverts back to initial state)
7. Trigger Completion -> Succeeds as it did in 3.

Manually downloaded `terraform-ls` versions 0.29.2, 0.29.1, 0.29.0, 0.28.1 and tested with eglot:
Connect using: `~/downloads/terraform-ls/<version>/terraform-ls serve` -> 0.29.x all error.  0.28.1 behaves as expected.

Tested on Windows with Emacs 29.0.50, Terraform 1.2.9 and terraform-ls 0.29.2 -> No issues
Tested on Windows with VSCode -> No issues
Tested on Archlinux with VSCode -> Same issue as in Emacs (worked on initial test but multiple tests since have the same issue as Emacs)

I was able once or twice to get terraform-ls 0.29.2 to work in Emacs after stopping/restarting the server multiple (10+) times, but this was inconsistent and I could not determine what made it start working.

Also see: https://github.com/emacs-lsp/lsp-mode/issues/3713 (issue on MacOS as well)

Edit: VSCode does not work on subsequent tests

[...]

PR #1172, opened 2023-02-10
Continue parsing module files on error

dbanck:
This PR changes the parsing of module files so that if we encounter an inaccessible file, we continue parsing all the other module files instead of aborting with an error.

That should improve the experience for emacs users, relying on magic symlinks to keep track of open files.

Fixes https://github.com/hashicorp/terraform-ls/issues/1067

## UX

**Before**
CleanShot 2023-02-13 at 11 55 20@2x
No features available within the module

**After**
CleanShot 2023-02-13 at 11 56 03@2x

### DIFF
--- a/internal/terraform/parser/module.go
+++ b/internal/terraform/parser/module.go
@@ -32,7 +32,9 @@ func ParseModuleFiles(fs FS, modPath string) (ast.ModFiles, ast.ModDiags, error)
 
 		src, err := fs.ReadFile(fullPath)
 		if err != nil {
-			return nil, nil, err
+			// If a file isn't accessible, continue with reading the
+			// remaining module files
+			continue
 		}
 
 		filename := ast.ModFilename(name)
--- a/internal/terraform/parser/module_test.go
+++ b/internal/terraform/parser/module_test.go
@@ -83,6 +83,15 @@ func TestParseModuleFiles(t *testing.T) {
 				},
 			},
 		},
+		{
+			"invalid-links",
+			map[string]struct{}{
+				"resources.tf": {},
+			},
+			map[string]hcl.Diagnostics{
+				"resources.tf": nil,
+			},
+		},
 	}
 
 	fs := osFs{}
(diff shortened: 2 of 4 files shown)


jpogran (approved):

internal/terraform/parser/module.go
```diff
@@ -32,7 +32,9 @@ func ParseModuleFiles(fs FS, modPath string) (ast.ModFiles, ast.ModDiags, error)
 
 		src, err := fs.ReadFile(fullPath)
 		if err != nil {
-			return nil, nil, err
+			// If a file isn't accessible, continue with reading the
+			// remaining module files
+			continue
```
radeksimko, line 37: This makes sense for the edge case described in https://github.com/hashicorp/terraform-ls/issues/1067 but have we considered what impact this would have on real `*.tf` files which aren't accessible, e.g. due to mis-set permissions?

I'm guessing that we are just hoping that both client and server would be aligned on the "readability" of the same file, i.e. they'd be launched under the same user and have the same level of access. i.e. if we can't index the file, it also won't come through `textDocument/didOpen`, right?

So probably nothing to do here - just thinking out loud about the consequences and making sure we're aligned.

One edge case which comes to mind is - what if the user changes permissions at runtime - I'm guessing that this would come through the watch event, right? And presumably it comes before `didOpen` event, so there should be no race condition.

radeksimko (approved):

radeksimko (approved):

internal/terraform/parser/module.go
dbanck: That's a valid concern, thanks for raising it. I did some testing with mis-set permissions, but I can't create a test case. It seems impossible commit such a file, because git can't access it 😅 

CleanShot 2023-02-13 at 12 02 10@2x

The LS will skip it, and VS Code can't open the file, so it shouldn't come through `textDocument/didOpen`.
CleanShot 2023-02-13 at 11 56 10@2x

If the file permissions change, the watcher will pick it up and the LS will rerun all module related jobs:
CleanShot 2023-02-13 at 12 06 12@2x

PR #1178, opened 2023-02-14
Add copywrite config

jpogran:


radeksimko (changes requested):
Given that the `.github` folder somehow appears to be ignored when `header_ignore` is not defined in the config, I'm getting the sense there's some inconsistency and possibly bug, which seems worth reporting and fixing first, before we merge this.

.copywrite.hcl
```diff
@@ -0,0 +1,12 @@
+schema_version = 1
+
+project {
+  license        = "MPL-2.0"
```
radeksimko, line 4:
```suggestion
  license = "MPL-2.0"
```

radeksimko (approved):
I just got confused about the output of `copywrite`, `.github` is not ignored unless explicitly listed under `header_ignore`.

:shipit:

PR #1184, opened 2023-02-16
Update changelog / docs about release assets

dbanck:
This PR updates the changelog with an extra note about the location of the release assets.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## 0.30.2 (15 February 2023)
 
+NOTES / BREAKING CHANGES:
+
+ - We have changed our release process: all assets are available (and always were and will be) from the [HashiCorp Releases site](https://releases.hashicorp.com/terraform-ls)
+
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## 0.30.2 (15 February 2023)
 
+NOTES / BREAKING CHANGES:
+
+ - We have changed our release process: all assets are available (and always were and will be) from the [HashiCorp Releases site](https://releases.hashicorp.com/terraform-ls)
```
radeksimko, line 5:
```suggestion
 - We have changed our release process: all assets continue to be available from the [HashiCorp Releases site](https://releases.hashicorp.com/terraform-ls) and/or via the [Releases API](https://releases.hashicorp.com/docs/api/v1/), not as GitHub Release assets anymore.
```

Not that I anticipate the Releases site to go away, but we don't need to use as strong language in the forward-looking statement here.

radeksimko (approved):

PR #1193, opened 2023-02-22
Disable cgo for builds

dbanck:
When migrating to CRT, I missed disabling cgo for native (Linux) builds. We had this set before in the [`.goreleaser.yml`](https://github.com/hashicorp/terraform-ls/blob/f3eb1948a92f9b308fc33f8a665f13cece024979/.goreleaser.yml#L1-L2)

The current binaries released for Linux are dynamically linked, as reported in https://github.com/hashicorp/vscode-terraform/issues/1346


radeksimko (approved):
LGTM 🚢

PR #1208, opened 2023-03-09
Add TFC usage detection

jpogran:
This builds on the existing methods to detect remote backend usage and adds detection of cloud blocks in the terraform block.


internal/langserver/handlers/hooks_module.go
```diff
@@ -44,6 +44,25 @@ func moduleTelemetryData(mod *state.Module, ch state.ModuleChanges, store *state
...
+		// Required for Terraform Enterprise;
+		// Defaults to app.terraform.io for Terraform Cloud
+		if hostname == "" {
+			hostname = "app.terraform.io"
+		}
+
+		// anonymize any non-default hostnames
+		if hostname != "app.terraform.io" {
+			hostname = "custom-hostname"
+		}
```
radeksimko, line 62: This looks slightly inconsistent with how we check & report hostnames below for backends.

I don't have too strong opinion on whether we should be reporting the difference between default and empty (implied default) hostname, but I do think we should approach both the same way.

radeksimko (commented):
Aside from that in-line question about how we handle empty/default hostnames, this LGTM and works as expected:

```
2023/03/13 09:36:22 opts.go:205: Posting server notification "telemetry/event" {"v":1,"name":"moduleData","properties":{"cloud":true,"cloud.hostname":"custom-hostname","moduleId":"d1a79a40-a1df-414d-03ef-d1ff802e721d","tfVersion":"1.4.0"}}
```

internal/langserver/handlers/hooks_module.go
jpogran: According to https://developer.hashicorp.com/terraform/cli/cloud/settings#hostname, hostname is optional, and is an empty string if not specified. Using the same logic as backend would result is us sending an empty string in telemetry, which defeats the purpose in doing this.

radeksimko: I would assume that telemetry metrics with empty string still get sent through? As in - we'd still be able to see any `cloud` blocks parsed, the only difficulty would be treating empty strings as `app.terraform.io` when processing the data.

To put it differently - I'm happy with your implementation - which implies that we do no further interpretation of empty strings in AppInsights, but in that case I think we should do the same for the remote `backend` below, otherwise we end up comparing apples to pears.

jpogran: Yes, sending empty data is allowed.

If the default is known, I don't see the need for sending empty strings and then having to do conversion when reporting.

I'll open a ticket for backend reporting.

PR #1218, opened 2023-03-20
Fix remote backend tracking

jpogran:
Previously we would submit an empty backend if the configuration used the default value. This detects an empty value and uses the default "app.terraform.io" value instead.


radeksimko (approved):
LGTM, thanks for the fix!

PR #1256, opened 2023-04-24
deps: Pull in gopls v0.10.0 tsprotocol.go

radeksimko:
The main motivation here is to address a long-time bug which was originally "planted" in August 2021 as part of https://github.com/hashicorp/terraform-ls/pull/608 and made visible via https://github.com/hashicorp/terraform-ls/pull/1017 about 1 year later.

TL;DR older structs representing the LSP spec don't allow `LabelDetails` to be empty because the underlying type is a struct, rather than pointer which could be `nil`-ed and this codepath became active when we introduced completion resolution.

https://github.com/hashicorp/terraform-ls/blob/9e399569dc1a62ded8e424293bb5c38840c33343/internal/protocol/protocol.go#L871-L876

It also takes into account upstream changes made in https://github.com/golang/tools/commit/b15dac2b8849754d58ebde3b57bbb75776ba869b


dbanck (approved):
👍 Great work

PR #1258, opened 2023-04-26
Cache registry module errors

dbanck:
This PR introduces caching for registry module HTTP request errors. We're currently only caching 404 errors, but it might also be sensible to cache other errors.

The change dramatically improves the performance when many jobs run in the background after a more significant text change.

## UX

**Before**

**After**

### DIFF
--- a/internal/state/registry_modules.go
+++ b/internal/state/registry_modules.go
@@ -14,6 +14,7 @@ import (
 type RegistryModuleData struct {
 	Source  tfaddr.Module
 	Version *version.Version
+	Error   bool
 	Inputs  []registry.Input
 	Outputs []registry.Output
 }
@@ -28,6 +29,13 @@ func (s *RegistryModuleStore) Exists(sourceAddr tfaddr.Module, constraint versio
 
 	for obj := iter.Next(); obj != nil; obj = iter.Next() {
 		p := obj.(*RegistryModuleData)
+		// There was an error fetching the module, so we can't compare
+		// any versions
+		if p.Error {
+			return true, nil
+		}
+
+		// Check if there an entry with a matching version
 		if constraint.Check(p.Version) {
 			return true, nil
 		}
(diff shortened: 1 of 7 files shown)


internal/state/registry_modules.go
```diff
@@ -67,3 +75,31 @@ func (s *RegistryModuleStore) Cache(sourceAddr tfaddr.Module, modVer *version.Ve
 	txn.Commit()
 	return nil
 }
+
+func (s *RegistryModuleStore) CacheError(sourceAddr tfaddr.Module) error {
```
radeksimko, line 79: Similar question to above - should we call this e.g. `CacheNotFound()`?

radeksimko (approved):
LGTM, aside from the minor comments about naming. 👍🏻

internal/terraform/module/module_ops.go
```diff
@@ -638,6 +638,16 @@ func GetModuleDataFromRegistry(ctx context.Context, regClient registry.Client, m
 		metaData, err := regClient.GetModuleData(ctx, sourceAddr, declaredModule.Version)
 		if err != nil {
 			errs = multierror.Append(errs, err)
+
+			clientError := registry.RegistryClientError{}
+			if errors.As(err, &clientError) && clientError.Status == 404 {
```
jpogran, line 643: We may want more than a 404 here: https://developer.hashicorp.com/terraform/registry/api-docs#http-status-codes

>Client Errors: Invalid requests will receive the relevant 4xx status. Except where noted below, the request should not be retried.
> 
>Rate Limiting: Clients placing excessive load on the service might be rate-limited and receive a 429 code. This should be interpreted as a sign to slow down, and wait some time before retrying the request.
> 
>Service Errors: The usual 5xx errors will be returned for service failures. In all cases it is safe to retry the request after receiving a 5xx response.
> 
> Load Shedding: A 503 response indicates that the service is under load and can't process your request immediately. As with other 5xx errors you may retry after some delay, although clients should consider being more lenient with retry schedule in this case.

So, probably any 4XX error should be cached.

I could see the usefulness of 429 and 503, but probably out of scope of this attempt to implement.

radeksimko: I think we do _not_ want to be caching transient errors? At least not without some invalidation? AFAICT the primary (or common enough) root cause we identified behind the slow down is lack of access to private registry, which seems to be always surfaced as 404 as to not leak whether a module exists or not publicly.

429 and 5xx specifically I'd expect to be transient and something we should be actually retrying. Not saying we should not be caching those, but that seems to be a bit different scenario that may require different approach IMHO.

dbanck: I agree that we should only cache `429` if we have some kind of retry mechanism in place.

We could still cache all other client errors, like `clientError.StatusCode >= 400 && clientError.StatusCode < 429`?

radeksimko: I think 408 (Request Timeout) could also be transient? Otherwise 👌🏻

dbanck: 👍 PTAL

internal/state/registry_modules.go
```diff
@@ -14,6 +14,7 @@ import (
 type RegistryModuleData struct {
 	Source  tfaddr.Module
 	Version *version.Version
+	Error   bool
```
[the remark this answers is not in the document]
dbanck: Since we now cache most 4XX errors, I'll keep it as is?

radeksimko (approved):

PR #1268, opened 2023-04-28
docs: Add matrix with LSP features

radeksimko:
This aims to provide a quick overview of (un)supported methods, which in turn could be used as a guide for future planning and may be useful for client maintainers too.

This is similar to lists in other clients or servers, e.g. 

 - https://github.com/sublimelsp/LSP/blob/main/docs/src/features.md
 - https://github.com/golang/tools/blob/master/gopls/doc/design/design.md#features
 - https://github.com/juliosueiras/terraform-lsp#lsp-support-table

We may end up restructuring this and/or making it otherwise a bit more useful for common users, e.g. by adding GIFs and screenshots etc. and by linking to various relevant open issues and describing what exact parts are, will be and won't be implemented. That said I didn't intend this PR to be perfect, but rather a starting point on a path towards overall better documentation.

One thing we could do to aid readability though is to sort this by ✅/❌ and _then_ alphabetically, if you all think this may help.

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -1,0 +1,117 @@
+# Features of the server
+
+## LSP methods
+
+The LSP is a relatively extensive protocol with many features/methods, not all of which are implemented and not all of which are relevant for Terraform. The following matrix should provide some visibility into the current and future state.
+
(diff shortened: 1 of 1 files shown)


jpogran (approved):
😍

PR #1295, opened 2023-05-30
Change nvim_create_autocmd callback to a lambda

chiting:
Related to #1140 

`callback` in `vim.api.nvim_create_autocmd()` seems to require a lambda.

I didn't test it for VimScript or Neovim v0.5.0+ so I didn't change it for these.

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -175,7 +175,9 @@ autocmd BufWritePre *.tf lua vim.lsp.buf.format()
 require'lspconfig'.terraformls.setup{}
 vim.api.nvim_create_autocmd({"BufWritePre"}, {
   pattern = {"*.tf", "*.tfvars"},
-  callback = vim.lsp.buf.format(),
+  callback = function()
+    vim.lsp.buf.format()
+  end,
 })
 ```
 


radeksimko (commented):
> callback in vim.api.nvim_create_autocmd() seems to require a lambda.

Can you be a bit more specific - are you suggesting that auto-formatting is not triggered without the lambda, or that you get an error?

I'm not _that_ familiar with Lua or Vim, but the two snippets of code seem equivalent to me so I'm somewhat surprised if this makes a difference.

chiting:
> > callback in vim.api.nvim_create_autocmd() seems to require a lambda.
> 
> 
> 
> Can you be a bit more specific - are you suggesting that auto-formatting is not triggered without the lambda, or that you get an error?
> 
> 
> 
> I'm not _that_ familiar with Lua or Vim, but the two snippets of code seem equivalent to me so I'm somewhat surprised if this makes a difference.

I was getting an error and went to check the lua documentation which uses a lua function as parameter for the callback (or a string if it's a VimScript function name): https://neovim.io/doc/user/api.html#nvim_create_autocmd()

I'm not that familiar with lua either but it seems to me we're assigning the result returned by the function to the callback in the original snippet, instead the function itself.

radeksimko (approved):
> it seems to me we're assigning the result returned by the function to the callback in the original snippet, instead of the function itself

That looks like a plausible explanation which I'm going to trust 😅 Thanks for this extra context.

PR #1296, opened 2023-05-31
Add copyright headers automagically instead of failing a check on PRs

xiehan:
This is a solution we've been trialing on the CDK team and it's working well for us so I figured I'd propose it to this team as well.

Rather than failing a check on PRs if the copyright headers are missing (which is potentially annoying to community contributors who then either have to manually copy-paste the headers or install HashiCorp's copywrite tool), this adds a new commit on any PRs that are missing the headers that adds them in automagically. (If no headers are missing then it does nothing.)

### DIFF
--- a/.github/workflows/add-copyright-headers.yml
+++ b/.github/workflows/add-copyright-headers.yml
@@ -18,7 +18,7 @@ jobs:
       contents: write
     steps:
       - name: Checkout
-        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.3.0
+        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # https://github.com/actions/checkout/releases/tag/v3.5.2
         with:
           ref: ${{ github.event.pull_request.head.ref }}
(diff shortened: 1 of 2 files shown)


.github/workflows/add-copyright-headers.yml
```diff
@@ -0,0 +1,41 @@
...
+jobs:
+  add-copyright-headers:
+    runs-on: ubuntu-latest
+    env:
+      HEAD_REF: ${{ github.event.pull_request.head.ref }}
+    permissions:
+      contents: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.3.0
```
dbanck, line 21:
```suggestion
        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # https://github.com/actions/checkout/releases/tag/v3.5.2
```

I thought we need to bump the version here, but the hash is already correct. Only the comment has an older version.

dbanck (approved):
LGTM 👍

Issue #1271, opened 2023-05-02
Use `window/logMessage` instead of STDERR for log messages


radeksimko:
## Background

Currently the server uses STDERR as the default channel for logs. This can be overridden via `-log-file` flag of the `terraform-ls serve` subcommand, to send logs to a file.

The default behaviour is however undesirable for a couple of reasons:

 - it makes trawling through the logs difficult as it's all just plaintext, so we have to resort to `Ctrl/Cmd+F`, `grep` and similar techniques
 - some clients, such as NeoVim (understandably) treat STDERR as "errors", which then causes users to think there's "lots of errors" when these are just log messages
   - https://github.com/hashicorp/terraform-ls/issues/1234
   - https://github.com/hashicorp/terraform-ls/issues/1124
   - https://discuss.hashicorp.com/t/terraform-lsp-log-going-crazy/47715

Using STDERR originally was mostly a pragmatic "shortcut" we took in the early versions of LS development, which felt fine given that `window/logMessage` is not considered mandatory for servers to implement per the spec.

## Concerns

There were also a few concerns we didn't know how to best address at that point:

 - where do we send logs related to anything that happens prior to server setup, i.e. before initialization is finished and `window/logMessage` can be sent?
   - This is still an ongoing concern, but such situations and need for logging seem to be rare enough that we could only log errors and still send those to STDERR or something like that.
 - how do we decide on the log level inside the LS codebase?
   - This could be addressed either by adopting a library such as [go-hclog](https://github.com/hashicorp/go-hclog) or possibly the upcoming experimental [`slog` package](https://pkg.go.dev/golang.org/x/exp/slog) with some kind of custom adapters for LSP
 - how do we debug/log the LSP traffic itself without doubling the number of requests, i.e. without logging every e.g. `textDocument/completion` via `window/logMessage`?
   - This is still an ongoing concern, mostly dependent on various clients. The good news is that there are some relatively mature clients out there now which can do such logging on the client side.

## Proposal

Explore solutions for the above ^ concerns and look into adopting `window/logMessage`

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_logMessage

PR #1321, opened 2023-06-30
LSP Logging

jpogran:
This uses the LSP window/logMessage notification to send terraform-ls logs to the client instead of to stderr. This has the benefit of not using stderr for clients that report that as errors as well as shipping logs for TCP connections.

How these logs are displayed varies by editor, with VS Code, SublimeText and neovim appearing to handle these fine.

This is currently a hackweek project that needs some work to get into a state that can be shipped

### DIFF
--- a/internal/cmd/serve_command.go
+++ b/internal/cmd/serve_command.go
@@ -109,8 +109,11 @@ func (c *ServeCommand) Run(args []string) int {
 	}
 
 	srv := langserver.NewLangServer(ctx, handlers.NewSession)
-	srv.SetLogger(logger)
-
+	// TODO: Not sure about removing this here. We can't use window/LogMessage yet
+	// because we haven't started
+	// If we leave this here, some stuff still goes to stderr
+	// srv.SetLogger(logger)
+	
 	if c.port != 0 {
 		err := srv.StartTCP(fmt.Sprintf("localhost:%d", c.port))
 		if err != nil {
--- a/internal/langserver/handlers/initialize.go
+++ b/internal/langserver/handlers/initialize.go
@@ -12,6 +12,7 @@ import (
 	"github.com/creachadair/jrpc2"
 	lsctx "github.com/hashicorp/terraform-ls/internal/context"
 	"github.com/hashicorp/terraform-ls/internal/document"
+	"github.com/hashicorp/terraform-ls/internal/logging"
 	ilsp "github.com/hashicorp/terraform-ls/internal/lsp"
 	lsp "github.com/hashicorp/terraform-ls/internal/protocol"
 	"github.com/hashicorp/terraform-ls/internal/settings"
@@ -39,6 +40,9 @@ func (svc *service) Initialize(ctx context.Context, params lsp.InitializeParams)
 	expClientCaps := lsp.ExperimentalClientCapabilities(clientCaps.Experimental)
 
 	svc.server = jrpc2.ServerFromContext(ctx)
+	svc.logger = logging.NewLspLogger(&logging.LspLogger{
+		Context: ctx,
+	})
 
 	setupTelemetry(expClientCaps, svc, ctx, properties)
 	defer svc.telemetry.SendEvent(ctx, "initialize", properties)
--- a/internal/langserver/langserver.go
+++ b/internal/langserver/langserver.go
@@ -17,6 +17,7 @@ import (
 	"github.com/creachadair/jrpc2/channel"
 	"github.com/creachadair/jrpc2/server"
 	"github.com/hashicorp/terraform-ls/internal/langserver/session"
+	"github.com/hashicorp/terraform-ls/internal/logging"
 )
 
 type langServer struct {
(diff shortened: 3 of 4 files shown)


internal/cmd/serve_command.go
```diff
@@ -109,8 +109,11 @@ func (c *ServeCommand) Run(args []string) int {
 	}
 
 	srv := langserver.NewLangServer(ctx, handlers.NewSession)
-	srv.SetLogger(logger)
-
+	// TODO: Not sure about removing this here. We can't use window/LogMessage yet
+	// because we haven't started
+	// If we leave this here, some stuff still goes to stderr
+	// srv.SetLogger(logger)
```
radeksimko, line 115: Yes, this is one of the concerns I mentioned in https://github.com/hashicorp/terraform-ls/issues/1271

IMHO I think we'll just need to review what those log messages are, maybe remove any non-critical ones and continue sending the rest to STDERR as there just isn't any better location.

internal/langserver/langserver.go
```diff
@@ -105,6 +106,10 @@ func (ls *langServer) StartAndWait(reader io.Reader, writer io.WriteCloser) erro
 		cancelFunc()
 	}()
 
+	ls.SetLogger(logging.NewLspLogger(&logging.LspLogger{
+		Context: ctx,
```
radeksimko, line 110: In general storing context is considered anti-pattern in Go. We could pass the server instance here instead of the context. It would also make the contract much clearer, where the `LspLogger` depends on "LSP server" rather than arbitrary context which may or may not contain the instance of the server.

internal/langserver/langserver.go
```diff
@@ -85,7 +86,7 @@ func (ls *langServer) startServer(reader io.Reader, writer io.WriteCloser) (*sin
 		return nil, err
 	}
 	srv.Start(channel.LSP(reader, writer))
-
+
```
radeksimko, line 89:
```suggestion

```

internal/langserver/handlers/initialize.go
```diff
@@ -39,6 +40,9 @@ func (svc *service) Initialize(ctx context.Context, params lsp.InitializeParams)
 	expClientCaps := lsp.ExperimentalClientCapabilities(clientCaps.Experimental)
 
 	svc.server = jrpc2.ServerFromContext(ctx)
+	svc.logger = logging.NewLspLogger(&logging.LspLogger{
+		Context: ctx,
```
radeksimko, line 44:
```suggestion
		Server: svc.server,
```
As per other comment this would make for a cleaner contract.

internal/logging/lsp.go
```diff
@@ -0,0 +1,36 @@
...
+
+type LspLogger struct {
+	Context context.Context
+}
+
+func (l LspLogger) Write(p []byte) (int, error) {
+	logMessage := string(p)
+	// there appears to be an extra newline that's helpful for stderr
+	// but not for outputchannel
+	logMessage = strings.TrimSuffix(logMessage, "\n")
```
radeksimko, line 27: I wonder if this just originates from our use of some `Println`s in various places? 🤔 

I'd be slightly worried about the performance impact here in such a critical path but maybe it's negligible.

radeksimko (commented):
First of all thank you for looking into this!

Have you considered the concerns mentioned in https://github.com/hashicorp/terraform-ls/issues/1271 ?

Aside from in-line comments I'd be specifically still concerned about this

> how do we debug/log the LSP traffic itself without doubling the number of requests, i.e. without logging every e.g. `textDocument/completion` via `window/logMessage`?

If I'm not mistaken, at this current stage of the draft, the server would continue logging every LSP request, maybe even notifications? Aside from doubling the number of requests it would also turn into an endless loop of `window/logMessage` logging `window/logMessage`, would it not?

radeksimko:
FWIW there is a [structured logging package](https://tip.golang.org/doc/go1.21#slog) coming up in Go 1.21 (planned for August '23), currently available via https://pkg.go.dev/golang.org/x/exp/slog which could help answer one of the concerns regarding the log levels (warn, info, debug etc.) and maybe better align the logging with LSP's logging, which has multiple levels too.

I'd be open to finishing this but we should find some answers to the mentioned concerns. Even if the solution isn't ideal we should agree on _a_ solution.

We should also ensure we're not boxing ourselves into a corner when it comes to structured logging which we should IMHO adopt eventually.

It's not super clear in the spec but I am guessing that LSP expects "human-readable" log messages inside the `Message` field in https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_logMessage rather than stringified JSON, so the main benefit we'd gain from structured logging is mostly setting levels, rather than actually logging structured data. 🤔 Correct me if I'm wrong here.

On a related note I was hoping this would present an opportunity to extract/collect some metrics more easily, where we could essentially just send some specific log messages somewhere where we aggregate the structured data inside each such log message. The LSP spec suggests that won't work though, or at least isn't meant to work that way.

PR #1331, opened 2023-07-24
Set codelensReferenceCount to experimental

jpogran:
This moves the `terraform.codelens.referenceCount` setting to `terraform.experimentalFeatures.codelensReferenceCount` to properly reflect the experimental nature of this setting.

While `terraform.codelens.referenceCount` was set to false and declared to be experimental in the README, it was not marked as experimental in the name of the setting. We have been seeing reports of usage where it has been affecting performance and want to ensure that users are opting in with intent to test usage of this feature.

When this feature moves to stable, the previous setting name can return.

Needs https://github.com/hashicorp/vscode-terraform/pull/1545

### DIFF
--- a/internal/settings/settings.go
+++ b/internal/settings/settings.go
@@ -14,8 +14,9 @@ import (
 )
 
 type ExperimentalFeatures struct {
-	ValidateOnSave        bool `mapstructure:"validateOnSave"`
-	PrefillRequiredFields bool `mapstructure:"prefillRequiredFields"`
+	CodelensReferenceCount bool `mapstructure:"codelensReferenceCount"`
+	ValidateOnSave         bool `mapstructure:"validateOnSave"`
+	PrefillRequiredFields  bool `mapstructure:"prefillRequiredFields"`
 }
 
 type Indexing struct {


radeksimko (changes requested):
This is a good start.

For context, the whole feature is currently driven mostly by the LSP capabilities, which on both (client and server) ends is considered experimental. The word "experimental" there though just means it's not part of the LSP spec. It does not communicate the stability of the feature.

https://github.com/hashicorp/terraform-ls/blob/b9eabdc32f40c9494787c3d982c6a5d1cc0f3d4d/internal/protocol/experimental.go#L6-L11
https://github.com/hashicorp/terraform-ls/blob/b9eabdc32f40c9494787c3d982c6a5d1cc0f3d4d/internal/protocol/experimental.go#L13-L29

Making it a (experimental) setting option looks like a good idea - given that our intention is to emphasise the experimental nature of the whole feature. The implication is though that:

a. Clients avoid sending the `showReferencesCommandId` capability if they also set the refcount setting option to true. That already seems to be reflected in https://github.com/hashicorp/vscode-terraform/pull/1545 - so all good there!

b. Server avoids sending `referenceCountCodeLens` capability if the refcount feature is disabled - this is already implied by the above (`showReferencesCommandId` essentially serves as a gating mechanism), but since we do not enforce client's behaviours I think we should go further and ensure the server capability is also _not_ sent when the refcount _feature_ is disabled, for additional safety/integrity.

Finally, is it worth updating the docs - firstly to reflect the changes being made + note the experimental nature of the feature?

https://github.com/hashicorp/terraform-ls/blob/main/docs/language-clients.md#code-lens

PR #1332, opened 2023-07-27
jobs: Don't wait for `GetModuleDataFromRegistry` job

dbanck:
This job may make an HTTP request and we schedule it in the low priority queue so we don't want to wait for it.

This can significantly reduce the request time for the `didOpen` and `didChange` calls.

**Before**
CleanShot 2023-07-27 at 18 03 40@2x

**After**
CleanShot 2023-07-27 at 18 04 07@2x

Another improvement might be not to schedule this job if there are no external module calls in a file.


radeksimko (approved):
LGTM but I think we should at least track an issue for the other bug we found while working on this, where the two reference collection jobs right above run without sufficient schema, resulting in missing references.

PR #1327, opened 2023-07-11
OTEL instrumentation

radeksimko:
Related to https://github.com/hashicorp/terraform-ls/issues/1056

## Example trace

Screenshot 2023-07-18 at 13 57 49

Screenshot 2023-07-18 at 14 53 44

## TODO

 - [ ] Document what is being traced

## Later TODO
 - [ ] Investigate if increasing indexer parallelism makes a difference - separate issue/PR
 - [ ] Report total time to index a URI in walker - separate issue/PR
 - [ ] Report traces with more details from within individual jobs
 - [ ] Raise a separate draft PR for Honeycomb integration

Issue #1056, opened 2022-08-30
Measure performance of indexer(s)


radeksimko:
## Background

The LS (re)indexes files at various stages, as necessary per LSP:

 - `initialize`
 - `textDocument/didOpen`
 - `textDocument/didChange`
 - `textDocument/didChangeWatchedFiles`
 - `workspace/didChangeWorkspaceFolders`

The first one in particular is likely the most impactful on resource usage as it likely to index deeper hierarchy with many modules, the other ones are more likely to work with much more limited scope of 1 or few modules only.

We currently run **benchmarks** of various modules in https://github.com/hashicorp/terraform-ls/blob/2738c6ca944dd3b24dfb431edaabaf2a46cd95b7/.github/workflows/benchmarks.yml#L36-L44

This however only tells us how these particular modules perform. Users may or may not use these modules, or any number of any modules and combination of those. To understand the real impact on real end-users we need some way of measuring the performance **at runtime** (as opposed to in isolation).

Measuring any _aggregate_ performance impact - e.g. by logging total time spent indexing via `initialize` is currently not possible as indexing happens asynchronously. This includes both the filesystem walking and the actual indexing jobs on discovered modules.

We have a mechanism for tracking job IDs in the walker:

https://github.com/hashicorp/terraform-ls/blob/2738c6ca944dd3b24dfb431edaabaf2a46cd95b7/internal/walker/walker.go#L197-L201

but this isn't suitable for runtime (yet) - it's only used in tests. In order to use it at runtime we'd need to track those IDs per walked path.

## Proposal

TODO

PR #1334, opened 2023-07-28
OTEL: Instrument `PreloadEmbeddedSchema` job

radeksimko:
Follow-up from https://github.com/hashicorp/terraform-ls/pull/1327

Related to https://github.com/hashicorp/terraform-ls/issues/1056

One area identified as consuming significant time when working through some data is the `PreloadEmbeddedSchema` job. While this isn't necessarily an indication of a bottleneck in itself, this PR adds some further instrumentation to help gain further visibility into that job.

**Example dataset:**

Screenshot 2023-07-28 at 16 31 40


dbanck (approved):
Nice!

PR #1346, opened 2023-08-04
Introduce a new job for running early validation

### LINKED ISSUE #1338: Introduce early validation module hook
### Language Server Version

v0.31.3

### Problem Statement

After evaluating several solutions for running early validation, we decided that running it as a module hook made the most sense. This ensures that the validation is always done against current module/file data and we don't have to modify several LSP handlers (`didOpen`, `didChange`).

### Proposal

* Introduce a new module hook

```go

func earlyValidation(store *state.StateStore, dNotifier *diagnostics.Notifier) notifier.Hook {
	return func(ctx context.Context, changes state.ModuleChanges) error {
		if changes.IsRemoval {
			// we ignore removed modules for now
			return nil
		}

		isOpen, err := notifier.ModuleIsOpen(ctx)
		if err != nil {
			return err
		}
		if !isOpen {
			// only run validation for open files
			return nil
		}

		mod, err := notifier.ModuleFromContext(ctx)
		if err != nil {
			return err
		}

		// TODO: run early validation logic against module files

		return nil
	}
}
```

* Running early validation as a module hook allows us to validate the whole module at once

---

Part of https://github.com/hashicorp/vscode-terraform/issues/720

dbanck:
This PR adds a new job for running the early validation logic.

After discussing different approaches (e.g., running validation inside a module hook), we decided to use a new job instead. This allows us to decouple the validation from the diagnostic publishing, which still runs as a module hook. The decoupling makes integrating with the existing `terraform validate` logic in the `didSave` handler easier.

It is planned to create a follow-up PR that splits the job into two different jobs. Two jobs will allow us to publish AST-based validations earlier than reference-based validations (which require the results from additional jobs).

Closes #1338


internal/indexer/document_change.go
```diff
@@ -124,6 +124,19 @@ func (idx *Indexer) decodeModule(ctx context.Context, modHandle document.DirHand
 	}
 	ids = append(ids, refOriginsId)
 
+	_, err = idx.jobStore.EnqueueJob(ctx, job.Job{
```
radeksimko, line 127: Is there a thought process behind why we should or should not wait for this job to finish? We seem to wait for all of them except for the one below.

I believe the original idea was that anything that would be affected by the change (such as parsing) and where it matters that we process changes in the right order, we should wait. Technically the client won't _really_ wait in the sense that the UI shouldn't be blocked but it may affect how we process next change notifications.

Here I think we should try to avoid publishing outdated diagnostics.

TL;DR I think we should wait, i.e. return job ID and append to the slice here.

internal/terraform/module/module_ops.go
```diff
@@ -662,6 +662,45 @@ func DecodeVarsReferences(ctx context.Context, modStore *state.ModuleStore, sche
 	return rErr
 }
 
+func EarlyValidation(ctx context.Context, modStore *state.ModuleStore, schemaReader state.SchemaReader, modPath string) error {
+	mod, err := modStore.ModuleByPath(modPath)
+	if err != nil {
+		return err
+	}
+
+	// Avoid validation if it is already in progress or already known
```
radeksimko, line 671:
```suggestion
	// Avoid validation if it is already in progress or already finished
```

internal/indexer/document_change.go
radeksimko: We can observe the UX to later understand the impact of that. I suppose most of the concerns I have about outdated diagnostics could be alleviated by us using the document `version` in our requests

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#publishDiagnosticsParams

```ts
	/**
	 * Optional the version number of the document the diagnostics are published
	 * for.
	 *
	 * @since 3.15.0
	 */
	version?: [integer](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#integer);
```

It still seems wasteful to be computing diagnostics for a version which is already outdated by the time we compute them but maybe that's a different problem that we can't really address here anyway.

So TL;DR feel free to merge as-is, we can look into it later if necessary.

PR #1357, opened 2023-08-10
Implement unreferenced variable validation

jpogran:
Add the ability to use the collected origin and target references in early validation by providing a hook for validation funcs. This also adds a validator for unreferenced variables.

Validation funcs will be provided by terraform-ls for now, but may be moved into hcl-lang in the future.

Original ticket:  https://github.com/hashicorp/hcl-lang/issues/288

Needs https://github.com/hashicorp/hcl-lang/pull/302


radeksimko (commented):
Mostly LGTM aside from my in-line comments. I expect we will find a few other cases of references which are reliable enough to report on - e.g. `locals` but that can be a subject of another future PR!

btw. I expect that once https://github.com/hashicorp/hcl-lang/pull/302 is merged, this PR can also bump `hcl-lang` as a dependency.

radeksimko (approved):

PR #1370, opened 2023-08-31
Only validate a single file on `didChange`

dbanck:
This PR updates the schema validation job to check if it was scheduled by a `didChange` request. This request is sent by a client when a user edits a file.

If the content of a single file changes, it should not be necessary to re-validate all the files in the module. We can reuse our stored diagnostics for most of the module's files, and re-validate only the file being edited.

To achieve this we introduce a new `RPCContext`, that we attach to all jobs.

### DIFF
--- a/internal/context/context.go
+++ b/internal/context/context.go
@@ -5,6 +5,7 @@ package context
 
 import (
 	"context"
+	"path"
 	"time"
 
 	"github.com/hashicorp/terraform-ls/internal/langserver/diagnostics"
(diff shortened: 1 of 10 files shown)


internal/context/context.go
```diff
@@ -20,6 +21,26 @@ func (k *contextKey) String() string {
...
+	}
+}
+
+func (rpcc RPCContextData) IsSingleFileChange() (string, bool) {
+	if rpcc.Method == "textDocument/didChange" {
+		return path.Base(rpcc.URI), true
+	}
+
+	return "", false
+}
```
radeksimko, line 42: I was originally under the impression that we'd want to check for multiple RPC methods but if we're only basically checking for `textDocument/didChange`, would it make more sense to just move this logic into the job and get rid of it here? It feels like unnecessary abstraction layer now.

radeksimko (approved):
Just one nitpick otherwise LGTM.

PR #1372, opened 2023-09-01
langserver: Avoid re-processing open files after file saving

### LINKED ISSUE #1366: Investigate possible reduction on job scheduling on `workspace/didChangeWatchedFiles`
## Context

Currently, as per LSP spec, as interpreted by maintainers - and more importantly - implemented by VS Code language client, the client will send both `textDocument/didSave` after saving a file and _also_ `workspace/didChangeWatchedFiles`.

`workspace/didChangeWatchedFiles` is normally expected to deal with situations where files change outside of the IDE, such as when user runs `git checkout branch-x`, runs any other tooling which mutates the underlying FS tree, or simply edits files manually outside of that one IDE (e.g. uses a different IDE).

The most likely reason a client will send `workspace/didChangeWatchedFiles` alongside `textDocument/didSave` even if the content has not changed is because the metadata on the saved file has changed (i.e. modification timestamp was updated).

Unfortunately, the spec does not allow us to differentiate between metadata changes and content changes. Currently we pessimistically assume these are always content changes and run various jobs with the same assumptions, possibly wasting resources on re-computing things we already know.

## Proposal

Investigate whether it's worth computing and comparing checksums in order to avoid re-computing things for the same version of a document/module.

Computing and comparing checksums may potentially be expensive on its own, so this will require some careful consideration, perhaps use of the OTEL data to ensure we're not just replacing one bottleneck with another.

radeksimko:
Closes #1366

## Context

A common use case for `workspace/didChangeWatchedFiles` is to receive updates about files changed _outside_ of the IDE - such as when the user runs `git checkout` to switch branches.

The LSP spec isn't 100% clear about whether clients should always send `workspace/didChangeWatchedFiles` also for open files. A popular language client (VS Code) always does - i.e. we receive both `textDocument/didSave` on save and then subsequently `workspace/didChangeWatchedFiles` even if the file content has not changed - only the modification timestamp has.

This leads us to processing the exact same files (content-wise) multiple times in row (1st on `textDocument/didChange` 2nd and any other time on `workspace/didChangeWatchedFiles` as followed after `textDocument/didSave`), in turn wasting resources.

## Solution

Reduce the work we do after file saving by assuming that changes have already been processed as part of `textDocument/didChange`.

### DIFF
--- a/internal/langserver/handlers/did_change_watched_files.go
+++ b/internal/langserver/handlers/did_change_watched_files.go
@@ -146,6 +146,20 @@ func (svc *service) DidChangeWatchedFiles(ctx context.Context, params lsp.DidCha
 		}
 
 		if change.Type == protocol.Changed {
+			// Check if document is open and skip running any jobs
+			// as we already did so as part of textDocument/didChange
+			// which clients should always send for *open* documents
+			// even if they change outside of the IDE.
+			docHandle := document.HandleFromURI(rawURI)
+			isOpen, err := svc.stateStore.DocumentStore.IsDocumentOpen(docHandle)
+			if err != nil {
+				svc.logger.Printf("error when checking open document (%q changed): %s", rawURI, err)
+			}
+			if isOpen {
+				svc.logger.Printf("document is open - ignoring event for %q", rawURI)
+				continue
+			}
+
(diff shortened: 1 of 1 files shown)


dbanck (approved):
👍

PR #1379, opened 2023-09-04
validations: Limit ref. validation to only LocalOrigin(s)

radeksimko:
## Context

While testing https://github.com/hashicorp/terraform-ls/pull/1368 we discovered that validating all "matchable" references means we also end up validating cross-module references, such as module inputs.

This, in combination with the fact that we limit lookups to inside of the single module, leads to false diagnostics, such as shown below:

Screenshot 2023-09-04 at 14 23 09

## Proposal

I'm proposing in this PR to treat module input validation as a _feature_ we can enable later, when we had the time to properly assess the validation feature as a whole, and especially its performance impact and for the time being just avoid validating it.

## After patch

Screenshot 2023-09-04 at 14 25 21

## Follow-up

If agreed/merged, we should probably create an issue to track validation of module inputs (DirectOrigin) and also module sources (PathOrigin).

### DIFF
--- a/internal/decoder/validations/unreferenced_origin.go
+++ b/internal/decoder/validations/unreferenced_origin.go
@@ -27,6 +27,13 @@ func UnreferencedOrigins(ctx context.Context, pathCtx *decoder.PathContext) lang
 			continue
 		}
 
+		_, ok = origin.(reference.LocalOrigin)
+		if !ok {
+			// we avoid reporting on origins outside of the current module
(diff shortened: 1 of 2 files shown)


dbanck (changes requested):
LGTM, but I think we should add a simple test case. I can push one, if you like

radeksimko:
@dbanck I added two tests, PTAL.

Issue #653, opened 2021-09-27
Provide references to variables of complex types


radeksimko:
### Current Version

```
0.22.0
```

### Use-cases

See https://discuss.hashicorp.com/t/references-are-no-longer-detected-when-using-complex-types/29789

### Proposal

Modify https://github.com/hashicorp/hcl-lang to ensure that reference targets are collected not only for the block itself, but also for any nested types, if the block itself is represented as `map`, `list`, `object` etc..

amaralc:
This has been an issue here. While trying to enhance the legibility of the code, using objects, I lost the IntelliSense capabilities and ended up losing productivity.

radeksimko:
Hi @amaralc 
There is still some work to do around expressions and complex types (tracked in [this meta issue](https://github.com/hashicorp/vscode-terraform/issues/641)) but we have implemented a bulk of it as part of https://github.com/hashicorp/terraform-ls/releases/tag/v0.31.0

Could you be more specific and mention some examples of what doesn't work for you? Ideally attach a screenshot or a screen recording?

awbx:
Confirming this still reproduces against `terraform-ls` `main` (commit 43f2d8d) and the bundled `v0.39.0-beta4` in vscode-terraform `v2.39.2026030417`.

**Symptom:** completion inside `${...}` template expressions only offers references whose declared type is string-coercible. Composite-typed declared variables (`map`, `list`, `set`, `tuple`, `object`) are filtered out — even though `${var.m_map["k"]}` and `${var.l_list[0]}` are valid Terraform.

**Repro:**
```hcl
variable "s_string" { type = string default = "hello" }
variable "m_map"    { type = map(string) default = { red = "x" } }
variable "l_list"   { type = list(string) default = ["a"] }
variable "o_object" { type = object({ name = string }) default = { name = "ada" } }

# cursor between dot and brace
output "o" { value = "x ${var.|}" }
```
Completion offers only `var.s_string` (and other string-typed scope refs). `var.m_map`, `var.l_list`, `var.o_object` are missing.

**Root cause:** `hashicorp/hcl-lang/decoder/expr_any_template.go:35-38` (and parallel sites at 47-50, 58-61, 70-73) hardcodes the inner constraint as `cty.String`, which makes downstream candidate-gathering filter out anything not directly string-coercible.

[...]

PR #1380, opened 2023-09-04
validations: Validate only first 2 segments of a reference

radeksimko:
## Context

We don't yet support references to complex variables (or local values), which means that validating such references would raise false diagnostics. We can disable validation of such references until https://github.com/hashicorp/terraform-ls/issues/653 is addressed.

## UX Before

Screenshot 2023-09-04 at 16 12 36

## UX After

Screenshot 2023-09-04 at 16 14 41


dbanck (approved):
🙌

Issue #1364, opened 2023-08-29
Reference validation for remaining parts of the language


radeksimko:
## Context

Reference validation was introduced as part of https://github.com/hashicorp/terraform-ls/pull/1357 and further refactored in https://github.com/hashicorp/terraform-ls/pull/1346

In short, this produces the following diagnostics

![Screenshot 2023-08-29 at 14 41 28](https://github.com/hashicorp/terraform-ls/assets/287584/3b558ac8-18ca-43fb-8b93-c650cdfcb51c)

We have decided to implement only validation of variables, i.e. `var.*` references. All other references are currently excluded from validation.

 - local values (`local.*`)
 - data sources (`data.*`)
 - resources
 - modules
 - "extensions"
   - `self.*` references
   - `each.*`
   - `count.index`
 - provider aliases (e.g. `aws.east` in a `provider` attribute of a `resource`)

## Proposal

 - TODO: local values issue
 - TODO: resources issue
 - TODO: data sources issue
 - TODO: modules issue
 - TODO: provider aliases issue
 - TODO: extensions issue

PR #1381, opened 2023-09-04
validations: Enable reference validation of locals

radeksimko:
This addresses one small part of https://github.com/hashicorp/terraform-ls/issues/1364 - in particular `local.*` reference validation.

## UX Before

Screenshot 2023-09-04 at 16 24 08

## UX After

Screenshot 2023-09-04 at 16 24 33

### DIFF
--- a/internal/decoder/validations/unreferenced_origin.go
+++ b/internal/decoder/validations/unreferenced_origin.go
@@ -6,6 +6,7 @@ package validations
 import (
 	"context"
 	"fmt"
+	"slices"
 
 	"github.com/hashicorp/hcl-lang/decoder"
 	"github.com/hashicorp/hcl-lang/lang"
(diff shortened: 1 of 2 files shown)


internal/decoder/validations/unreferenced_origin.go
```diff
@@ -44,11 +44,11 @@ func UnreferencedOrigins(ctx context.Context, pathCtx *decoder.PathContext) lang
...
 		}
 
-		// we only initially validate variables
+		// we only initially validate variables & local values
 		// resources and data sources can have unknown schema
 		// and will be researched at a later point
-		firstStep := address[0]
-		if firstStep.String() != "var" {
+		firstStep := address[0].String()
+		if firstStep != "var" && firstStep != "local" {
```
dbanck, line 51:
```suggestion
		supported := []string{"var", "local"}
		firstStep := address[0].String()
		if !slices.Contains(supported, firstStep) {
```
With Go 1.21 we could use `slices.Contains` here, but it's up to you to decide if it's more readable

dbanck (approved):
👍  You're on fire today with all of these fixes

PR #1394, opened 2023-09-07
indexer: Ensure declared module calls get decoded

radeksimko:
Related: https://github.com/hashicorp/terraform-schema/pull/254

The PR is not strictly hard dependency for this one but it is helpful in testing, to verify that the race condition is no longer present.

--- 

## Context

Some time ago, we discovered a race condition, which may not have been as well understood at the time. We needed to make sure that go-to-definition and go-to-references works on first load for module inputs.

Module inputs are represented as reference origins and as with everything else, the attribute schema is what instructs the job collecting the reference origins how and whether to collect those origins. Obtaining the schema for _modules_ is basically the job of [`LoadModuleMetadata`](https://github.com/hashicorp/terraform-ls/blob/da68b876c05ac9b1e46697650d4785da8d498ed0/internal/terraform/module/module_ops.go#L494C8-L537) - in particular we generate the schema based on parsed variables & outputs.

Prior to this PR, we never processed "module calls" directly from `didOpen` or `didChange` and only relied on the walker to get to them by the time we needed them. This would usually work fine if no documents were open at the time of initial walking. Often times though the walking may take a lot of time for larger workspaces and users may open files very early. This meant submodule data was unavailable when the user opened the module which calls those submodules, making the reference origin collection job "clue-less" about those module inputs.

We worked around that race condition by inferring the inputs - basically assuming all declared inputs have their corresponding variable declarations. As with most workarounds, this eventually caught up with us 😅 . In particular, validation (as implemented in https://github.com/hashicorp/terraform-ls/pull/1368) was also acting based on this inferred (potentially incorrect) schema and worse - the inferred schema was assumed to be relevant for all modules with the same `source`.

Therefore, https://github.com/hashicorp/terraform-schema/pull/254 removes the workaround and this PR addresses the race condition.

Technically this may result in more jobs getting scheduled but the overall performance impact should be relatively minimal because:

 - it only impacts open files/modules and the assumption is that users generally don't open _all_ modules that are in their workspace at the same time
 - the jobs exit early if the module data is already available

### DIFF
--- a/internal/indexer/document_change.go
+++ b/internal/indexer/document_change.go
@@ -76,25 +76,49 @@ func (idx *Indexer) decodeModule(ctx context.Context, modHandle document.DirHand
 		Type:        op.OpTypeLoadModuleMetadata.String(),
 		DependsOn:   dependsOn,
 		IgnoreState: ignoreState,
+		Defer: func(ctx context.Context, jobErr error) (jobIds job.IDs, err error) {
+			if jobErr != nil {
+				err = jobErr
+				return
+			}
+			modCalls, mcErr := idx.decodeDeclaredModuleCalls(ctx, modHandle, ignoreState)
+			if mcErr != nil {
+				idx.logger.Printf("decoding declared module calls for %q failed: %s", modHandle.URI, mcErr)
+				// We log the error but still continue scheduling other jobs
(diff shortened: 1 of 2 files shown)


radeksimko:
Re-opening from a different branch

PR #1404, opened 2023-09-13
Update ParseModuleConfiguration to only parse changed file

jpogran:
This modifies ParseModuleConfiguration to only parse changed file if file already parsed and this is a didChange request.

Previously ParseModuleConfiguration would reparse not only the changed file but also the module containing the changed file.

Now ParseModuleConfiguration detects if the file has previously been parsed and the request is from a didChange event. If both are true, then it only parses the changed file and skips re-parsing the entire module. If this is false, it falls back to the previous behavior of re-parsing the entire module including the changed file.

### DIFF
--- a/internal/terraform/module/module_ops.go
+++ b/internal/terraform/module/module_ops.go
@@ -374,7 +374,18 @@ func ParseModuleConfiguration(ctx context.Context, fs ReadOnlyFS, modStore *stat
 	if mod.ModuleParsingState == op.OpStateLoaded && rpcContext.IsDidChangeRequest() {
 		// the file has already been parsed, so only examine this file and not the whole module
 		fileName, _ := uri.PathFromURI(rpcContext.URI)
-		files, diags, err = parser.ParseModuleFile(fs, fileName)
+
+		f, fDiags, err := parser.ParseModuleFile(fs, fileName)
+		if err != nil {
+			return err
+		}
+		existingFiles := mod.ParsedModuleFiles.Copy()
+		existingFiles[ast.ModFilename(fileName)] = f
+		files = existingFiles
+
+		existingDiags := mod.ModuleDiagnostics.Copy()
+		existingDiags[ast.ModFilename(fileName)] = fDiags
+		diags = existingDiags
 	} else {
 		// this is the first time file is opened so parse the whole module
 		files, diags, err = parser.ParseModuleFiles(fs, modPath)
--- a/internal/terraform/module/module_ops.go
+++ b/internal/terraform/module/module_ops.go
@@ -18,6 +18,7 @@ import (
 	"github.com/hashicorp/hcl-lang/decoder"
 	"github.com/hashicorp/hcl-lang/lang"
 	tfjson "github.com/hashicorp/terraform-json"
+	lsctx "github.com/hashicorp/terraform-ls/internal/context"
 	idecoder "github.com/hashicorp/terraform-ls/internal/decoder"
 	"github.com/hashicorp/terraform-ls/internal/document"
 	"github.com/hashicorp/terraform-ls/internal/job"
@@ -25,9 +26,11 @@ import (
 	"github.com/hashicorp/terraform-ls/internal/registry"
 	"github.com/hashicorp/terraform-ls/internal/schemas"
 	"github.com/hashicorp/terraform-ls/internal/state"
+	"github.com/hashicorp/terraform-ls/internal/terraform/ast"
 	"github.com/hashicorp/terraform-ls/internal/terraform/datadir"
 	op "github.com/hashicorp/terraform-ls/internal/terraform/module/operation"
 	"github.com/hashicorp/terraform-ls/internal/terraform/parser"
+	"github.com/hashicorp/terraform-ls/internal/uri"
 	tfaddr "github.com/hashicorp/terraform-registry-address"
 	"github.com/hashicorp/terraform-schema/earlydecoder"
 	"github.com/hashicorp/terraform-schema/module"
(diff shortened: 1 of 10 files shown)


internal/terraform/module/module_ops.go
```diff
@@ -359,19 +363,47 @@ func ParseModuleConfiguration(ctx context.Context, fs ReadOnlyFS, modStore *stat
...
 	if mod.ModuleParsingState != op.OpStateUnknown && !job.IgnoreState(ctx) {
 		return job.StateNotChangedErr{Dir: document.DirHandleFromPath(modPath)}
 	}
 
-	err = modStore.SetModuleParsingState(modPath, op.OpStateLoading)
+	var files ast.ModFiles
+	var diags ast.ModDiags
+	rpcContext := lsctx.RPCContext(ctx)
+	// Only parse the file that's being changed/opened, unless this is 1st-time parsing
+	if mod.ModuleParsingState == op.OpStateLoaded && rpcContext.IsDidChangeRequest() {
```
radeksimko, line 375:
```suggestion
	if mod.ModuleParsingState == op.OpStateLoaded && rpcContext.IsDidChangeRequest() && lsctx.IsLanguageId(ctx, ilsp.Terraform.String()) {
```
This is important so that we don't end up parsing tfvars files as if they're part of the module ("module" meaning the collection of `*.tf` and `*.tf.json` files).

On a related note, there's an opportunity for another future optimisation PR similar to this one, which involves `ParseVariables` job below.

internal/terraform/module/module_ops_test.go
```diff
@@ -931,6 +941,78 @@ func TestPreloadEmbeddedSchema_raceCondition(t *testing.T) {
...
+
+	// ignore job state
+	ctx = job.WithIgnoreState(ctx, true)
+
+	// say we're coming from did_change request
+	fooURI, _ := filepath.Abs(filepath.Join(singleFileModulePath, "foo.tf"))
+	x := lsctx.RPCContextData{
+		Method: "textDocument/didChange",
+		URI:    uri.FromPath(fooURI),
+	}
```
radeksimko, line 983: It would be nice (but not a must) to have one more test with `didChange` request for `*.tfvars` to confirm that nothing unexpected happens inside that job.

internal/terraform/module/module_ops.go
```diff
@@ -359,19 +363,47 @@ func ParseModuleConfiguration(ctx context.Context, fs ReadOnlyFS, modStore *stat
...
 	// TODO: Avoid parsing if the content matches existing AST
 
-	// TODO: Only parse the file that's being changed/opened, unless this is 1st-time parsing
-
 	// Avoid parsing if it is already in progress or already known
 	if mod.ModuleParsingState != op.OpStateUnknown && !job.IgnoreState(ctx) {
 		return job.StateNotChangedErr{Dir: document.DirHandleFromPath(modPath)}
 	}
 
-	err = modStore.SetModuleParsingState(modPath, op.OpStateLoading)
```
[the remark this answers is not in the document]
jpogran: Correct, I missed adding it back. However I have to duplicate it inside the decision statement in order to correctly catch the existing StateLoaded state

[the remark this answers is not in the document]
jpogran: Actually, @radeksimko do we want to set the module parsing state to loading when we're only parsing a single file? Or am I reading too much into the name 'ModuleParsingState'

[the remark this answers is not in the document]
radeksimko: This should basically only affect the condition above. Given that we always ignore the state when scheduling the job here https://github.com/hashicorp/terraform-ls/blob/94e47bd3a6371c6d56c2ab92d0d33b1ce84e9c0d/internal/indexer/document_change.go#L25 _and_ because the other branch already sets it to `Loading` here and `Loaded` here https://github.com/hashicorp/terraform-ls/blob/94e47bd3a6371c6d56c2ab92d0d33b1ce84e9c0d/internal/state/module.go#L858-L860
it shouldn't matter too much.

I think it's probably better to _not_ be setting it to `Loading` on single file parsing.

internal/terraform/module/module_ops_test.go
```diff
@@ -931,6 +942,141 @@ func TestPreloadEmbeddedSchema_raceCondition(t *testing.T) {
...
+	before, err := ss.Modules.ModuleByPath(singleFileModulePath)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// ignore job state
+	ctx = job.WithIgnoreState(ctx, true)
+
+	// say we're coming from did_change request
+	fooURI, _ := filepath.Abs(filepath.Join(singleFileModulePath, "foo.tf"))
```
radeksimko, line 980:
```suggestion
	fooURI, err := filepath.Abs(filepath.Join(singleFileModulePath, "foo.tf"))
	if err != nil {
		t.Fatal(err)
	}
```

internal/terraform/module/module_ops_test.go
```diff
@@ -931,6 +942,141 @@ func TestPreloadEmbeddedSchema_raceCondition(t *testing.T) {
...
+	before, err := ss.Modules.ModuleByPath(singleFileModulePath)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// ignore job state
+	ctx = job.WithIgnoreState(ctx, true)
+
+	// say we're coming from did_change request
+	fooURI, _ := filepath.Abs(filepath.Join(singleFileModulePath, "example.tfvars"))
```
radeksimko, line 1052:
```suggestion
	fooURI, err := filepath.Abs(filepath.Join(singleFileModulePath, "example.tfvars"))
	if err != nil {
		t.Fatal(err)
	}
```

internal/terraform/module/module_ops_test.go
```diff
@@ -931,6 +942,141 @@ func TestPreloadEmbeddedSchema_raceCondition(t *testing.T) {
...
+	x := lsctx.RPCContextData{
+		Method: "textDocument/didChange",
+		URI:    uri.FromPath(fooURI),
+	}
+	ctx = lsctx.WithLanguageId(ctx, ilsp.Tfvars.String())
+	ctx = lsctx.WithRPCContext(ctx, x)
```
radeksimko, line 1058:
```suggestion
	ctx = lsctx.WithLanguageId(ctx, ilsp.Tfvars.String())
	ctx = lsctx.WithRPCContext(ctx, lsctx.RPCContextData{
		Method: "textDocument/didChange",
		URI:    uri.FromPath(fooURI),
	})
```

internal/terraform/module/module_ops_test.go
```diff
@@ -931,6 +942,141 @@ func TestPreloadEmbeddedSchema_raceCondition(t *testing.T) {
...
+
+	// example.tfvars should be the same
+	if before.ParsedModuleFiles["example.tfvars"] != after.ParsedModuleFiles["example.tfvars"] {
+		t.Fatal("file should match")
+	}
+
+	// diags should not change for example.tfvars
+	if before.ModuleDiagnostics["example.tfvars"] != nil {
+		t.Fatal("there should be no diags for tfvars files right now")
+	}
```
radeksimko, line 1077:
```suggestion
	// example.tfvars should be missing
	_, beforeExists := before.ParsedModuleFiles["example.tfvars"]
	if beforeExists {
		t.Fatal("example.tfvars file should be missing")
	}
	_, afterExists := after.ParsedModuleFiles["example.tfvars"]
	if afterExists {
		t.Fatal("example.tfvars file should be missing")
	}

	// diags should be missing for example.tfvars
	if _, ok := before.ModuleDiagnostics["example.tfvars"]; ok {
		t.Fatal("there should be no diags for tfvars files right now")
	}
```
I'm kind of surprised this isn't crashing, because there is no entry for `example.tfvars` in that map, but I suppose we just end up comparing two `nil`s.

Regardless though I think that map entry existence is what we should be testing for.

radeksimko (approved):
Mostly minor nitpicky suggestions, feel free to merge once applied.

PR #1412, opened 2023-09-18
context: Introduce `LanguageId`

radeksimko:


dbanck (commented):
Why do you use a `string` instead of `ilsp.LanguageID` in the function signatures? I guess it's unlikely that the string `"terraform"` will ever change, but it feels cleaner to me to use the constant.

With the new context methods here, can we get rid of https://github.com/hashicorp/terraform-ls/blob/main/internal/decoder/context.go?

radeksimko:
@dbanck I'm not against it but it goes much further than I was hoping to go with the PR due to conflicts.

Meaning, that we should update it here https://github.com/hashicorp/terraform-ls/blob/f8445e8e28239c7bf904e5f0f28a22935755f54f/internal/document/document.go#L18 and anywhere downstream from there + in the `didOpen` handler we should also add a switch statement rather than be casting an arbitrary string to the desired type.

That is to say it's the **right** thing to do but I'm not sure it's worth doing _right now_ due to the long-running branch. 😅

Issue #1410, opened 2023-09-18
Provide diagnostics in `tfvars` files


dbanck:

radeksimko:
This was implemented as part of https://github.com/hashicorp/terraform-ls/pull/1413 and the "nice to haves" can be followed in individual issues as linked above.

PR #1413, opened 2023-09-19
Introduce validation of `*.tfvars` files

radeksimko:
Depends on:
 - https://github.com/hashicorp/terraform-ls/pull/1412
 - https://github.com/hashicorp/hcl-lang/pull/318
 - https://github.com/hashicorp/terraform-ls/pull/1404 (strictly speaking it's "just" a performance improvement but it would likely make rebasing easier if it was merged first)

Closes https://github.com/hashicorp/terraform-ls/issues/1410

--- 

## UX Examples

Screenshot 2023-09-19 at 09 47 37

Screenshot 2023-09-19 at 09 48 31

### DIFF
--- a/internal/state/module.go
+++ b/internal/state/module.go
@@ -988,7 +988,7 @@ func (s *ModuleStore) SetModuleDiagnosticsState(path string, source ast.Diagnost
 func (s *ModuleStore) UpdateVarsDiagnostics(path string, source ast.DiagnosticSource, diags ast.VarsDiags) error {
 	txn := s.db.Txn(true)
 	txn.Defer(func() {
-		s.SetVarsDiagnosticsState(path, ast.HCLParsingSource, op.OpStateLoaded)
+		s.SetVarsDiagnosticsState(path, source, op.OpStateLoaded)
 	})
 	defer txn.Abort()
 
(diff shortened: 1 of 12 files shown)


internal/state/module.go
```diff
@@ -988,7 +988,7 @@ func (s *ModuleStore) SetModuleDiagnosticsState(path string, source ast.Diagnost
 func (s *ModuleStore) UpdateVarsDiagnostics(path string, source ast.DiagnosticSource, diags ast.VarsDiags) error {
 	txn := s.db.Txn(true)
 	txn.Defer(func() {
-		s.SetVarsDiagnosticsState(path, ast.HCLParsingSource, op.OpStateLoaded)
+		s.SetVarsDiagnosticsState(path, source, op.OpStateLoaded)
```
radeksimko, line 991: The correct source is already being passed in https://github.com/hashicorp/terraform-ls/blob/69cc1b857dc07f4c23df5f24bcf597202125f2a2/internal/terraform/module/module_ops.go#L422 so this seems to be just a bug which was never surfaced as we never had any other sources up to this point.

dbanck: Good catch!

dbanck (approved):
👍

PR #1426, opened 2023-09-29
Update early validation setting to `enabled` by default

dbanck:
This PR renames and enables the early validation setting by default. We use https://github.com/mcuadros/go-defaults to set a default value.

We could also rename the JSON `earlyValidation` key if we wanted, but in case that case we would need to update https://github.com/hashicorp/vscode-terraform/pull/1554 as well


radeksimko (approved):
LGTM but I would also let @jpogran review this as he has more context about how this would/wouldn't work well in VS Code as we mostly mirror the settings names and structure.

jpogran (approved):
🚢

PR #1427, opened 2023-09-29
context: Replace `RPCContextData` with `DocumentContext`

radeksimko:
This ensures that we propagate the language ID between job scheduling and execution, which in turn enables https://github.com/hashicorp/terraform-ls/pull/1404 at runtime (and not just in unit tests).

While I am labelling this as a bug, it's a bug that was never surfaced to the user in a release, so it does not need to be mentioned in the Changelog AFAICT.

### DIFF
--- a/internal/langserver/handlers/did_open.go
+++ b/internal/langserver/handlers/did_open.go
@@ -8,7 +8,6 @@ import (
 	"fmt"
 
 	"github.com/creachadair/jrpc2"
-	lsctx "github.com/hashicorp/terraform-ls/internal/context"
 	"github.com/hashicorp/terraform-ls/internal/document"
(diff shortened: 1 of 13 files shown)


internal/langserver/handlers/did_open.go
```diff
@@ -38,8 +37,6 @@ func (svc *service) TextDocumentDidOpen(ctx context.Context, params lsp.DidOpenT
 		return err
 	}
 
-	ctx = lsctx.WithLanguageId(ctx, params.TextDocument.LanguageID)
```
radeksimko, line 41: I don't know if there's actually any other method that has the same structure that will benefit from this but given that we already do some unmarshaling prior to passing the request to the handler I assume we may as well unmarshal the LanguageID there instead of here.

dbanck (approved):
👍

PR #1447, opened 2023-10-05
fix: Avoid validating *.tfvars outside of modules

radeksimko:
Originally reported in https://github.com/hashicorp/vscode-terraform/issues/1574

--- 

Prior to this patch we would consider any folder with `*.tfvars` as a module. Users may however place `*.tfvars` files into folders to be loaded anywhere else. In that case we cannot make any assumptions about those `*.tfvars` files and hence we cannot validate them.


internal/decoder/decoder.go
```diff
@@ -64,7 +64,13 @@ func varsPathContext(mod *state.Module) (*decoder.PathContext, error) {
...
 		Files:            make(map[string]*hcl.File),
-		Validators:       varsValidators,
+	}
+
+	if len(mod.ParsedModuleFiles) > 0 {
+		// Only validate if this is actually a module
+		// as we may come across standalone tfvars files
+		// for which we have no context.
+		pathCtx.Validators = varsValidators
 	}
```
jpogran, line 74: neat!

jpogran (approved):

PR #1453, opened 2023-10-10
Resolve version correctly for backend schema

radeksimko:
This is to address:
 - https://github.com/hashicorp/vscode-terraform/issues/1580

Depends on:

 - https://github.com/hashicorp/terraform-schema/pull/275

--- 

This PR changes the meaning of the _version_ in the snippet of code being changed. Previously it referred to the discovered/installed Terraform version, whether or not we had the schema for it. With this patch, the meaning changes to "language/schema version".

In practice though it remains the same for the likely majority of cases (any stable version between `0.12` and `1.6`) but not for the other cases, where we fall back on either 0.12.0 or latest known version (currently 1.6.0).

Importantly, this does _not_ impact the discovered/installed version we persist and make available, e.g. here

Also, the new logic inside `ResolveVersion` reflects the way we handled various edge cases in the previous implementation, as shown in the attached test in https://github.com/hashicorp/terraform-schema/pull/275

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/hashicorp/terraform-exec v0.19.0
 	github.com/hashicorp/terraform-json v0.17.1
 	github.com/hashicorp/terraform-registry-address v0.2.2
-	github.com/hashicorp/terraform-schema v0.0.0-20231006123128-72331e948317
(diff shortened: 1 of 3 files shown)


dbanck (approved):
I like this approach

PR #1456, opened 2023-10-11
fix: Work around 'unreliable' input data for Registry modules

radeksimko:
As discovered in https://github.com/hashicorp/vscode-terraform/issues/1582 it turns out that we assume inputs in some modules as required, when they are actually not.

 - Why? - Because we use data from the Registry API, which (incorrectly) indicates some inputs as required.
 - Which ones? - Modules which have "nullable" inputs (variables with `default = null`)
 - Why? - Because the Registry API did not recognise "nullable" inputs at the time of ingesting the module.
 - Can't the Registry fix this? - The Registry already recognises "nullable" inputs since 19th August 2022.
 - Can't the Registry fix the data then? - This was considered but it involves potentially a _lot_ of modules and module versions, so this is a non-trivial task, esp. given that this would involve basically re-ingesting the modules/versions.

So this ^ is why we're taking a more pragmatic approach and basically treat any _required_ inputs as _optional_ for any modules which we know were published prior to the date when the Registry implemented support for "nullable" inputs.

Importantly, this does not impact recently published modules and module versions. Those will continue to have inputs reflected as-is.

### DIFF
--- a/internal/registry/module.go
+++ b/internal/registry/module.go
@@ -11,6 +11,7 @@ import (
 	"net/http"
 	"net/http/httptrace"
 	"sort"
+	"time"
 
 	"github.com/hashicorp/go-version"
 	tfaddr "github.com/hashicorp/terraform-registry-address"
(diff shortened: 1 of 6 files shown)


dbanck (approved):
Looks like a good pragmatic solution to me

PR #1485, opened 2023-11-03
decoder: validate required attributes more selectively

radeksimko:
This is one way to dampen the impact of https://github.com/hashicorp/vscode-terraform/issues/1616 although it does not really solve it entirely.


dbanck (approved):
👍 

Feel free to merge after bumping hcl-lang

PR #1522, opened 2023-12-18
Alert on wsl.localhost Path

jpogran:
This adds path detection for network WSL paths to warn the user to open the path in WSL instead.

A user can open a WSL path directly using the network mount path, which results in opening inside the host VS Code instance instead of inside the WSL instance. This is a problem for terraform-ls, which can't resolve the network paths correctly.


dbanck (approved):
Nice, thanks!

PR #1529, opened 2023-12-19
Rename TokenTypeCapsule to TokenTypeComplex

jpogran:
This renames the existing token type TokenTypeCapsule to TokenTypeComplex to aid readability and reflect existing naming conventions.

This token represents `list` in `type = list(string)` and similar, i.e. the name of the complex type. The term "capsule" has slightly different meaning in cty and HCL and we should not use it in this context.

Pending https://github.com/hashicorp/hcl-lang/pull/352


radeksimko (commented):
I know this is still a draft but just FYI we'll also need to update the docs [here](https://github.com/hashicorp/terraform-ls/blob/a2e542e112238eadb4b11becefa2757823b57194/docs/syntax-highlighting.md?plain=1#L57)

radeksimko (approved):
LGTM, but would you mind also drafting a PR in vscode-terraform, just to ensure we don't forget about it, prior to merging?

PR #1546, opened 2023-12-22
Replace `mitchellh/*` with maintained forks

dbanck:
- [x] Replace `mitchellh/cli` with `hashicorp/cli`
- [ ] Replace `mitchellh/mapstructure` with `go-viper/mapstructure`
- [ ] Replace `mitchellh/go-homedir` with ?


radeksimko:
> Replace `mitchellh/cli` with `hashicorp/cli`

👍🏻 

> Replace `mitchellh/mapstructure` with `go-viper/mapstructure`

👍🏻 

> Replace `mitchellh/go-homedir` with ?

I think we may be able to yank that dependency altogether. It needs double checking but since the library was born, there was some more work done in stdlib, specifically in `os`. [Assuming it no longer requires CGO](https://github.com/mitchellh/go-homedir/issues/20), the `homedir.Dir()` part could be replaced with [`os.UserHomeDir()`](https://pkg.go.dev/os#UserHomeDir). The stdlib still doesn't provide a convenience equivalent `Expand()` function but we could just copy [these few LOC](https://github.com/mitchellh/go-homedir/blob/af06845cf3004701891bf4fdb884bfe4920b3727/homedir.go#L58-L77) out into our codebase. 🤷🏻 

There is still a risk of missing some edge cases described by Mitchell in https://github.com/mitchellh/go-homedir/issues/20#issuecomment-467476994 but I think the cost vs risk here is in our favour given the size of our user base compared to e.g. Vagrant or Terraform (CLI), i.e. it's statistically less likely our users will run into these problems.

Issue #1530, opened 2023-12-20
New Code Action: Fill block with optional+required attributes


radeksimko:
## Background

The following two PRs introduced support for pre-filling required attributes on completion:

 - https://github.com/hashicorp/terraform-ls/pull/657
 - https://github.com/hashicorp/vscode-terraform/pull/799

which can be enabled via [`prefillRequiredFields`](https://github.com/hashicorp/terraform-ls/blob/main/docs/SETTINGS.md#prefillrequiredfields-bool) and corresponding setting in VS Code `terraform.experimentalFeatures.prefillRequiredFields`. It looks like this in action:

![2023-12-20 09 54 45](https://github.com/hashicorp/terraform-ls/assets/287584/0d9d1b4b-ae97-4220-b3e5-ea25c50042ea)

One of the problems with the above approach is that it's unclear what the user wants to do and what the completion should do when there are pre-existing attributes in the block body. There may be contexts where overwriting it is correct and other contexts where just adding missing attributes is correct. In the specific example above it would most likely make more sense to overwrite things because two different resources are unlikely to have common attributes. This may not be the case for all blocks (other than `resource`) though.

Additionally, manipulating larger ranges of configuration during completion, especially if they require a lot of context is not common in LSP and similar problems in other servers are more frequently implemented as code actions. The LSP makes this difficult (intentionally or not) by mandating textEdits to be computed by the time the completion is requested and these cannot be resolved at a later stage via `completionItem/resolve`, [as the spec says](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion):

> All other properties (usually sortText, filterText, insertText and textEdit) must be provided in the [textDocument/completion](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion) response and must not be changed during resolve.

This in turn can make such pre-filling mechanism very expensive in the sense that the server pre-computes text edits for potentially hundreds of options (and send all of them through RPC) when only one of them will be chosen and the rest is thrown away.

Another example where similar approach won't work for the reasons above is pre-filling of inputs in `module` block based on `source` because it would be prohibitevely expensive to download all metadata from the Registry API for all modules the user may wish to complete.

Code actions could provide a different and more flexible way (from maintenance perspective) to solve this problem, even if that comes at the cost of different UX for the user.

Example of a similar action in gopls:
[...]

PR #1549, opened 2024-01-04
feat: add a first code-action quickfix

kevineor:
### Related issues
https://github.com/hashicorp/vscode-terraform/issues/801
https://github.com/hashicorp/terraform-ls/issues/1530

### Proposal

The implementation relies on `Diagnostics` data property : https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnostic

This property is stated as an extra data that is kept between `textDocument/publishDiagnostics` notifications and `textDocument/codeAction` requests, this permits to avoid the need of parsing again the document.

This could be proposed for other issues that are pushed through Diagnostics

### Actual state

Jan-04-2024 21-41-52

Currently filling the missing attribute using a `null` value, it may be possible to use the default value, when one is provided.


jeremychauvet:
Waw!

radeksimko (changes requested):
There is a plan for introducing some abstraction layers first before we begin to introduce more code actions. This is to ensure they all stay reasonably maintainable as we can anticipate many more will come.

I outlined some more details in https://github.com/hashicorp/terraform-ls/issues/1575 but this is mostly for transparency, _not_ in the anticipation of an external contributor picking up any of that work. We do expect external contributors like you to participate in maintaining those individual code actions however, which will hopefully make use of the abstraction layers.

I know this may not be the answer you were hoping for but I prefer to be honest and open, even if that implies high likelihood of the PR being closed in the near term and/or need significant rewrite later.

For now I'm putting it on hold and a team member will revisit this PR later. I cannot commit to any timeline at this point unfortunately.

Thank you for your effort and understanding!

PR #1584, opened 2024-01-15
Stage 0.32.5

dbanck:


CHANGELOG.md
```diff
@@ -1,3 +1,30 @@
...
+* build(deps): Bump github.com/hashicorp/terraform-exec from 0.19.0 to 0.20.0 ([#1543](https://github.com/hashicorp/terraform-ls/pull/1543))
+* build(deps): Bump github.com/vektra/mockery/v2 from 2.38.0 to 2.40.1 ([#1542](https://github.com/hashicorp/terraform-ls/pull/1542), [#1544](https://github.com/hashicorp/terraform-ls/pull/1544), [#1576](https://github.com/hashicorp/terraform-ls/pull/1576), [#1581](https://github.com/hashicorp/terraform-ls/pull/1581))
+* build(deps): Bump workflows to latest trusted versions ([#1520](https://github.com/hashicorp/terraform-ls/pull/1520), [#1571](https://github.com/hashicorp/terraform-ls/pull/1571), [#1579](https://github.com/hashicorp/terraform-ls/pull/1579))
+* Rename `TokenTypeCapsule` to `TokenTypeComplex` ([#1529](https://github.com/hashicorp/terraform-ls/pull/1529))
+* Rename `CandidatesAtPos` to `CompletionAtPos` ([#1548](https://github.com/hashicorp/terraform-ls/pull/1548))
+* Rename `*Traversal*` to `*Reference*` ([#1574](https://github.com/hashicorp/terraform-ls/pull/1574))
```
dbanck, line 26:
```suggestion
```

Do we want to omit these refactorings?

radeksimko: The first one I think we should certainly include because it may imply changes on the client side - strictly speaking it may even be thought of as a breaking change but given that the logic has fallback it may be fine. Also, as far as I'm aware there aren't other clients leveraging the _custom_ tokens (which would imply custom HCL/Terraform specific themes) as of yet.

The other two however we can omit.

PR #1637, opened 2024-02-15
chore: improve error message if test fails due to local generated schemas

ansgarm:
This is helping future me remember why this test might fail


dbanck:
Superseded by https://github.com/hashicorp/terraform-ls/pull/1638

PR #1643, opened 2024-02-22
Remove getting provider schema log line

dbanck:
This PR removes the log line for accessing a provider schema in memory. We do this a lot and it causes a lot of noise in the log output.

It's also confusing because it might imply that there is an intensive operation involved, such as an HTTP request, which is not the case.


ansgarm (approved):
Oh, nice! Stumbled upon this already as well. Was even confused at first 😄

Issue #1671, opened 2024-03-27
Provider-defined functions are available for Terraform versions < 1.8


dbanck:
### Language Server Version

v0.33.0-beta2

### Terraform Version

1.7.5

### Client Version

VS Code 1.87.2

### Terraform Configuration

```terraform
terraform {
  required_providers {
    time = {
      source  = "hashicorp/time"
      version = "0.11.1"
    }
    terraform = {
      source = "terraform.io/builtin/terraform"
    }
    local = {
      source  = "hashicorp/local"
      version = "2.5.1"
    }
  }
}

output "test" {
  value = provider::local::direxists()
}
```

### Steps to Reproduce

1. Have Terraform 1.7.5 installed
2. Use the language server 0.33.0-beta2

### Expected Behavior

Provider-defined functions included in the preloaded schemas should not be available

### Actual Behavior

Provider-defined functions are available

### Help Wanted

- [X] I'm interested in contributing a fix myself

### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

PR #1672, opened 2024-03-28
Pass Terraform version to functions merger

dbanck:
* Depends on https://github.com/hashicorp/terraform-schema/pull/336
* Closes https://github.com/hashicorp/terraform-ls/issues/1671

---

Prior to this PR, we would always merge any provider-defined functions included in a schema with the Terraform core functions. Even if the Terraform version doesn't support them yet.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/hashicorp/terraform-exec v0.20.0
 	github.com/hashicorp/terraform-json v0.21.0
 	github.com/hashicorp/terraform-registry-address v0.2.3
-	github.com/hashicorp/terraform-schema v0.0.0-20240326154150-560df6bbc4dc
+	github.com/hashicorp/terraform-schema v0.0.0-20240327170238-ec799b9b2e3e
 	github.com/mcuadros/go-defaults v1.2.0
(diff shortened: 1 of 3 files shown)


jpogran (approved):

go.mod
```diff
@@ -18,7 +18,7 @@ require (
 	github.com/hashicorp/terraform-exec v0.20.0
 	github.com/hashicorp/terraform-json v0.21.0
 	github.com/hashicorp/terraform-registry-address v0.2.3
-	github.com/hashicorp/terraform-schema v0.0.0-20240326154150-560df6bbc4dc
```
ansgarm, line 21: Since the last release of `terraform-schema` happened 8 months ago, I suppose we can just update this to the (rebased) commit on `main`, right? (I just merged the other PR that this PR depends on)

ansgarm: updated to `364ac1` which is the current `main` (after the "depends on"-PR named above had been merged)

ansgarm:
(rebased onto `main`)

go.mod
dbanck: Yes, that is the process we usually follow:
- Merge the dependent PR in terraform-schema
- Rebase this PR with the new commit hash from terraform-schema's `main` branch

To make the test pass during review, we first pin it to the PR commit

PR #1708, opened 2024-05-15
module: remove deadcode for parsing error

giautm:


giautm:
cc @dbanck, here is another PR.

dbanck:
I've seen this PR, thank you for raising it.

I'm still thinking about whether we should stop collecting the parsing error or rather want to remove the early return in L507.

PR #1723, opened 2024-05-27
Ensure license file gets packaged on build

dbanck:
We have a new requirement from Legal to include a copy of the license with all release .zip files and linux packages.

This PR updates the CRT build step similar to
* https://github.com/hashicorp/hc-install/pull/205
* https://github.com/hashicorp/terraform/pull/34977
* https://github.com/hashicorp/nomad/pull/20345


.github/workflows/build.yml
```diff
@@ -137,6 +137,13 @@ jobs:
                 -X 'main.algoliaAppID=${{ secrets.ALGOLIA_APP_ID }}'
                 -X 'main.algoliaAPIKey=${{ secrets.ALGOLIA_API_KEY }}'
               "
+            cp LICENSE "$TARGET_DIR/LICENSE.txt"
+      - name: Copy license file to config_dir # for Linux packages
+        if: ${{ matrix.goos == 'linux' }}
+        env:
+          LICENSE_DIR: ".release/linux/package/usr/share/doc/${{ env.PKG_NAME }}"
+        run: |
+          mkdir -p "$LICENSE_DIR" && cp LICENSE "$LICENSE_DIR/LICENSE.txt"
```
jpogran, line 146: Why only for linux packages?

dbanck: This is an additional step for Linux to follow their conventions for placing a license file in the doc directory. There is also a copy step for the zip archives in line 140.

PR #1745, opened 2024-06-26
Implement support for Terraform Stacks

jpogran:
This commit adds initial support for Terraform Stacks files and Deployment files. This provides completion, hover, and diagnostics for Terraform Stacks and Deployment files.

Needs https://github.com/hashicorp/terraform-schema/pull/346

### DIFF
--- a/internal/features/stacks/decoder/path_reader.go
+++ b/internal/features/stacks/decoder/path_reader.go
@@ -1,0 +1,133 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package decoder
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/hcl-lang/decoder"
+	"github.com/hashicorp/hcl-lang/lang"
+	"github.com/hashicorp/hcl-lang/reference"
+	"github.com/hashicorp/hcl/v2"
+	"github.com/hashicorp/terraform-ls/internal/features/stacks/ast"
+	"github.com/hashicorp/terraform-ls/internal/features/stacks/state"
+	ilsp "github.com/hashicorp/terraform-ls/internal/lsp"
+	stackschema "github.com/hashicorp/terraform-schema/schema"
+)
+
+type PathReader struct {
+	StateReader StateReader
+}
+
+type StateReader interface {
+	List() ([]*state.StackRecord, error)
+	StackRecordByPath(modPath string) (*state.StackRecord, error)
+}
+
+// PathContext returns a PathContext for the given path based on the language ID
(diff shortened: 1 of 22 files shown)


internal/features/stacks/ast/deploy.go
```diff
@@ -0,0 +1,75 @@
...
+
+func (dd DeployDiags) Copy() DeployDiags {
+	m := make(DeployDiags, len(dd))
+	for name, diags := range dd {
+		m[name] = diags
+	}
+	return m
+}
+
+func (dd DeployDiags) AutoloadedOnly() DeployDiags {
```
ansgarm, line 51: What does Autoloaded mean? I'm sure this concept existed before, but I could really use a half sentence comment here that explains what autoloaded means

internal/features/stacks/decoder/path_reader.go
```diff
@@ -0,0 +1,96 @@
...
+// PathContext returns a Stacks PathContext for the given path
+func (pr *PathReader) PathContext(path lang.Path) (*decoder.PathContext, error) {
+	record, err := pr.StateReader.StackRecordByPath(path.Path)
+	if err != nil {
+		return nil, err
+	}
+
+	// get terrafom version from statereader and use that to get the schema
+
+	// TODO: This only provides tfstacks schema. There is also tfdeploy schema
```
ansgarm, line 37: I think this comment is outdated

internal/features/stacks/state/stack_meta.go
```diff
@@ -0,0 +1,25 @@
...
+package state
+
+import (
+	"github.com/hashicorp/go-version"
+)
+
+// StackMetadata contains the result of the early decoding of a module,
+// it will be used obtain the correct provider and related module schemas
+type StackMetadata struct {
+	CoreRequirements version.Constraints
```
ansgarm, line 13: What kind of version constraint is this? Asking because `.terraform-version` is a single specific version and no constraint (I don't know if there's a difference in handling this though)

internal/features/stacks/ast/deploy.go
ansgarm: Added a commit with a half sentence documenting these

internal/features/stacks/state/stack_meta.go
ansgarm: (we removed this here: https://github.com/hashicorp/terraform-ls/pull/1753/commits/b2c17586960f5aaa94e3f863db30701a1907addc)

ansgarm: (this can be resolved if/when we merge the proposed edits)

ansgarm (approved):

PR #1751, opened 2024-07-03
Publish diagnostics for Stack & Deploy files

dbanck:


jpogran (changes requested):
I think we need to add queueRecordChange in `add` and `Remove` as well

ansgarm (approved):

PR #1756, opened 2024-07-08
feat: parse `.terraform-version` files for stacks dirs

ansgarm:
### tasks
- [x] write tests


internal/features/stacks/jobs/version.go
```diff
@@ -0,0 +1,53 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Avoid parsing if it is already in progress or already known
+	if stackRecord.TerraformVersionState != operation.OpStateUnknown && !job.IgnoreState(ctx) {
+		return job.StateNotChangedErr{Dir: document.DirHandleFromPath(stackPath)}
+	}
+
+	stackStore.SetTerraformVersionState(stackPath, operation.OpStateLoading)
```
dbanck, line 34: I think we should handle the error which this function returns

internal/features/stacks/state/stack_record.go
```diff
@@ -23,6 +24,10 @@ type StackRecord struct {
 	ParsingErr       error
 	Diagnostics      ast.SourceDiagnostics
 	DiagnosticsState globalAst.DiagnosticSourceState
+
+	TerraformVersion      *version.Version
```
dbanck, line 28: I think `RequiredTerraformVersion` would be a better fit here, since it does not reflect the actual installed version of Terraform and we don't end up confusing it with the `TerraformVersion` from the root modules.

But feel free to ignore this suggestion as I know you have strong feelings about it.

internal/features/stacks/jobs/version.go
```diff
@@ -0,0 +1,53 @@
...
+
+	// read version file
+	v, err := fs.ReadFile(filepath.Join(stackPath, ".terraform-version"))
+	if err != nil {
+		stackStore.SetTerraformVersionError(stackPath, err)
```
dbanck, line 39: This error as well

internal/features/stacks/jobs/version.go
```diff
@@ -0,0 +1,53 @@
...
+		return err
+	}
+
+	// parse version
+	version, err := version.NewVersion(strings.TrimSpace(string(v)))
+	if err != nil {
+		stackStore.SetTerraformVersionError(stackPath, err)
```
dbanck, line 46: This error as well

internal/features/stacks/jobs/version.go
```diff
@@ -0,0 +1,53 @@
...
+		return err
+	}
+
+	stackStore.SetTerraformVersion(stackPath, version)
```
dbanck, line 50: This error as well

internal/features/stacks/state/stack_store.go
```diff
@@ -253,6 +309,26 @@ func stackCopyByPath(txn *memdb.Txn, path string) (*StackRecord, error) {
...
+	// record removed
+	case oldRecord != nil && newRecord == nil:
+		changes.IsRemoval = true
+
+		if oldRecord.TerraformVersion != nil {
+			changes.TerraformVersion = true
+		}
+	// record changed
+	default:
+		if oldRecord.TerraformVersion == nil || !oldRecord.TerraformVersion.Equal(newRecord.TerraformVersion) {
```
dbanck, line 327:
```suggestion
		if !oldRecord.TerraformVersion.Equal(newRecord.TerraformVersion) {
```

We don't need the `nil` check here, because `.Equal` already does it.

internal/terraform/module/operation/operation.go
```diff
@@ -36,4 +36,5 @@ const (
 	OpTypeReferenceValidation
 	OpTypeTerraformValidate
 	OpTypeParseStackConfiguration
+	OpTypeLoadTerraformVersion
```
dbanck, line 39: Should we mention `Stack` in the job name here to avoid confusion? We already have `OpTypeGetTerraformVersion` to get the version from the Terraform CLI

dbanck (changes requested):
Nice work! Thanks for adding tests for this feature!

I left some comments for some things I noticed. I'm also happy to discuss them over Zoom if you prefer.

internal/features/stacks/events.go
```diff
@@ -168,6 +168,20 @@ func (f *StacksFeature) decodeStack(ctx context.Context, dir document.DirHandle,
 	}
 	ids = append(ids, parseId)
 
+	tfVersion, err := f.stateStore.JobStore.EnqueueJob(ctx, job.Job{
+		Dir: dir,
+		Func: func(ctx context.Context) error {
+			return jobs.LoadTerraformVersion(ctx, f.fs, f.store, path)
+		},
+		Type:        operation.OpTypeLoadModuleMetadata.String(),
+		DependsOn:   job.IDs{parseId},
```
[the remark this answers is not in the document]
ansgarm: copy paste error 😄

internal/features/stacks/state/stack_store.go
ansgarm: Ah, TIL – I was unsure whether that could crash if `TerraformVersion` is nil 😅

dbanck: yeah, it sometimes feels weird to have methods available on `nil`

dbanck (approved):
LGTM ⭐ Thanks for addressing all my feedback!

PR #1759, opened 2024-07-10
Introduce a workflow that checks for changelog entries

dbanck:
This workflow checks for the adding or modification of files in the `.changie/unreleased` folder.

You can see it working for this PR
- The jobs fails for the first commit: https://github.com/hashicorp/terraform-ls/actions/runs/9876936239/job/27277296043?pr=1759
- After adding a changelog entry, it passes: https://github.com/hashicorp/terraform-ls/actions/runs/9877100325/job/27277866068?pr=1759

### DIFF
--- a/.github/workflows/check-changelog.yml
+++ b/.github/workflows/check-changelog.yml
@@ -1,0 +1,36 @@
+# Checks if a file has been committed under the .changes/unreleased directory
+#
+# Skip PRs labeled with 'dependencies'
+
+on:
+  pull_request:
+    branches:
+      - main
+      - pre-release
+
+name: Check if changelog entry exists
+
+jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/check-changelog.yml
```diff
@@ -0,0 +1,33 @@
+# Checks if a file has been committed under the .changes/unreleased directory
+#
+# Skip PRs labeled with 'dependencies'
+
+on:
+  pull_request:
```
ansgarm, line 6: I think I would restrict this to PRs to `main` and `pre-release`. If I create a PR onto another PR, I wouldn't want the second one to need a changie entry 🤔

ansgarm (approved):

PR #1761, opened 2024-07-11
(TF-18286) Add jobs to load Stack metadata

jpogran:
This commit adds a job to load the metadata for a stack. This metadata is used to decode the some of the configuration early before we have all the information we need to decode the full configuration. This is a first step towards decoding the full configuration of a stack.

### DIFF
--- a/internal/features/stacks/ast/stacks.go
+++ b/internal/features/stacks/ast/stacks.go
@@ -79,6 +79,14 @@ func (sf Files) Copy() Files {
 	return m
 }
 
+func (mf Files) AsMap() map[string]*hcl.File {
+	m := make(map[string]*hcl.File, len(mf))
+	for name, file := range mf {
+		m[name.String()] = file // TODO: is this right?
+	}
+	return m
+}
+
 type Diagnostics map[Filename]hcl.Diagnostics
 
 func (sd Diagnostics) Copy() Diagnostics {
--- a/internal/features/stacks/events.go
+++ b/internal/features/stacks/events.go
@@ -191,7 +191,7 @@ func (f *StacksFeature) decodeStack(ctx context.Context, dir document.DirHandle,
 		Func: func(ctx context.Context) error {
 			return jobs.LoadStackMetadata(ctx, f.store, path)
 		},
-		Type:        operation.OpTypeLoadModuleMetadata.String(),
+		Type:        operation.OpTypeLoadStackMetadata.String(),
 		DependsOn:   job.IDs{parseId},
 		IgnoreState: ignoreState,
 	})
@@ -201,7 +201,6 @@ func (f *StacksFeature) decodeStack(ctx context.Context, dir document.DirHandle,
 	ids = append(ids, metaId)
 
 	// TODO: Implement the following functions where appropriate to stacks
-	// Future: LoadModuleMetadata(ctx, f.Store, path)
 	// Future: decodeDeclaredModuleCalls(ctx, dir, ignoreState)
 	// TODO: PreloadEmbeddedSchema(ctx, f.logger, schemas.FS,
 	// Future: DecodeReferenceTargets(ctx, f.Store, f.rootFeature, path)
--- a/internal/features/stacks/events.go
+++ b/internal/features/stacks/events.go
@@ -186,6 +186,20 @@ func (f *StacksFeature) decodeStack(ctx context.Context, dir document.DirHandle,
 	}
 	ids = append(ids, parseId)
 
+	metaId, err := f.stateStore.JobStore.EnqueueJob(ctx, job.Job{
+		Dir: dir,
+		Func: func(ctx context.Context) error {
+			return jobs.LoadStackMetadata(ctx, f.store, path)
+		},
+		Type:        operation.OpTypeLoadModuleMetadata.String(),
+		DependsOn:   job.IDs{parseId},
+		IgnoreState: ignoreState,
+	})
+	if err != nil {
+		return ids, err
+	}
+	ids = append(ids, metaId)
+
 	// TODO: Implement the following functions where appropriate to stacks
 	// Future: LoadModuleMetadata(ctx, f.Store, path)
 	// Future: decodeDeclaredModuleCalls(ctx, dir, ignoreState)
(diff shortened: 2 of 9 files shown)


internal/features/stacks/events.go
```diff
@@ -186,6 +186,20 @@ func (f *StacksFeature) decodeStack(ctx context.Context, dir document.DirHandle,
 	}
 	ids = append(ids, parseId)
 
+	metaId, err := f.stateStore.JobStore.EnqueueJob(ctx, job.Job{
+		Dir: dir,
+		Func: func(ctx context.Context) error {
+			return jobs.LoadStackMetadata(ctx, f.store, path)
+		},
+		Type:        operation.OpTypeLoadModuleMetadata.String(),
```
dbanck, line 194: I think we should introduce a new job type, so that these jobs won't get merged with module metadata jobs.

internal/features/stacks/state/stack_meta.go
```diff
@@ -3,10 +3,17 @@
...
+)
+
 // StackMetadata contains the result of the early decoding of a module,
 // it will be used obtain the correct provider and related module schemas
 type StackMetadata struct {
-	Filenames []string
+	Filenames  []string
+	Components map[string]tfstack.Component
+	Variables  map[string]tfstack.Variable
+	Outputs    map[string]tfstack.Output
```
dbanck, line 16: I believe we're missing `ProviderRequirements` here

internal/features/stacks/state/stack_meta.go
```diff
@@ -3,10 +3,17 @@
...
 }
 
 func (sm StackMetadata) Copy() StackMetadata {
```
dbanck, line 19: We need to copy all fields of the struct in the `Copy()` method

internal/features/stacks/state/stack_store.go
```diff
@@ -258,6 +259,60 @@ func (s *StackStore) SetTerraformVersionState(path string, state operation.OpSta
...
+	if err != nil {
+		return err
+	}
+
+	mod := oldMod.Copy()
+	mod.Meta = StackMetadata{
+		Components: meta.Components,
+		Variables:  meta.Variables,
+		Outputs:    meta.Outputs,
+		Filenames:  meta.Filenames,
```
dbanck, line 298: I believe we're missing `ProviderRequirements` here

internal/features/stacks/ast/stacks.go
```diff
@@ -79,6 +79,14 @@ func (sf Files) Copy() Files {
 	return m
 }
 
+func (mf Files) AsMap() map[string]*hcl.File {
+	m := make(map[string]*hcl.File, len(mf))
+	for name, file := range mf {
+		m[name.String()] = file // TODO: is this right?
```
dbanck, line 85: Yep 👍

internal/features/stacks/jobs/metadata.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/hashicorp/terraform-ls/internal/document"
+	"github.com/hashicorp/terraform-ls/internal/features/stacks/state"
+	"github.com/hashicorp/terraform-ls/internal/job"
+	"github.com/hashicorp/terraform-ls/internal/terraform/module/operation"
+	earlydecoder "github.com/hashicorp/terraform-schema/earlydecoder/stacks"
+)
+
+// LoadStackMetadata loads data about the stack in a version-independent
+// way that enables us to decode the rest of the configuration,
+// e.g. by knowing provider versions, Terraform Core constraint etc.
```
dbanck, line 18:
```suggestion
// e.g. by knowing provider versions, etc.
```

internal/features/stacks/jobs/metadata.go
```diff
@@ -0,0 +1,63 @@
...
+	// }
+	// meta.ProviderRequirements = providerRequirements
+
+	// providerRefs := make(map[tfmodule.ProviderRef]tfaddr.Provider, len(meta.ProviderReferences))
+	// for localRef, pAddr := range meta.ProviderReferences {
+	// 	// TODO: check pAddr for migrations via Registry API?
+	// 	providerRefs[localRef] = pAddr
+	// }
+	// meta.ProviderReferences = providerRefs
+
```
dbanck, line 56:
```suggestion
```

At this point, I don't know why we need to create a copy of the provider requirements for module metadata. Let's not do it for stack metadata for now and see if we run into any problems.

dbanck (changes requested):
Nice work!

I left some small suggestions

internal/features/stacks/state/stack_meta.go
```diff
@@ -3,16 +3,27 @@
...
 	newSm := StackMetadata{
-		// version.Constraints is practically immutable once parsed
-		Filenames: sm.Filenames,
+		Filenames:            sm.Filenames,
+		Components:           sm.Components,
+		Variables:            sm.Variables,
+		Outputs:              sm.Outputs,
+		ProviderRequirements: sm.ProviderRequirements,
```
dbanck, line 26:
```suggestion
		Filenames: sm.Filenames,
	}

	if sm.Components != nil {
		newSm.Components = make(map[string]tfstack.Component, len(sm.Components))
		for k, v := range sm.Components {
			newSm.Components[k] = v
		}
	}

	if sm.Variables != nil {
		newSm.Variables = make(map[string]tfstack.Variable, len(sm.Variables))
		for k, v := range sm.Variables {
			newSm.Variables[k] = v
		}
	}

	if sm.Outputs != nil {
		newSm.Outputs = make(map[string]tfstack.Output, len(sm.Outputs))
		for k, v := range sm.Outputs {
			newSm.Outputs[k] = v
		}
	}

	if sm.ProviderRequirements != nil {
		newSm.ProviderRequirements = make(map[string]tfstack.ProviderRequirement, len(sm.ProviderRequirements))
		for k, v := range sm.ProviderRequirements {
			newSm.ProviderRequirements[k] = v
		}
```

Since most of the fields in the struct are maps, we need to copy their entries individually. I've taken this code from the embedded schema branch

dbanck (approved):

PR #1763, opened 2024-07-12
feat: load embedded provider schemas for providers found in stacks files

ansgarm:
Depends on https://github.com/hashicorp/terraform-schema/pull/375

---

This PR adds a job to load provider schemas that are embedded into the language server binary into state after the early decoder parsed which providers where used in a given stacks project. We will be going to use this information in the schema merger to e.g. enhance provider blocks with their respective attributes within the `config` block to support hover and completions.

### DIFF
--- a/internal/features/stacks/events.go
+++ b/internal/features/stacks/events.go
@@ -208,7 +208,7 @@ func (f *StacksFeature) decodeStack(ctx context.Context, dir document.DirHandle,
 				f.store, f.stateStore.ProviderSchemas, path)
 		},
 		DependsOn:   job.IDs{metaId},
-		Type:        operation.OpTypePreloadEmbeddedSchema.String(),
+		Type:        operation.OpTypeStacksPreloadEmbeddedSchema.String(),
 		IgnoreState: ignoreState,
 	})
 	if err != nil {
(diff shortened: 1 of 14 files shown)


internal/features/stacks/state/stack_record.go
```diff
@@ -42,10 +46,17 @@ func (m *StackRecord) Copy() *StackRecord {
...
-		path:             m.path,
+		path: m.path,
+
+		PreloadEmbeddedSchemaState: m.PreloadEmbeddedSchemaState,
+
 		Meta:             m.Meta.Copy(),
 		ParsingErr:       m.ParsingErr,
 		DiagnosticsState: m.DiagnosticsState.Copy(),
+
+		RequiredTerraformVersion:      m.RequiredTerraformVersion,
```
dbanck, line 57: Good catch!

internal/features/stacks/events.go
```diff
@@ -200,9 +201,23 @@ func (f *StacksFeature) decodeStack(ctx context.Context, dir document.DirHandle,
...
 	ids = append(ids, metaId)
 
+	eSchemaId, err := f.stateStore.JobStore.EnqueueJob(ctx, job.Job{
+		Dir: dir,
+		Func: func(ctx context.Context) error {
+			return jobs.PreloadEmbeddedSchema(ctx, f.logger, schemas.FS,
+				f.store, f.stateStore.ProviderSchemas, path)
+		},
+		DependsOn:   job.IDs{metaId},
+		Type:        operation.OpTypePreloadEmbeddedSchema.String(),
```
dbanck, line 211: I think we should introduce a new job type, so that these jobs won't get merged with module preload schema jobs.

dbanck (commented):
Nice work! I think it's only missing a separate job type.

PR #1768, opened 2024-07-16
(TF-18292) Load Stack component sources from metadata

jpogran:
This change adds a new job to load the component sources of a stack after parsing the stack metadata. For a component with a module source, the job will trigger a `did_open` event for the Modules feature to handle. This will allow the Modules feature to load the module and its dependencies, leading to eventually being able to complete Module references.

This only handles local modules at the moment. Remote modules are not supported yet.

Needs https://github.com/hashicorp/terraform-schema/pull/374 to parse module source addresses.


dbanck (approved):
Nice work! 👍

PR #1770, opened 2024-07-18
feat: use stacks schema merger

ansgarm:
Merge all the schemas (for stacks stack files) 💯

This brings support for completions for `inputs` and `providers` in `component` blocks and for showing descriptions on hover among other schema related features. Note: this currently only supports local components with a source in a relative directory.

### DIFF
--- a/internal/features/stacks/decoder/path_reader.go
+++ b/internal/features/stacks/decoder/path_reader.go
@@ -16,19 +16,27 @@ import (
 	"github.com/hashicorp/terraform-ls/internal/features/stacks/state"
 	ilsp "github.com/hashicorp/terraform-ls/internal/lsp"
 	tfaddr "github.com/hashicorp/terraform-registry-address"
+	tfmod "github.com/hashicorp/terraform-schema/module"
 	tfschema "github.com/hashicorp/terraform-schema/schema"
 	stackschema "github.com/hashicorp/terraform-schema/schema/stacks"
 	tfstack "github.com/hashicorp/terraform-schema/stack"
 )
 
 type PathReader struct {
-	StateReader StateReader
+	StateReader  StateReader
+	ModuleReader ModuleReader
 }
 
 type StateReader interface {
 	List() ([]*state.StackRecord, error)
 	StackRecordByPath(modPath string) (*state.StackRecord, error)
 	ProviderSchema(modPath string, addr tfaddr.Provider, vc version.Constraints) (*tfschema.ProviderSchema, error)
+}
+
+type ModuleReader interface {
+	// LocalModuleMeta returns the module meta data for a local module. This is the result
+	// of the [earlydecoder] when processing module files
+	LocalModuleMeta(modPath string) (*tfmod.Meta, error)
 }
 
 // PathContext returns a PathContext for the given path based on the language ID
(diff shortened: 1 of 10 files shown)


internal/features/stacks/state/stack_store.go
```diff
@@ -306,6 +308,11 @@ func (s *StackStore) UpdateMetadata(path string, meta *tfstack.Meta, mErr error)
 		return err
 	}
 
+	err = txn.Insert(s.tableName, record)
+	if err != nil {
+		return err
+	}
+
```
dbanck, line 315:
```suggestion
```

It was so nice, we inserted it twice

internal/features/stacks/events.go
```diff
@@ -205,34 +205,38 @@ func (f *StacksFeature) decodeStack(ctx context.Context, dir document.DirHandle,
 			}
 
 			spawnedIds, err := loadStackComponentSources(ctx, f.store, f.bus, path)
+			if err != nil {
+				return spawnedIds, err
```
dbanck, line 209: Do we need to return here? I think we can still try to load the schemas, even if decoding the components fails

internal/features/stacks/events.go
```diff
@@ -205,34 +205,38 @@ func (f *StacksFeature) decodeStack(ctx context.Context, dir document.DirHandle,
...
+			}
+			spawnedIds = append(spawnedIds, eSchemaId)
 
-	eSchemaId, err := f.stateStore.JobStore.EnqueueJob(ctx, job.Job{
-		Dir: dir,
-		Func: func(ctx context.Context) error {
-			return jobs.PreloadEmbeddedSchema(ctx, f.logger, schemas.FS,
-				f.store, f.stateStore.ProviderSchemas, path)
+			return spawnedIds, nil
```
dbanck, line 233: What do you think about adding a `deferIds := make(job.IDs, 0)` to the top of this function, appending to it, and returning it here? I think that would make it more readable.

internal/features/stacks/decoder/path_reader.go
```diff
@@ -35,28 +48,46 @@ func (pr *PathReader) PathContext(path lang.Path) (*decoder.PathContext, error)
...
 	}
 
 	schema, err := stackschema.CoreStackSchemaForVersion(version)
 	if err != nil {
 		return nil, err
 	}
 
+	sm := stackschema.NewStackSchemaMerger(schema)
+	sm.SetStateReader(stateReader)
+	sm.SetModuleReader(moduleReader)
```
dbanck, line 73: In line with the [comment on terraform-schema](https://github.com/hashicorp/terraform-schema/pull/376#discussion_r1695506664), what do you think about having only one reader in the schema merger?

dbanck (commented):
Nice work! I left some suggestions

internal/features/stacks/events.go
ansgarm: Would you then just ignore the error or keep it for later? Which error would you return if loading the schemas returns one as well?

dbanck: I would log the error from `loadStackComponentSources` and proceed with loading the schemas.

If scheduling the schema job fails, something must really be broken, so I think returning there makes sense.

dbanck (approved):
Thanks for addressing the feedback

PR #1776, opened 2024-07-24
(TF-18290) Early validation for Terraform Stack files

jpogran:
This change enables early validation for Terraform Stacks in tfstack. hcl and tfdeploy.hcl files.

Available validations:

- Missing required attributes
- Deprecated attributes
- Deprecated blocks
- Unexpected attributes
- Unexpected blocks
- Maximum number of blocks
- Minimum number of blocks
- Block labels length

This adds a new job to validate the schema of the these files. The job is enqueued when the `EnableEnhancedValidation` option is set to true.

### DIFF
--- a/internal/features/stacks/decoder/path_reader.go
+++ b/internal/features/stacks/decoder/path_reader.go
@@ -102,6 +102,7 @@ func stackPathContext(record *state.StackRecord, stateReader CombinedReader) (*d
 		ReferenceTargets: make(reference.Targets, 0),
 		Files:            make(map[string]*hcl.File, 0),
 		Functions:        mustFunctionsForVersion(version),
+		Validators:       stackValidators,
 	}
 
 	// TODO: Add reference origins and targets if needed
@@ -148,6 +149,7 @@ func deployPathContext(record *state.StackRecord) (*decoder.PathContext, error) 
 		ReferenceOrigins: make(reference.Origins, 0),
 		ReferenceTargets: make(reference.Targets, 0),
 		Files:            make(map[string]*hcl.File, 0),
+		Validators:       stackValidators,
 	}
 
 	// TODO: Add reference origins and targets if needed
--- a/internal/features/stacks/decoder/validations/missing_required_attribute.go
+++ b/internal/features/stacks/decoder/validations/missing_required_attribute.go
@@ -1,0 +1,72 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package validations
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/hcl-lang/schema"
+	"github.com/hashicorp/hcl-lang/schemacontext"
+	"github.com/hashicorp/hcl/v2"
+	"github.com/hashicorp/hcl/v2/hclsyntax"
+)
+
+type MissingRequiredAttribute struct{}
+
+func (mra MissingRequiredAttribute) Visit(ctx context.Context, node hclsyntax.Node, nodeSchema schema.Schema) (context.Context, hcl.Diagnostics) {
+	var diags hcl.Diagnostics
+	if HasUnknownRequiredAttributes(ctx) {
+		return ctx, diags
+	}
+
+	switch nodeType := node.(type) {
+	case *hclsyntax.Block:
+		// Providers were excluded from Terraform validation due to https://github.com/hashicorp/vscode-terraform/issues/1616
+		// TODO: See if we can remove this exclusion for Terraform Stacks
+		// We would need to check if the provider can be configured via environment variables in Stacks. Since you can
+		// have multiple configurations for the same provider, this may be challenging.
+		// If it is possible, we may need to update this code to reflect the nested config { } block. But we're not sure right now.
+		nestingLvl, nestingOk := schemacontext.BlockNestingLevel(ctx)
+		if nodeType.Type == "provider" && (nestingOk && nestingLvl == 0) {
+			ctx = WithUnknownRequiredAttributes(ctx)
+		}
+	case *hclsyntax.Body:
+		if nodeSchema == nil {
+			return ctx, diags
+		}
+
+		bodySchema := nodeSchema.(*schema.BodySchema)
+		if bodySchema.Attributes == nil {
+			return ctx, diags
+		}
+
+		for name, attr := range bodySchema.Attributes {
+			if attr.IsRequired {
+				_, ok := nodeType.Attributes[name]
+				if !ok {
+					diags = append(diags, &hcl.Diagnostic{
+						Severity: hcl.DiagError,
+						Summary:  fmt.Sprintf("Required attribute %q not specified", name),
+						Detail:   fmt.Sprintf("An attribute named %q is required here", name),
+						Subject:  nodeType.SrcRange.Ptr(),
+					})
+				}
+			}
+		}
+	}
+
+	return ctx, diags
+}
+
+type unknownRequiredAttrsCtxKey struct{}
+
+func HasUnknownRequiredAttributes(ctx context.Context) bool {
+	_, ok := ctx.Value(unknownRequiredAttrsCtxKey{}).(bool)
+	return ok
+}
+
+func WithUnknownRequiredAttributes(ctx context.Context) context.Context {
+	return context.WithValue(ctx, unknownRequiredAttrsCtxKey{}, true)
+}
--- a/internal/features/stacks/decoder/validations/valid_name.go
+++ b/internal/features/stacks/decoder/validations/valid_name.go
@@ -1,0 +1,71 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package validations
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/hcl-lang/schema"
+	"github.com/hashicorp/hcl/v2"
+	"github.com/hashicorp/hcl/v2/hclsyntax"
+)
+
+type StackBlockValidName struct{}
+
+func (sb StackBlockValidName) Visit(ctx context.Context, node hclsyntax.Node, nodeSchema schema.Schema) (context.Context, hcl.Diagnostics) {
+	var diags hcl.Diagnostics
+
+	block, ok := node.(*hclsyntax.Block)
+	if !ok {
+		return ctx, diags
+	}
+
+	if nodeSchema == nil {
+		return ctx, diags
+	}
+
+	// TODO: Revist checking Stack concepts in the static validators
+	// 	Other validators are more generic and can be applied to any block/attribute/etc, but
+	// 	this one is specific to the stack block types. I'm sure this could be done in a more generic way,
+	// 	but I'm not sure if it's worth it at the moment. The names of these blocks are used as identifiers
+	//  in the stack and are used in the UI, so they have to be valid Terraform identifiers or else the stack
+	//  will not be able to be created. This is a very basic check to ensure that the names are valid until we can
+	//  do more advanced checks or use the cloud api.
+	//  I tried adding this to the earlydecoder, but we do not seem to report the diagnostics there at all currently,
+	//  so nothing was being reported. I'm not sure if that's a bug or intended.
+
+	switch block.Type {
+	// stack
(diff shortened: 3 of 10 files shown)


internal/features/stacks/events.go
```diff
@@ -234,6 +235,26 @@ func (f *StacksFeature) decodeStack(ctx context.Context, dir document.DirHandle,
...
+		return ids, err
+	}
+	if validationOptions.EnableEnhancedValidation {
+		_, err = f.stateStore.JobStore.EnqueueJob(ctx, job.Job{
+			Dir: dir,
+			Func: func(ctx context.Context) error {
+				return jobs.SchemaModuleValidation(ctx, f.store, dir.Path())
+			},
+			Type:        operation.OpTypeSchemaModuleValidation.String(),
+			DependsOn:   ids,
```
ansgarm, line 249: I think it's fine to have the job at this level currently, but I think we'll need to move it into defer of LoadStackMetadata soon, don't we? 🤔
(Mainly asking so that @dbanck can give some background 😄)

ansgarm (approved):
Just two small questions – but it's looking good for all that I understand!

internal/features/stacks/events.go
dbanck: Once we start adding inputs for components and they become available for validation, we need to depend on the `spawnedIds`. This ensures that the component metadata is fully loaded, before we run the validation.

And since this job depends on `PreloadEmbeddedSchema`, which also needs to move to `Defer`, this job needs to follow. _I think we moved it a few of days ago in another branch ?_

ansgarm: Cool, thanks!

> I think we moved it a few of days ago in another branch ?

Yes, I did in https://github.com/hashicorp/terraform-ls/pull/1770

jpogran: Apologies, in my haste I habitually hit ready for review, but I really did intend on waiting for all of Ansgar's work to land before this was included so it could be properly ordered. Just waiting that

internal/features/stacks/events.go
```diff
@@ -191,6 +192,9 @@ func (f *StacksFeature) decodeStack(ctx context.Context, dir document.DirHandle,
 	}
 	ids = append(ids, parseId)
 
+	validationOptions, _ := lsctx.ValidationOptions(ctx)
+	f.logger.Printf("validationOptions: %v", validationOptions)
```
dbanck, line 196:
```suggestion
```

internal/features/stacks/events.go
```diff
@@ -233,6 +237,23 @@ func (f *StacksFeature) decodeStack(ctx context.Context, dir document.DirHandle,
 			}
 			deferIds = append(deferIds, eSchemaId)
 
+			f.logger.Printf("validationOptions: %v", validationOptions)
```
dbanck, line 240:
```suggestion
```

internal/features/stacks/events.go
```diff
@@ -233,6 +237,23 @@ func (f *StacksFeature) decodeStack(ctx context.Context, dir document.DirHandle,
...
+			if validationOptions.EnableEnhancedValidation {
+				validationIds, err := f.stateStore.JobStore.EnqueueJob(ctx, job.Job{
+					Dir: dir,
+					Func: func(ctx context.Context) error {
+						return jobs.SchemaStackValidation(ctx, f.store, f.moduleFeature, dir.Path())
+					},
+					Type:        operation.OpTypeSchemaModuleValidation.String(),
```
dbanck, line 247: We should use a new job type to avoid conflicts

internal/features/stacks/jobs/validation.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/hashicorp/terraform-ls/internal/job"
+	lsp "github.com/hashicorp/terraform-ls/internal/lsp"
+	globalAst "github.com/hashicorp/terraform-ls/internal/terraform/ast"
+	"github.com/hashicorp/terraform-ls/internal/terraform/module/operation"
+)
+
+func SchemaStackValidation(ctx context.Context, stackStore *state.StackStore, moduleFeature stackDecoder.ModuleReader, stackPath string) error {
+	rpcContext := lsctx.DocumentContext(ctx)
+	isMatchingLanguageId := rpcContext.LanguageID == lsp.Stacks.String() || rpcContext.LanguageID == lsp.Deploy.String()
+	if !isMatchingLanguageId {
```
dbanck, line 28: Since this job is scheduled during the decoding of a stack, how can the language ID not match here?

internal/features/stacks/jobs/validation.go
```diff
@@ -0,0 +1,90 @@
...
+
+	d := decoder.NewDecoder(&stackDecoder.PathReader{
+		StateReader:  stackStore,
+		ModuleReader: moduleFeature,
+	})
+	d.SetContext(idecoder.DecoderContext(ctx))
+
+	decoder, err := d.Path(lang.Path{
+		Path:       stackPath,
+		LanguageID: rpcContext.LanguageID,
```
dbanck, line 55: I think we should not create the path context based on the language ID from rpcContext, but rather use a static value.

If we use the ID from the context, we would only be validating the current file and not the whole Stack. When you open a folder, you expect the whole Stack to be validated, so you don't have to open deployment files separately to see their diagnostics.

dbanck: What do you think about creating two separate path contexts and the combining their diagnostics? Something along the lines of

```go
d := decoder.NewDecoder(&fdecoder.PathReader{
	StateReader:  modStore,
	ModuleReader: moduleFeature,
})
d.SetContext(idecoder.DecoderContext(ctx))

diags := make(lang.DiagnosticsMap)

stackPathCtx, err := d.Path(lang.Path{
	Path:       modPath,
	LanguageID: ilsp.Stacks.String(),
})
if err != nil {
	return err
}
stacksDiags, err := stackPathCtx.Validate(ctx)
if err != nil {
	return err
}
diags = diags.Extend(stacksDiags)

deployPathCtx, err := d.Path(lang.Path{
	Path:       modPath,
	LanguageID: ilsp.Deploy.String(),
})
if err != nil {
	return err
}
deployDiags, err := deployPathCtx.Validate(ctx)
if err != nil {
	return err
}
diags = diags.Extend(deployDiags)
```

internal/features/stacks/decoder/validations/missing_required_attribute.go
```diff
@@ -0,0 +1,69 @@
...
+func (mra MissingRequiredAttribute) Visit(ctx context.Context, node hclsyntax.Node, nodeSchema schema.Schema) (context.Context, hcl.Diagnostics) {
+	var diags hcl.Diagnostics
+	if HasUnknownRequiredAttributes(ctx) {
+		return ctx, diags
+	}
+
+	switch nodeType := node.(type) {
+	case *hclsyntax.Block:
+		// Providers were excluded from Terraform validation due to https://github.com/hashicorp/vscode-terraform/issues/1616
+		// TODO: See if we can remove this exclusion for Terraform Stacks
```
dbanck, line 27: We would need to check if the provider can be configured via environment variables in Stacks. Since you can have multiple configurations for the same provider, this may be challenging.

If it is possible, we may need to update this code to reflect the nested `config {` block. But I'm not sure.

dbanck (commented):
Nice draft! I left some thoughts

internal/features/stacks/decoder/validations/missing_required_attribute.go
jpogran: Leaving this as a future TODO

internal/features/stacks/jobs/validation.go
jpogran: I initially had it this way, then in an attempt to reduce duplication consolidated to one. I can see your point about the other types of change notifications (open, etc) requiring a validation of the entire stack and deploy files which would require having both decoders.

In the situation of a didChange event however, we know we're dealing with one file and do not have to walk everything all over again. So I kept the single decoder here.

internal/features/stacks/decoder/path_reader.go
```diff
@@ -55,10 +55,11 @@ func (pr *PathReader) PathContext(path lang.Path) (*decoder.PathContext, error)
 
 	switch path.LanguageID {
 	case ilsp.Stacks.String():
-		return stackPathContext(record, CombinedReader{
+		x := CombinedReader{
 			StateReader:  pr.StateReader,
 			ModuleReader: pr.ModuleReader,
-		})
+		}
+		return stackPathContext(record, x)
```
ansgarm, line 62: I think we could revert this and use the shorter version 😇

internal/features/stacks/decoder/validations/missing_required_attribute.go
```diff
@@ -0,0 +1,72 @@
...
+			}
+		}
+	}
+
+	return ctx, diags
+}
+
+type unknownRequiredAttrsCtxKey struct{}
+
+func HasUnknownRequiredAttributes(ctx context.Context) bool {
```
ansgarm, line 65: Just out of curiosity (for my Golang understanding; no need to change): Do these need to be public?

internal/features/stacks/decoder/validations/valid_name.go
```diff
@@ -0,0 +1,73 @@
...
+	switch block.Type {
+	// stack
+	case "component":
+		diags = hasValidNameLabel(block, diags)
+	// case "provider":
+	// case "required_providers":
+	case "variable":
+		diags = hasValidNameLabel(block, diags)
+	case "input":
+		diags = hasValidNameLabel(block, diags)
```
ansgarm, line 48: Does this block actually exist?

internal/features/stacks/events.go
```diff
@@ -233,6 +238,22 @@ func (f *StacksFeature) decodeStack(ctx context.Context, dir document.DirHandle,
 			}
 			deferIds = append(deferIds, eSchemaId)
 
+			if validationOptions.EnableEnhancedValidation {
+				validationIds, err := f.stateStore.JobStore.EnqueueJob(ctx, job.Job{
```
ansgarm, line 242:
```suggestion
				validationId, err := f.stateStore.JobStore.EnqueueJob(ctx, job.Job{
```

internal/features/stacks/events.go
```diff
@@ -233,6 +238,22 @@ func (f *StacksFeature) decodeStack(ctx context.Context, dir document.DirHandle,
...
+						return jobs.SchemaStackValidation(ctx, f.store, f.moduleFeature, dir.Path())
+					},
+					Type:        operation.OpTypeSchemaStackValidation.String(),
+					DependsOn:   deferIds,
+					IgnoreState: ignoreState,
+				})
+				if err != nil {
+					return deferIds, err
+				}
+				deferIds = append(deferIds, validationIds)
```
ansgarm, line 254:
```suggestion
				deferIds = append(deferIds, validationId)
```

internal/features/stacks/jobs/validation.go
```diff
@@ -0,0 +1,118 @@
...
+			return err
+		}
+
+		diags := make(lang.DiagnosticsMap)
+
+		stacksDiags, err := stackDecoder.Validate(ctx)
+		if err != nil {
+			// TODO: Should we really return here or continue with the other decoders?
+			// Is this really a complete fail case? Shouldn't a failure in a stack file
+			// not prevent the deploy file from being validated?
```
ansgarm, line 100: I think it's fine to return the error here as we don't return diags in any error case anyway currently – Or am I missing something?

ansgarm:
```suggestion
```

ansgarm (commented):
This looks really good! Just a few small cleanup things mostly.

I'm fine with having just one decoder for `didChange` events 👍 We might need to change that once we support references between stack and deployment files (for inputs) but that's for later!

internal/features/stacks/jobs/validation.go
jpogran: Going to create a followup ticket for this. Leaving the comment for when we come back around.

internal/features/stacks/decoder/validations/missing_required_attribute.go
jpogran: This was copied from Modules, and I think it used the existing pattern from icl-lang

PR #1786, opened 2024-08-05
(TF-18664) Add DecodeReferenceOrigins and DecodeReferenceTargets jobs

jpogran:
This commit adds two new jobs, DecodeReferenceOrigins and DecodeReferenceTargets, and their supporting plumbing to the stacks feature. These jobs are responsible for collecting reference origins and targets, respectively. Reference origins and targets are used to determine where a reference is defined and where it is used.

This information is useful for features like go-to-definition and go-to-references.

### DIFF
--- a/internal/features/stacks/decoder/path_reader.go
+++ b/internal/features/stacks/decoder/path_reader.go
@@ -106,6 +106,19 @@ func stackPathContext(record *state.StackRecord, stateReader CombinedReader) (*d
 	}
 
 	// TODO: Add reference origins and targets if needed
+	for _, origin := range record.RefOrigins {
+		if ast.IsStackFilename(origin.OriginRange().Filename) {
+			pathCtx.ReferenceOrigins = append(pathCtx.ReferenceOrigins, origin)
+		}
+	}
+
+	for _, target := range record.RefTargets {
+		if target.RangePtr != nil && ast.IsStackFilename(target.RangePtr.Filename) {
+			pathCtx.ReferenceTargets = append(pathCtx.ReferenceTargets, target)
+		} else if target.RangePtr == nil {
+			pathCtx.ReferenceTargets = append(pathCtx.ReferenceTargets, target)
+		}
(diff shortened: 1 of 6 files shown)


internal/features/stacks/decoder/path_reader.go
```diff
@@ -153,6 +166,19 @@ func deployPathContext(record *state.StackRecord) (*decoder.PathContext, error)
...
+			pathCtx.ReferenceOrigins = append(pathCtx.ReferenceOrigins, origin)
+		}
+	}
+
+	for _, target := range record.RefTargets {
+		if target.RangePtr != nil && ast.IsDeployFilename(target.RangePtr.Filename) {
+			pathCtx.ReferenceTargets = append(pathCtx.ReferenceTargets, target)
+		} else if target.RangePtr == nil {
+			pathCtx.ReferenceTargets = append(pathCtx.ReferenceTargets, target)
+		}
```
ansgarm, line 180: Asking, because I don't know much about references yet: What are these two cases handling? Maybe you can explain this to me in sync :)

internal/features/stacks/events.go
```diff
@@ -262,6 +270,25 @@ func (f *StacksFeature) decodeStack(ctx context.Context, dir document.DirHandle,
...
+				return jobs.SchemaStackValidation(ctx, f.store, f.moduleFeature, dir.Path())
+			},
+			Type:        operation.OpTypeSchemaStackValidation.String(),
+			DependsOn:   ids,
+			IgnoreState: ignoreState,
+		})
+		if err != nil {
+			return ids, err
+		}
+	}
```
ansgarm, line 290: I think we should keep this in defer – there was a fix on `main` for the modules feature for exactly this: https://github.com/hashicorp/terraform-ls/pull/1777

ansgarm:
(fyi: ran `go fmt`, amended your last commit and force pushed to fix CI failing)

PR #1797, opened 2024-08-15
(TF-18673) Validate Stack and Deployment files for unreferenced origins

jpogran:
This commit adds a new validation to check for unreferenced origins in stack and deployment files. This validation is performed after the reference targets and origins are decoded.

The validation checks for references to variables and local values that do not have a corresponding target. The validation is performed for variables, local values, providers and identity_tokens only, as components can have unknown schema.


ansgarm (commented):
Nice work! Test drove it and it works for the most parts – the only thing I thought would work but didn't was a reference to a component with a name that doesn't exist:
image
(^ this could and should fail)

jpogran:
> Nice work! Test drove it and it works for the most parts – the only thing I thought would work but didn't was a reference to a component with a name that doesn't exist

Components are not in the list of things that are supported.

ansgarm (approved):

PR #1798, opened 2024-08-16
feat: store early decoded deploy config and bump terraform-schema

ansgarm:
Depends on https://github.com/hashicorp/terraform-schema/pull/390 to be merged and this PR to be updated with the squashed commit hash (_edit: done_)

store_references

### DIFF
--- a/internal/features/stacks/decoder/path_reader.go
+++ b/internal/features/stacks/decoder/path_reader.go
@@ -89,6 +89,8 @@ func stackPathContext(record *state.StackRecord, stateReader CombinedReader) (*d
 		Variables:            record.Meta.Variables,
 		Outputs:              record.Meta.Outputs,
 		Filenames:            record.Meta.Filenames,
+		Deployments:          record.Meta.Deployments,
+		Stores:               record.Meta.Stores,
 	}
 
 	mergedSchema, err := sm.SchemaForStack(meta)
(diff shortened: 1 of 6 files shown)


internal/features/stacks/decoder/path_reader.go
```diff
@@ -150,6 +152,8 @@ func deployPathContext(record *state.StackRecord) (*decoder.PathContext, error)
 		Variables:            record.Meta.Variables,
 		Outputs:              record.Meta.Outputs,
 		Filenames:            record.Meta.Filenames,
+		Deployments:          record.Meta.Deployments,
+		Stores:               record.Meta.Stores,
```
jpogran, line 156: I'm not sure on naming for `store`, as we use `store` for the database a lot in this repo. Do you think it would be better to be more descriptive?

ansgarm: I don't have a strong opinion, but I'd keep it as is because that way it's the closest to the config and because it's usage is fairly limited and we have types that show its difference to the stores we use 🤔
If you prefer something like "StoreBlocks" instead (or have a better idea) I'm happy to rename it 👍

jpogran: What's in a name? That which we call a store
By any other name would smell as sweet;
So Store would, were he not Store call'd,
Retain that dear perfection which he owes
Without that title. Store, doff thy name,
And for that name which is no part of thee

With apologies to the bard, let's keep store then

jpogran (approved):
🚢

PR #1799, opened 2024-08-16
(TF-19244) feat: support a subset of functions in deployment configs

ansgarm:
This adds support for a subset of the functions available in Terraform for deployment configuration.
deploy_functions

For example, `sum` is not supported currently:
image


jpogran (approved):
:shipit:

PR #1813, opened 2024-08-29
(TFECO-7339) feat: Support context references within orchestrate blocks in deployment configuration

ansgarm:
depends on https://github.com/hashicorp/terraform-schema/pull/398

context

### DIFF
--- a/internal/features/stacks/jobs/builtin_references.go
+++ b/internal/features/stacks/jobs/builtin_references.go
@@ -37,26 +37,16 @@ func builtinReferences(record *state.StackRecord) reference.Targets {
 
 	// The ranges of the orchestrate blocks as we have to create targets with these ranges
 	// to ensure they are only available within orchestrate blocks
-	ranges := make([]hcl.Range, 0)
 	for _, rule := range record.Meta.OrchestrationRules {
-		ranges = append(ranges, rule.Range)
-	}
+		rng := rule.Range
 
-	// The names of the existing components in the stack
-	// We use this to offer completions for the component_changes map
-	componentNames := make([]string, 0)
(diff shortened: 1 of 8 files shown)


internal/features/stacks/jobs/builtin_references.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	// The names of the existing components in the stack
+	// We use this to offer completions for the component_changes map
+	componentNames := make([]string, 0)
+	for name := range record.Meta.Components {
+		componentNames = append(componentNames, name)
+	}
+
+	for _, rng := range ranges {
```
jpogran, line 52: Do we need to store the ranges again when they're already in the shape we need here?

internal/features/stacks/jobs/builtin_references.go
```diff
@@ -0,0 +1,223 @@
...
+		// create the static base targets (like context.operation, context.success, etc.)
+		targets = append(targets, baseTargets(rng)...)
+		// create the static plan targets (like context.plan.mode, context.plan.applyable, etc.)
+		targets = append(targets, staticPlanTargets(rng)...)
+
+		// targets for each component for the component_changes map (like context.plan.component_changes["vpc"].total)
+		for _, name := range componentNames {
```
jpogran, line 59: Same question as above, not sure they need to be new allocations

internal/features/stacks/jobs/builtin_references.go
ansgarm: Yeah, makes sense! I was unsure whether I would need this multiple time but ended up with just two loops over them 👍

jpogran (approved):
:shipit:

PR #1815, opened 2024-09-01
fix brew install command

mo18:
the current Installation document has an outdated command for macos installation.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -16,7 +16,7 @@ You can install the language server manually using one of the many package manag
 You can install via [Homebrew](https://brew.sh)
 
 ```
-brew install hashicorp/tap/terraform-ls
+brew install terraform-ls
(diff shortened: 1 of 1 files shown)


jpogran (changes requested):
Hi @mo18, thanks for the contribution and I appreciate the intent but this is not correct.

We do publish terraform-ls to the HashiCorp [homebrew tap](https://github.com/hashicorp/homebrew-tap/blob/master/Formula/terraform-ls.rb) and that is the source we recommend to use. These are official builds built and signed by HashiCorp, you can see more information here: https://github.com/hashicorp/homebrew-tap?tab=readme-ov-file#why-should-i-install-packages-from-this-tap.

mo18:
hi @jpogran, I appreciate the response and apologies for the confusion

PR #1828, opened 2024-09-13
(TFECO-7524) chore: add Stacks Terraform version to telemetry

ansgarm:
A minimal addition to the existing telemetry data that lets us measure usage of the new stacks feature.


jpogran (approved):
not sure this is the best shape for this, but this is good until we figure that out

PR #1850, opened 2024-10-17
(TFECO-7637) Raise HCL diagnostics

jpogran:
This change updates the stack metadata loading to merge the new diagnostics with the existing ones. This is necessary to ensure that we don't lose any diagnostics that were previously reported.

Needs https://github.com/hashicorp/terraform-schema/pull/412

### DIFF
--- a/internal/features/stacks/jobs/metadata.go
+++ b/internal/features/stacks/jobs/metadata.go
@@ -7,8 +7,10 @@ import (
 	"context"
 
 	"github.com/hashicorp/terraform-ls/internal/document"
+	"github.com/hashicorp/terraform-ls/internal/features/stacks/ast"
 	"github.com/hashicorp/terraform-ls/internal/features/stacks/state"
 	"github.com/hashicorp/terraform-ls/internal/job"
+	globalAst "github.com/hashicorp/terraform-ls/internal/terraform/ast"
 	"github.com/hashicorp/terraform-ls/internal/terraform/module/operation"
 	earlydecoder "github.com/hashicorp/terraform-schema/earlydecoder/stacks"
 )
@@ -17,7 +19,7 @@ import (
 // way that enables us to decode the rest of the configuration,
 // e.g. by knowing provider versions, etc.
 func LoadStackMetadata(ctx context.Context, stackStore *state.StackStore, stackPath string) error {
-	stack, err := stackStore.StackRecordByPath(stackPath)
+	record, err := stackStore.StackRecordByPath(stackPath)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


internal/features/stacks/jobs/metadata.go
```diff
@@ -34,13 +36,31 @@ func LoadStackMetadata(ctx context.Context, stackStore *state.StackStore, stackP
...
+
+	// Merge the new diagnostics with the existing ones
+	existingDiags, ok := record.Diagnostics[globalAst.HCLParsingSource]
+	if !ok {
+		existingDiags = make(ast.Diagnostics)
+	} else {
+		existingDiags = existingDiags.Copy()
+	}
+	for fileName, diagnostic := range diags {
+		existingDiags[ast.FilenameFromName(fileName)] = diagnostic
```
dbanck, line 60: This will overwrite any existing diagnostics for the file. We could either add a new diagnostics source for the early decoder or implement an `.Extend` method to append them.

internal/features/stacks/jobs/metadata.go
```diff
@@ -34,13 +36,36 @@ func LoadStackMetadata(ctx context.Context, stackStore *state.StackStore, stackP
 		return err
 	}
 
+	meta, diags := earlydecoder.LoadStack(record.Path(), record.ParsedFiles.AsMap())
```
dbanck, line 39: The LS panics when encountering an invalid source or version constraint:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x101084860]

goroutine 52 [running]:
github.com/hashicorp/terraform-schema/earlydecoder/stacks.LoadStack({0x140003e4237, 0x3b}, 0x14005e60930)
	/Users/daniel.banck/Projects/terraform-schema/earlydecoder/stacks/decoder.go:52 +0x670
github.com/hashicorp/terraform-ls/internal/features/stacks/jobs.LoadStackMetadata({0x102151988, 0x14005e60240}, 0x14000449560, {0x140068cd2c7, 0x3b})
	/Users/daniel.banck/Projects/terraform-ls/internal/features/stacks/jobs/metadata.go:39 +0x1e4
github.com/hashicorp/terraform-ls/internal/features/stacks.(*StacksFeature).decodeStack.func2({0x102151988, 0x14005e60240})
	/Users/daniel.banck/Projects/terraform-ls/internal/features/stacks/events.go:207 +0x60
github.com/hashicorp/terraform-ls/internal/scheduler.(*Scheduler).eval(0x14000448d50, {0x1021519c0, 0x140003d4460})
	/Users/daniel.banck/Projects/terraform-ls/internal/scheduler/scheduler.go:95 +0x8b4
created by github.com/hashicorp/terraform-ls/internal/scheduler.(*Scheduler).Start in goroutine 49
	/Users/daniel.banck/Projects/terraform-ls/internal/scheduler/scheduler.go:57 +0x224
```

dbanck (changes requested):
I noticed a few things when I tried to use this feature. These don't require changes to this PR, but rather to terraform-schema. I am sorry that I missed them during the review of https://github.com/hashicorp/terraform-schema/pull/412 and am only bringing them up now.

Maybe we can add more test cases to the decoder in terraform-schema to cover them.

---

We need to check that `req.Source` and `req.VersionConstraints` are both not null when iterating over the requirements, to avoid crashing due to a null pointer dereference.

Both diagnostic messages contain `nil` because we are using an empty field in our string (e.g. `pr.source` instead of `source.AsString()`).

CleanShot 2024-11-11 at 11 59 41@2x

CleanShot 2024-11-11 at 11 59 22@2x

dbanck (approved):

PR #1879, opened 2024-11-28
Add tests for Stacks feature

dbanck:
This PR adds a couple of tests for Stacks related jobs and the Stacks state store.

### DIFF
--- a/internal/features/stacks/jobs/schema_test.go
+++ b/internal/features/stacks/jobs/schema_test.go
@@ -1,0 +1,357 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package jobs
+
+import (
+	"bytes"
+	"compress/gzip"
+	"context"
+	"errors"
+	"io/fs"
+	"log"
+	"path/filepath"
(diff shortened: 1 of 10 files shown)


internal/features/stacks/jobs/schema_test.go
```diff
@@ -0,0 +1,357 @@
...
+			Data: []byte{},
+		},
+		filepath.Join(stackPath, "providers.tfstack.hcl"): &fstest.MapFile{
+			Data: []byte(`required_providers {
+	random = {
+		source = "hashicorp/random"
+		version = "1.0.0"
+	}
+}
+`),
```
ansgarm, line 65: Why do they have different contents? Is the first one only used to check if something exists but not actually read?

ansgarm (approved):
🧪

internal/features/stacks/jobs/schema_test.go
dbanck: Yeah, without the first entry, listing the files in `stackPath` will fail on Windows:
Screenshot 2024-11-28 170616

With both entries:
Screenshot 2024-11-28 170806

PR #1880, opened 2024-11-28
Fix flaky `workspace/symbol` tests

dbanck:
In the `workspace/symbol` tests, we'll send multiple `didOpen` requests for files in the same directory. The first request will trigger parsing of the whole directory (which is empty on disk) and set the module state to "loaded". The second `didOpen` will then do nothing, as it will respect the state already marked as loaded.

Depending on the order of execution, sometimes both files are opened before the first parsing job runs, sometimes the job runs between the two calls, causing the test to fail.

By writing the files to disk first, we ensure that both are parsed by the first `didOpen` call.

### DIFF
--- a/.changes/unreleased/INTERNAL-20241128-173155.yaml
+++ b/.changes/unreleased/INTERNAL-20241128-173155.yaml
@@ -1,0 +1,6 @@
+kind: INTERNAL
+body: Fix flaky `workspace/symbol` tests
+time: 2024-11-28T17:31:55.311726+01:00
+custom:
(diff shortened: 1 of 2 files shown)


ansgarm (approved):
Nice find!

ansgarm:
This flaky failure is unrelated I guess? 😄 

image

dbanck:
> This flaky failure is unrelated I guess? 😄

Yeah, I will look into that one next.

This PR is about
CleanShot 2024-11-28 at 21 11 45@2x

PR #1936, opened 2025-01-23
Set the version to 0.36.4-dev

jpogran:
This sets the current version to dev.


.changes/unreleased/INTERNAL-20250123-114649.yaml
```diff
@@ -0,0 +1,6 @@
+kind: INTERNAL
```
dbanck, line 1: I would bypass the changelog policy for this commit by adding a `dependencies` label to the PR 🤫

dbanck (approved):

